PR #4, opened 2015-06-29
[JENKINS-26129] Prepare for multibranch workflows

jglick:
[JENKINS-26129](https://issues.jenkins-ci.org/browse/JENKINS-26129)

Here I am just trying to make the minimal changes necessary to allow creation of a multibranch Workflow project type without being very disruptive for existing callers such as `literate`. It is not source compatible but so far it _seems_ binary compatible.

@reviewbybees esp. @stephenc

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.609.1</version>
+    <version>1.580.3</version> <!-- SCMTriggerItem -->
   </parent>
 
   <artifactId>branch-api</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.509</version>
+    <version>1.609.1</version>
   </parent>
 
   <artifactId>branch-api</artifactId>
@@ -102,7 +102,7 @@
     <dependency>
       <groupId>com.google.code.findbugs</groupId>
       <artifactId>jsr305</artifactId>
-      <version>1.3.9</version>
+      <version>3.0.0</version>
     </dependency>
     <!-- plugin dependencies -->
     <dependency>
@@ -166,7 +166,7 @@
       <plugin>
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>findbugs-maven-plugin</artifactId>
-        <version>2.5.2</version>
+        <version>3.0.1</version>
         <configuration>
           <excludeFilterFile>src/findbugs/excludesFilter.xml</excludeFilterFile>
           <failOnError>${maven.findbugs.failure.strict}</failOnError>
--- a/src/main/java/jenkins/branch/BranchProjectFactory.java
+++ b/src/main/java/jenkins/branch/BranchProjectFactory.java
@@ -106,7 +106,7 @@ public abstract class BranchProjectFactory<P extends Job<P, R> & TopLevelItem,
     /**
      * Gets the {@link Branch} that a specific project was configured for.
      *
-     * @param project the project.
+     * @param project the project; should assume {@link #isProject} has already been tested on it
      * @return the {@link Branch} that the project was configured for.
      */
     @NonNull
--- a/src/main/java/jenkins/branch/BranchProjectFactory.java
+++ b/src/main/java/jenkins/branch/BranchProjectFactory.java
@@ -28,13 +28,13 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.ExtensionPoint;
 import hudson.BulkChange;
 import hudson.XmlFile;
-import hudson.model.AbstractBuild;
 import hudson.model.AbstractDescribableImpl;
-import hudson.model.AbstractProject;
 import hudson.model.Descriptor;
 import hudson.model.Item;
+import hudson.model.Job;
 import hudson.model.JobProperty;
 import hudson.model.Project;
+import hudson.model.Run;
 import hudson.model.Saveable;
 import hudson.model.TopLevelItem;
 import hudson.tasks.BuildWrapper;
@@ -56,8 +56,8 @@ import java.util.List;
  * @param <R> the type of the builds of the branch projects.
  * @author Stephen Connolly
  */
-public abstract class BranchProjectFactory<P extends AbstractProject<P, R> & TopLevelItem,
-        R extends AbstractBuild<P, R>>
+public abstract class BranchProjectFactory<P extends Job<P, R> & TopLevelItem,
+        R extends Run<P, R>>
         extends AbstractDescribableImpl<BranchProjectFactory<?, ?>> implements Saveable, ExtensionPoint {
 
     /**
@@ -172,7 +172,7 @@ public abstract class BranchProjectFactory<P extends AbstractProject<P, R> & Top
     }
 
     /**
-     * Decorates the project in with all the {@link BranchProperty#decorator(hudson.model.AbstractProject)} instances.
+     * Decorates the project in with all the {@link BranchProperty#decorator(hudson.model.Job)} instances.
      * NOTE: This method should suppress saving the project and only affect the in-memory state.
      * NOTE: Override if the default strategy is not appropriate for the specific project type.
      *
@@ -198,8 +198,8 @@ public abstract class BranchProjectFactory<P extends AbstractProject<P, R> & Top
                 if (decorator != null) {
                     // if Project then we can feed the publishers and build wrappers
                     if (project instanceof Project) {
-                        DescribableList<Publisher, Descriptor<Publisher>> publishersList = project.getPublishersList();
-                        DescribableList buildWrappersList = Project.class.cast(project).getBuildWrappersList();
+                        DescribableList<Publisher, Descriptor<Publisher>> publishersList = ((Project) project).getPublishersList();
+                        DescribableList buildWrappersList = ((Project) project).getBuildWrappersList();
                         List<Publisher> publishers = decorator.publishers(publishersList.toList());
                         List<BuildWrapper> buildWrappers = decorator.buildWrappers(buildWrappersList.toList());
                         publishersList.replaceBy(publishers);
--- a/src/main/java/jenkins/branch/BranchProjectFactory.java
+++ b/src/main/java/jenkins/branch/BranchProjectFactory.java
@@ -179,7 +179,7 @@ public abstract class BranchProjectFactory<P extends Job<P, R> & TopLevelItem,
      * @param project the project.
      * @return the project for nicer method chaining
      */
-    @SuppressWarnings("ConstantConditions")
+    @SuppressWarnings({"ConstantConditions", "unchecked"})
     public P decorate(P project) {
         if (!isProject(project)) {
             return project;
@@ -194,14 +194,14 @@ public abstract class BranchProjectFactory<P extends Job<P, R> & TopLevelItem,
             List<BranchProperty> properties = new ArrayList<BranchProperty>(branch.getProperties());
             Collections.sort(properties, DescriptorOrder.reverse(BranchProperty.class));
             for (BranchProperty property : properties) {
-                ProjectDecorator<P, R> decorator = property.decorator(project);
+                JobDecorator<P, R> decorator = property.jobDecorator(project.getClass());
                 if (decorator != null) {
                     // if Project then we can feed the publishers and build wrappers
-                    if (project instanceof Project) {
+                    if (project instanceof Project && decorator instanceof ProjectDecorator) {
                         DescribableList<Publisher, Descriptor<Publisher>> publishersList = ((Project) project).getPublishersList();
-                        DescribableList buildWrappersList = ((Project) project).getBuildWrappersList();
-                        List<Publisher> publishers = decorator.publishers(publishersList.toList());
-                        List<BuildWrapper> buildWrappers = decorator.buildWrappers(buildWrappersList.toList());
+                        DescribableList<BuildWrapper, Descriptor<BuildWrapper>> buildWrappersList = ((Project) project).getBuildWrappersList();
+                        List<Publisher> publishers = ((ProjectDecorator) decorator).publishers(publishersList.toList());
+                        List<BuildWrapper> buildWrappers = ((ProjectDecorator) decorator).buildWrappers(buildWrappersList.toList());
                         publishersList.replaceBy(publishers);
                         buildWrappersList.replaceBy(buildWrappers);
                     }
--- a/src/main/java/jenkins/branch/BranchPropertyDescriptor.java
+++ b/src/main/java/jenkins/branch/BranchPropertyDescriptor.java
@@ -95,10 +95,12 @@ public abstract class BranchPropertyDescriptor extends Descriptor<BranchProperty
      * Or perhaps {@code it == null} when {@code l:renderOnDemand} is in use (supposed to work due to {@code capture} attribute, but…)?
      */
     @Restricted(DoNotUse.class)
-    @Extension public static final class Visibility extends DescriptorVisibilityFilter {
+    @Extension
+    public static final class Visibility extends DescriptorVisibilityFilter {
 
         @SuppressWarnings("rawtypes")
-        @Override public boolean filter(Object context, Descriptor descriptor) {
+        @Override
+        public boolean filter(Object context, Descriptor descriptor) {
             if (context instanceof MultiBranchProject && descriptor instanceof BranchPropertyDescriptor) {
                 return ((BranchPropertyDescriptor) descriptor).isApplicable((MultiBranchProject) context);
             } else {
--- a/src/main/java/jenkins/branch/BranchPropertyDescriptor.java
+++ b/src/main/java/jenkins/branch/BranchPropertyDescriptor.java
@@ -24,11 +24,15 @@
 package jenkins.branch;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.Extension;
 import hudson.ExtensionList;
 import hudson.model.Descriptor;
+import hudson.model.DescriptorVisibilityFilter;
 
 import java.util.ArrayList;
 import java.util.List;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.DoNotUse;
 
 /**
  * {@link Descriptor} for {@link BranchProperty}.
@@ -62,6 +66,7 @@ public abstract class BranchPropertyDescriptor extends Descriptor<BranchProperty
 
     /**
      * All the registered {@link BranchPropertyDescriptor}s.
+     * Probably unused.
      */
     public static List<BranchPropertyDescriptor> all() {
         return ExtensionList.lookup(BranchPropertyDescriptor.class);
@@ -69,7 +74,7 @@ public abstract class BranchPropertyDescriptor extends Descriptor<BranchProperty
 
     /**
      * Gets all the {@link BranchPropertyDescriptor} instances applicable to the specified project.
-     *
+     * Probably unused.
      * @param project the project
      * @return all the {@link BranchPropertyDescriptor} instances  applicable to the specified project.
      */
@@ -81,6 +86,26 @@ public abstract class BranchPropertyDescriptor extends Descriptor<BranchProperty
             }
         }
         return result;
+    }
+
+    /**
+     * Ensures that the configuration screen of (for example) {@link DefaultBranchPropertyStrategy} shows only appropriate descriptors.
+     * TODO this trick does not work for {@link NamedExceptionsBranchPropertyStrategy} on initial configuration (i.e., when {@link DefaultBranchPropertyStrategy} was initially selected);
+     * seems {@code filterDescriptors} is not called when {@code instance == null} perhaps?
+     * Or perhaps {@code it == null} when {@code l:renderOnDemand} is in use (supposed to work due to {@code capture} attribute, but…)?
+     */
+    @Restricted(DoNotUse.class)
+    @Extension public static final class Visibility extends DescriptorVisibilityFilter {
+
+        @SuppressWarnings("rawtypes")
+        @Override public boolean filter(Object context, Descriptor descriptor) {
+            if (context instanceof MultiBranchProject && descriptor instanceof BranchPropertyDescriptor) {
+                return ((BranchPropertyDescriptor) descriptor).isApplicable((MultiBranchProject) context);
+            } else {
+                return true;
+            }
+        }
+
     }
 
 }
--- a/src/main/java/jenkins/branch/BranchPropertyDescriptor.java
+++ b/src/main/java/jenkins/branch/BranchPropertyDescriptor.java
@@ -24,9 +24,8 @@
 package jenkins.branch;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
-import hudson.DescriptorExtensionList;
+import hudson.ExtensionList;
 import hudson.model.Descriptor;
-import jenkins.model.Jenkins;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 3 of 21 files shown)


src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -1350,9 +1339,11 @@ protected void execute(TaskListener listener) throws InterruptedException, IOExc
                 LOGGER.fine("Disabled. Skipping execution");
                 return;
             }
-
-            for (MultiBranchProject<?, ?> p
-                    : Jenkins.getInstance().getAllItems(MultiBranchProject.class)) {
+            Jenkins j = Jenkins.getInstance();
```
amuniz: What's the rule to use ```Jenkins.getActiveInstance``` or ```Jenkins.getInstance()```?

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -1870,12 +1864,16 @@ public void run() {
...
-                        if (entry.getKey().scheduleBuild(0, new SCMTrigger.SCMTriggerCause("Branch indexing"))) {
-                            try {
-                                factory.setRevisionHash(entry.getKey(), entry.getValue());
-                            } catch (IOException e) {
-                                e.printStackTrace(listener.error("Could not update last revision hash"));
+                        final P _project = entry.getKey();
+                        listener.getLogger().println("    " + factory.getBranch(_project).getName());
+                        if (_project instanceof ParameterizedJobMixIn.ParameterizedJob) {
+                            // TODO need a static helper method to schedule a ParameterizedJob
+                            if (new ParameterizedJobMixIn() {@Override protected Job asJob() {return _project;}}.scheduleBuild(0, new SCMTrigger.SCMTriggerCause("Branch indexing"))) {
```
amuniz: It's difficult to see this long lines in GH for review. It should be split, IMO.

daniel-beck: Difficult to read code formatting. Move condition to a local var.

src/main/java/jenkins/branch/MultiBranchProject.java
jglick: Well `getInstance` is `@CheckForNull` whereas `getActiveInstance` is `@Nonnull`. So you use `getInstance` if you can gracefully recover, and `getActiveInstance` where you would do no better than to just throw an `IllegalStateException` yourself.

amuniz:
:bee:

src/main/java/jenkins/branch/BranchProjectFactory.java
```diff
@@ -198,8 +198,8 @@ public P decorate(P project) {
                 if (decorator != null) {
                     // if Project then we can feed the publishers and build wrappers
                     if (project instanceof Project) {
-                        DescribableList<Publisher, Descriptor<Publisher>> publishersList = project.getPublishersList();
-                        DescribableList buildWrappersList = Project.class.cast(project).getBuildWrappersList();
+                        DescribableList<Publisher, Descriptor<Publisher>> publishersList = ((Project) project).getPublishersList();
+                        DescribableList buildWrappersList = ((Project) project).getBuildWrappersList();
```
varmenise: [question] why are we casting here? Isn't it project an instance of Project?

src/main/java/jenkins/branch/DeadBranchStrategy.java
```diff
@@ -45,7 +45,7 @@
      */
     private transient MultiBranchProject<?, ?> owner;
 
-    public abstract <P extends AbstractProject<P, R> & TopLevelItem, R extends AbstractBuild<P, R>>
+    public abstract <P extends Job<P, R> & TopLevelItem, R extends Run<P, R>>
     void runDeadBranchCleanup(TaskListener listener, Map<String, P> deadBranches) throws IOException,
             InterruptedException;
```
varmenise: [question] what's the concrete reason for updating from AbstractProject to Job?

src/main/java/jenkins/branch/BranchProjectFactory.java
daniel-beck: @varmenise We only know it's `P extends Job` from the generic types, so it's not necessarily a `Project`. However we passed the `if` condition so the cast is safe.

src/main/java/jenkins/branch/DeadBranchStrategy.java
daniel-beck: Workflows are not inheriting from `AbstractProject`. You'll see similar changes for workflow compatibility all over the place.

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -261,22 +261,7 @@ private synchronized void init() {
...
-                    f = ListView.class.getDeclaredField("includePattern");
-                    f.setAccessible(true);
-                    f.set(lv, Pattern.compile(".*"));
-                } catch (Throwable e) {
-                    LOGGER.log(Level.WARNING, "Initial view may be configured incorrectly", e);
-                }
-                lv.save();
-            } catch (IOException e) {
-                LOGGER.log(Level.WARNING, "Failed to set up the initial view", e);
-            }
```
tfennelly: This part of the change is not relevant to this PR, right? Looks as though it's just repairing an earlier issue.

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -1551,10 +1550,11 @@ public String getUrl() {
          */
         @SuppressWarnings("unused") // used from jelly pages
         public Node getBuiltOn() {
-            if (builtOn == null || builtOn.equals("")) {
-                return Jenkins.getInstance();
+            Jenkins j = Jenkins.getInstance();
+            if (builtOn == null || builtOn.isEmpty() || j == null) {
+                return j;
```
tfennelly: So `getBuiltOn` should now declare `@CheckForNull` on the return ?

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -1870,12 +1873,21 @@ public void run() {
...
-                            try {
-                                factory.setRevisionHash(entry.getKey(), entry.getValue());
-                            } catch (IOException e) {
-                                e.printStackTrace(listener.error("Could not update last revision hash"));
+                        final P _project = entry.getKey();
+                        if (!factory.isProject(_project)) {
+                            listener.getLogger().println("Detected unsupported subitem " + _project + ", skipping");
+                            continue;
+                        }
+                        listener.getLogger().println("    " + factory.getBranch(_project).getName());
```
tfennelly: A log with just a branch name in it?

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -2025,7 +2037,7 @@ public boolean scheduleBuild(int quietPeriod, Cause c, Action... actions) {
      * {@inheritDoc}
      */
     public void checkAbortPermission() {
-        checkPermission(AbstractProject.ABORT);
+        checkPermission(Item.CANCEL);
```
tfennelly: Separate to this PR ... seems like `ABORT` is deprecated, so shouldn't these methods be deprecated (and renamed perhaps)?

tfennelly:
Not sure what the new code review rules are but I think I probably can't give a +1 to this because, even though it all looks fine to me (a few small comments aside), there are changes in there that I can't say for sure are "right" e.g. the changes to `MultiBranchProject.run` wrt job scheduling/triggering.

src/main/java/jenkins/branch/MultiBranchProject.java
KostyaSha: Also `Jenkins.getActiveInstance` only in newer jenkins versions.

src/main/java/jenkins/branch/MultiBranchProject.java
jglick: The cleanup is only possible after having updated the Jenkins baseline, which this PR does.

src/main/java/jenkins/branch/MultiBranchProject.java
jglick: In fact it could have returned null to begin with (from `getNode`), but I will mark it for clarity.

src/main/java/jenkins/branch/MultiBranchProject.java
jglick: Yes. It is part of a multiline message. (This code was already there, I am just reindenting basically.)

src/main/java/jenkins/branch/MultiBranchProject.java
jglick: Well, this is part of the `Queue.Task` interface, so it cannot really be changed.

jglick:
No longer thought to be a work in progress. (I might need to make further changes to support some fixups in https://github.com/jenkinsci/workflow-plugin/pull/147 but this PR ought to be safe to merge as it stands.)

pom.xml
```diff
@@ -29,7 +29,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.509</version>
+    <version>1.609.1</version>
```
stephenc: Do we really have to go this far?

src/main/java/jenkins/branch/DeadBranchStrategyDescriptor.java
```diff
@@ -53,12 +53,9 @@ public boolean isApplicable(java.lang.Class<? extends MultiBranchProject> jobTyp
     @SuppressWarnings("unused") // used by stapler
     public static List<DeadBranchStrategyDescriptor> forProject(Class<? extends MultiBranchProject> jobType) {
         List<DeadBranchStrategyDescriptor> result = new ArrayList<DeadBranchStrategyDescriptor>();
-        for (Descriptor<DeadBranchStrategy> d : Jenkins.getInstance().getDescriptorList(DeadBranchStrategy.class)) {
-            if (d instanceof DeadBranchStrategyDescriptor) {
-                DeadBranchStrategyDescriptor descriptor = (DeadBranchStrategyDescriptor) d;
-                if (descriptor.isApplicable(jobType)) {
-                    result.add(descriptor);
-                }
+        for (DeadBranchStrategyDescriptor descriptor : ExtensionList.lookup(DeadBranchStrategyDescriptor.class)) {
```
stephenc: Unless I am very much mistaken this will instantiate a new parallel extension list of singletons and there will be two DeadBranchStrategyDescriptor "singletons", one stored in `Jenkins.descriptorLists` from the boot sequence processing and the other stored in `extensionLists` resulting from this lookup.

src/main/resources/jenkins/branch/MultiBranchProject/configure-common.jelly
```diff
@@ -23,17 +23,11 @@
  -->
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:p="/lib/hudson/project" xmlns:this="this">
-  <p:config-disableBuild/>
```
stephenc: Is there matching pull requests to add these back in?

it would seem possible to check if the MBP is instantiating subtypes of Project/AbstractProject and use a `<j:if>` to include them. Otherwise we should bump the major version of the branch api and flag compatability issues via the hpi plugin confif

pom.xml
stephenc: I would prefer if we can get away with 1.565 or 1.580...

stephenc:
I believe there is at least one :bug: in the DeadBranch descriptor processing

pom.xml
jglick: It is easier to test coördinated changes across several plugins when they are all using the same baseline, which must be 1.609.x for https://github.com/jenkinsci/workflow-plugin/pull/147. If you insist, I can try to backdate the core requirement, but does it really matter? This plugin has not had a release since October, and the downstream plugins are all considered more or less experimental. IMO you should not be using them at all if you are on a conservative core version.

src/main/java/jenkins/branch/DeadBranchStrategyDescriptor.java
jglick: Nope, those two fields are merely cached views over the underlying set of `ExtensionComponent`s, which the plugin system treats as singletons.

From `/script` on `hpi:run` of this project:

```groovy
println(ExtensionList.lookup(jenkins.branch.DeadBranchStrategyDescriptor));
println(Jenkins.instance.getDescriptorList(jenkins.branch.DeadBranchStrategy));
```

→

```
[jenkins.branch.DefaultDeadBranchStrategy$DescriptorImpl@1e1aaec5]
[jenkins.branch.DefaultDeadBranchStrategy$DescriptorImpl@1e1aaec5]
```

src/main/resources/jenkins/branch/MultiBranchProject/configure-common.jelly
jglick: > Is there matching pull requests to add these back in?

No, these were all removed intentionally—they should never have been there to begin with. The `disabled` field, for example, is defined in `AbstractProject`, and is not available in `MultiBranchProject`.

pom.xml
KostyaSha: :-1: for bumping without need. Experimental may become not experimental.

src/main/resources/jenkins/branch/MultiBranchProject/configure-common.jelly
```diff
@@ -23,17 +23,11 @@
...
-  <p:config-concurrentBuild />
```
jglick: Similarly, `concurrentBuild` would make no sense in `MultiBranchProject` even if it existed, because you would not want concurrent branch indexing. You _might_ want to specify that the branch projects use `concurrentBuild`, but if so, this would have to be a `BranchProperty` or the like.

pom.xml
stephenc: [CloudBees hat]
We should try to ensure that the parent is not updated beyond our oldest supported version unless there is a technical reason mandating the update
[/CloudBees hat]

My personal view is that I do not see any technical need for upping beyond 1.565 or 1.580 so I would prefer to keep one of those as the lower limit... Given that we are close to 1 year since 1.565 was cut I would be ok with 1.580... similarly if you can find a technical reason requiring the move above 1.580 I would be happy to consider, but just jumping straight to 1.609 without providing a good reason is not something I am happy with

jglick: > Experimental may become not experimental.

Of course, and I hope it will, but it is likely some months before we can confidently release a non-beta 1.0 of all this, at which point 1.609.x will not only be an LTS, it will be an _unsupported_ LTS. Whatever 1.565 users are using Literate today, presumably they are happy enough with the current releases, because there has not been one of Literate since Mar 2014.

src/main/java/jenkins/branch/DeadBranchStrategyDescriptor.java
stephenc: OK. That is good to have confirmation on.

src/main/resources/jenkins/branch/MultiBranchProject/configure-common.jelly
stephenc: ok

src/main/resources/jenkins/branch/MultiBranchProject/configure-common.jelly
stephenc: true

pom.xml
jglick: Plugin maintainer wants 1.565 so I will try to switch to that.

src/main/resources/jenkins/branch/MultiBranchProject/configure-common.jelly
```diff
@@ -23,17 +23,11 @@
...
 
   <st:include page="configure-basic.jelly" optional="true"/>
 
   <f:section title="${%Advanced Project Options}">
     <f:advanced>
-      <p:config-quietPeriod />
```
stephenc: what about these ones?

pom.xml
KostyaSha: If i right understand this plugin provides branch features for job and that's what i'm long waiting for git plugin. So if it is, then i will be glad to see not updated core, git and many other tools (that not yet ..ed by workflow) works on 1.565. I think we can also start operate with statistics and verify how many < 1.609.1 are exists before enforcing updates without reason.

stephenc: I am happy with 1.580 though I would prefer 1.565 if it is easy to get

jglick: `SCMTriggerItem` is `@since 1.568` so I need at least 1.580.x, trying that…

stephenc: [CloudBees hat]
1.565.1 was 2014/07/30 which is less than 7 days away... all hail 1.580 as the last line of supported LTS lines
[/CloudBees hat]

I'm fine with that technical reason for bumping to first LTS > 1.568

src/main/resources/jenkins/branch/MultiBranchProject/configure-common.jelly
jglick: `AbstractProject.quietPeriod` (anyway makes no sense for branch indexing that I can see…and as with `concurrentBuild`, if you wanted it on the branch project, we would need a separate system such as `BranchProperty`)

src/main/resources/jenkins/branch/MultiBranchProject/configure-common.jelly
```diff
@@ -23,17 +23,11 @@
...
-      <p:config-retryCount />
```
jglick: `AbstractProject.scmCheckoutRetryCount`, ditto

src/main/resources/jenkins/branch/MultiBranchProject/configure-common.jelly
```diff
@@ -23,17 +23,11 @@
...
-      <p:config-blockWhenUpstreamBuilding />
-      <p:config-blockWhenDownstreamBuilding />
```
jglick: `AbstractProject.blockBuildWhenUpstreamBuilding`/`.blockBuildWhenDownstreamBuilding`, ditto

jglick:
@KostyaSha

> If i right understand this plugin provides branch features for job and that's what i'm long waiting for git plugin.

Yes, but remember that the feature is only available for dedicated job types such as Literate and (soon) Workflow. While there are a couple of implementations floating around that offer a freestyle-like GUI configuration, this only works with a number of restrictions, and AFAIK there is no automatic importer from `FreestyleProject`. So do not expect multibranch to be a drop-in replacement for wildcards in `hudson.plugins.git.BranchSpec`; it should rather be seen (at least from my PoV) as deprecating them.

jglick:
@stephenc still awaiting an updated review from you (and on the two downstream PRs, if you can manage it).

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -2152,6 +2182,12 @@ public Authentication getDefaultAuthentication() {
         return ACL.SYSTEM;
     }
 
+    /* TODO 1.592+
+    @Override public Authentication getDefaultAuthentication(Queue.Item item) {
```
stephenc: As plugin maintainer I decree that all annotations on a method shall be on lines preceding the `(public|protected|/*package*/|private)`

src/main/java/jenkins/branch/ParameterDefinitionBranchProperty.java
```diff
@@ -0,0 +1,89 @@
...
+    protected ParameterDefinitionBranchProperty() {}
+
+    private List<ParameterDefinition> parameterDefinitions;
+
+    @Exported
+    public final List<ParameterDefinition> getParameterDefinitions() {
+        return parameterDefinitions;
+    }
+
+    @DataBoundSetter public final void setParameterDefinitions(List<ParameterDefinition> parameterDefinitions) {
```
stephenc: code style

src/main/java/jenkins/branch/ParameterDefinitionBranchProperty.java
```diff
@@ -0,0 +1,89 @@
...
+        this.parameterDefinitions = parameterDefinitions;
+    }
+
+    /** Not to be confused with {@link BranchPropertyDescriptor#isApplicable(MultiBranchProjectDescriptor)}, this checks applicability for the child job type. */
+    protected <P extends Job<P, B>, B extends Run<P, B>> boolean isApplicable(Class<P> clazz) {
+        return true;
+    }
+
+    @Override public final <P extends Job<P, B>, B extends Run<P, B>> JobDecorator<P, B> jobDecorator(Class<P> clazz) {
```
stephenc: code style

src/main/java/jenkins/branch/BranchPropertyDescriptor.java
```diff
@@ -84,4 +88,24 @@ protected boolean isApplicable(@NonNull MultiBranchProjectDescriptor projectDesc
...
     }
 
+    /**
+     * Ensures that the configuration screen of (for example) {@link DefaultBranchPropertyStrategy} shows only appropriate descriptors.
+     * TODO this trick does not work for {@link NamedExceptionsBranchPropertyStrategy} on initial configuration (i.e., when {@link DefaultBranchPropertyStrategy} was initially selected);
+     * seems {@code filterDescriptors} is not called when {@code instance == null} perhaps?
+     * Or perhaps {@code it == null} when {@code l:renderOnDemand} is in use (supposed to work due to {@code capture} attribute, but…)?
+     */
+    @Restricted(DoNotUse.class)
+    @Extension public static final class Visibility extends DescriptorVisibilityFilter {
```
stephenc: code style, annotations should be on preceding lines

src/main/java/jenkins/branch/BranchPropertyDescriptor.java
```diff
@@ -84,4 +88,24 @@ protected boolean isApplicable(@NonNull MultiBranchProjectDescriptor projectDesc
...
+
+        @SuppressWarnings("rawtypes")
+        @Override public boolean filter(Object context, Descriptor descriptor) {
```
stephenc: code style

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -307,12 +293,13 @@ protected void primaryView(String name) {
         final BranchProjectFactory<P, R> factory = getProjectFactory();
         factory.setOwner(this);
         Map<String, Map<String, P>> branchItems = getBranchItems();
-        if (getBranchesDir().isDirectory()) {
-            for (File branch : getBranchesDir().listFiles(new FileFilter() {
-                public boolean accept(File pathname) {
+        File[] branches = getBranchesDir().listFiles(new FileFilter() {
+                @Override public boolean accept(File pathname) {
```
stephenc: code style

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -450,6 +437,14 @@ public synchronized void setProjectFactory(BranchProjectFactory<P, R> projectFac
     }
 
     /**
+     * Offers direct access to the configurable list of branch sources.
+     * Intended for use from scripting and testing.
+     */
+    public @NonNull PersistedList<BranchSource> getSourcesList() {
```
stephenc: No way not in the middle :bug: code style

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -1550,11 +1566,12 @@ public String getUrl() {
          * @return {@code null}, for example if the slave that this build run no longer exists.
          */
         @SuppressWarnings("unused") // used from jelly pages
-        public Node getBuiltOn() {
-            if (builtOn == null || builtOn.equals("")) {
-                return Jenkins.getInstance();
+        public @CheckForNull Node getBuiltOn() {
```
stephenc: :bug: code style

src/main/java/jenkins/branch/UntrustedBranchProperty.java
```diff
@@ -130,5 +138,17 @@ public String getDisplayName() {
             }
             return result;
         }
+
+        @Override protected boolean isApplicable(MultiBranchProjectDescriptor projectDescriptor) {
```
stephenc: code style

stephenc:
No technical issues, but you know my preference for annotation placement of annotations on classes and methods so :bug: to changing the existing code style

KostyaSha:
@stephenc what about compromise? Allow FB annotations inlined (they don't very influence on code readability and already used everywhere), but deny other one-lined annotations?

stephenc:
:bee:

stephenc:
@KostyaSha I created https://github.com/jenkinsci/branch-api-plugin/blob/master/CONTRIBUTING.md to document the code style for this plugin. Ideally I would like that files can be idempotently formatted by all three standard IDEs such that even PRs can be run through a reformatting by IDE. But to get to that state I need to get the code formatting configuration for all three IDEs defined and run through idempotency tests... I got close earlier today but I need a bigger idempotency test bed to confirm. (plus Eclipse limits what formatting you can do compared to NetBeans and IntelliJ)

amuniz:
:bee:

PR #5, opened 2015-07-31
Always run BranchIndexing on master

jglick:
`MultiBranchProject` is a `FlyweightTask` so it does not require a heavyweight executor; it could “run” on any node, even `master` when configured to have zero executors. (At least as of JENKINS-7291 in 1.510.) In fact `BranchIndexing` was not doing anything at all with the `Computer` it was handed, other than using its `Charset`, which was probably wrong anyway.

For compatibility of old indexing logs I am retaining the `charset` field, in case it was “run on” a slave with a different charset than master (and there were non-ASCII branch names!), though I am tempted to just delete the field and hardcode `Charset.default()`.

Why do I care about this issue? It turned out there was a bug in `durable-task` which I will fix as a separate PR, and the combination of these two bugs prevented use of `mock-slave` clouds in https://github.com/jenkinsci/workflow-plugin/pull/166: `BranchIndexing` would run periodically on a one-shot slave and when it finished it would kill the slave even while it was doing real work.

@reviewbybees esp. @stephenc

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -1435,13 +1435,6 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
         protected String charset;
 
         /**
-         * Name of the slave this project was built on.
-         * Null or "" if built by the master. (null happens when we read old record that didn't have this information.)
-         */
-        @CheckForNull
-        private String builtOn;
-
-        /**
          * Constructor.
          *
          * @param previousIndexing any previous indexing.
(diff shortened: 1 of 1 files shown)


src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -1563,32 +1556,6 @@ public String getUrl() {
...
-
-        /**
-         * Returns the name of the slave it was built on; null or "" if built by the master.
-         * (null happens when we read old record that didn't have this information.)
-         */
-        @Exported(name = "builtOn")
-        @SuppressWarnings("unused") // used by Jenkins API
-        public String getBuiltOnStr() {
-            return builtOn;
-        }
```
tfennelly: Is there any risk that `getBuiltOn` and `getBuiltOnStr` were used somewhere else?

tfennelly:
:bee: assuming the above question is not relevant

src/main/java/jenkins/branch/MultiBranchProject.java
jglick: There is some risk, though I do not know where that would have been. Not in `literate-plugin`.

stephenc:
Will this still run on master when there are no dedicated executors on the master?

jglick:
> Will this still run on master when there are no dedicated executors on the master?

Yes; see PR description. (https://github.com/jenkinsci/workflow-plugin/pull/166 in fact demonstrates that case.)

stephenc:
:bee: so

PR #6, opened 2015-08-18
Need to define getBuildStatusIconClassName for compatibility with 1.576+ t:buildCaption

jglick:
Without this, the branch indexing log shows a broken icon in newer Jenkins releases.

https://github.com/jenkinsci/jenkins/commit/42190ed (from @tfennelly) was incompatible since it assumed that Jelly views were rendering one of the classes to which a new method was actually added, but in fact they can be and are used to render other classes via duck typing. `it.buildStatusIconClassName` was thus a poor choice; should have used `it.iconColor.iconClassName`, which would work with anything that already defined `public BallColor getIconColor()` (a reasonably safe assumption since the view was already evaluating `it.iconColor.description`).

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -1801,6 +1801,11 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
             return baseColor.anime();
         }
 
+        @SuppressWarnings("unused") // used from Jelly
+        public String getBuildStatusIconClassName() {
+            return getIconColor().getIconClassName();
+        }
+
         /**
          * {@inheritDoc}
          */


tfennelly:
:bee:

jglick:
This is an example of why I want Jelly/Groovy views to be replaced with a mixture of static HTML, and Java code that compiles (or fails to compile!) against well-defined interfaces.

amuniz:
:bee:

Do you mean to use JSP?

jglick:
> Do you mean to use JSP?

Eek…seems like stepping back in time a decade! Anyway that only compiles Java code the first time the page is rendered. I would want something compiled with the plugin. (There _is_ [support in Stapler](https://github.com/stapler/stapler/blob/53489356a69909947bfb695ee4286eb5c9bd9af5/jsp/src/main/java/org/kohsuke/stapler/jsp/JSPFacet.java#L48) for it.)

amuniz:
Yeah, that was why I was asking :smile:
The only sound of JSP makes me cry...

PR #8, opened 2015-10-08
OrganizationFolder

jglick:
Downstream of https://github.com/jenkinsci/cloudbees-folder-plugin/pull/20 and https://github.com/jenkinsci/scm-api-plugin/pull/3.

- [X] baseline update
- [x] introduce `OrganizationFolder`
- [x] make projects inside organization immutable
- [x] make `MultiBranchProject` extend `ComputedFolder`
- [x] retrofit `DeadBranchStrategy` to `OrphanedItemStrategy`

@reviewbybees esp. @stephenc


jglick:
Not clear to me how this is compiling successfully, since upstream builds are not running `deploy:deploy`, and I have not done so manually either. Perhaps the CI builds happen to be sharing a common `~/.m2/repository/`?

recena:
@jglick It can be possible. I found something similar with `1.625.1-SNAPSHOT` this morning.

src/main/java/jenkins/branch/MultiBranchProjectFactory.java
```diff
@@ -0,0 +1,121 @@
...
+                }
+                boolean empty;
+                try {
+                    empty = ((SCMSourceOwnerHack) owner).withSCMSourceCriteria(scmSource, getSCMSourceCriteria(scmSource), new Callable<Boolean>() {
+                        @Override public Boolean call() throws Exception {
+                            return scmSource.fetch(listener).isEmpty();
+                        }
+                    });
+                // This is ugly but lambdas do not seem to help at all: http://stackoverflow.com/q/18198176/12916
+                } catch (IOException x) {
```
tfennelly: It might be a bit ugly, but it is very easy to grok what's you want to happen, which trumps brevity for me any day.

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -0,0 +1,247 @@
...
+                        }
+                    };
+                }
+                @Override public void addAttribute(String key, Object value) throws IllegalArgumentException, ClassCastException {
+                    throw new IllegalArgumentException();
+                }
+            });
+        }
+        return projects;
+    }
```
tfennelly: I find it hard to get through the above code. All the anonymous inner classes etc. Would it be possible to compose it in a way that makes it a bit easier to see the moving parts?

tfennelly:
LGTM so far insofar as I get it :smile:. There's a lot going on there and a large amount of context that I don't have, but at least it smells good to me :smile: (even if I'm not sure what it's all doing and why)

src/main/java/jenkins/branch/MultiBranchProjectFactory.java
jglick: Well, lambdas done right—which is to say, abandoning checked exceptions—would be both clear _and_ concise. Oh well.

src/main/java/jenkins/branch/OrganizationFolder.java
jglick: Mm, not sure. It actually does require some thought to see what happens in what order, and the current source layout is probably as close to intuitive as you are going to get. In fact I used `SCMSourceObserver` as an opportunity to _change_ the order: now `MultiBranchProjectFactory.createProject` etc. happens as soon as `complete` is called, before other repositories have been discovered; the previous API forced `computeChildren` to wait until the navigator was done before it took any action. (Currently `ComputedFolder` does not go ahead and add the child and call `onCreatedFromScratch` and `afterNewItemCreated` until the folder recomputation is done, though I think it could be improved to be more incremental. Deleting orphaned children does have to wait until the end.)

src/main/java/jenkins/branch/MultiBranchProjectFactory.java
jglick: > abandoning checked exceptions

Or allowing the functional interface to be parameterized with a set of exception types of variable size. Seems like this has actually been proposed, but dropped from scope for Java 8 at least (not sure about 9).

stephenc:
I need to dig into this in detail, but looks good so far

src/main/java/jenkins/branch/Branch.java
```diff
@@ -94,6 +95,7 @@ public String getSourceId() {
      * @return the name of the branch.
      */
     public String getName() {
+        // TODO this could include a uniquifying prefix defined in BranchSource
```
amuniz: +1 It would be really useful to identify special branches (like PRs)

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -0,0 +1,241 @@
...
+                public void addAttribute(String key, Object value) throws IllegalArgumentException, ClassCastException {
+                    throw new IllegalArgumentException();
+                }
+            });
+        }
+    }
+
+    @Override
+    public List<SCMSource> getSCMSources() {
+        return Collections.emptyList(); // irrelevant unless onSCMSourceUpdated implemented
```
amuniz: Not used in this plugin, but I see it useful outside (a way to get the list of repositories managed by this `OrganizationFolder`). I think returning the list `SCMSource`s owned by the multiple `SCMNavigator`s is harmless (and possibly useful outside).

amuniz:
:bee:

src/main/java/jenkins/branch/Branch.java
```diff
@@ -42,6 +42,7 @@
 
 /**
  * A source code branch.
+ * Not to be subclassed outside this plugin.
```
stephenc: Could we add a check of same into the constructor(s)?

src/main/java/jenkins/branch/MultiBranchProjectFactory.java
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Creates a multibranch project for a given set of SCM sources if they seem compatible.
+     * @param parent a folder
+     * @param name a project name
+     * @param scmSources a set of SCM sources as added by {@link jenkins.scm.api.SCMSourceObserver.ProjectObserver#addSource}
+     * @param attributes a set of metadata attributes as added by {@link jenkins.scm.api.SCMSourceObserver.ProjectObserver#addAttribute}
+     * @param listener a way of reporting progress
+     * @return a new uninitialized multibranch project (do not configure its {@link MultiBranchProject#getSourcesList} or call {@link MultiBranchProject#onCreatedFromScratch}), or null if unrecognized
+     */
+    public abstract @CheckForNull MultiBranchProject<?,?> createProject(@Nonnull ItemGroup<?> parent, @Nonnull String name, @Nonnull List<? extends SCMSource> scmSources, @Nonnull Map<String,Object> attributes, @Nonnull TaskListener listener) throws IOException, InterruptedException;
```
stephenc: style guide

src/main/java/jenkins/branch/MultiBranchProjectFactory.java
```diff
@@ -0,0 +1,124 @@
...
+     * Creates a particular kind of multibranch project insofar as at least one {@link SCMHead} satisfies a probe.
+     */
+    public static abstract class BySCMSourceCriteria extends MultiBranchProjectFactory {
+
+        /**
+         * Defines how to decide whether or not a given repository should host our type of project.
+         * @param source a repository
+         * @return criteria for treating its branches as a match
+         */
+        protected abstract @Nonnull SCMSourceCriteria getSCMSourceCriteria(@Nonnull SCMSource source);
```
stephenc: style guide

src/main/java/jenkins/branch/MultiBranchProjectFactory.java
```diff
@@ -0,0 +1,124 @@
...
+
+        /**
+         * Creates a project given that there seems to be a match.
+         * @param parent the folder
+         * @param name the project name
+         * @param attributes a set of metadata attributes as added by {@link jenkins.scm.api.SCMSourceObserver.ProjectObserver#addAttribute}
+         * @return a new project suitable for {@link #createProject}
+         */
+        protected abstract @Nonnull MultiBranchProject<?,?> doCreateProject(@Nonnull ItemGroup<?> parent, @Nonnull String name, @Nonnull Map<String,Object> attributes);
```
stephenc: style guide

src/main/java/jenkins/branch/MultiBranchProjectFactoryDescriptor.java
```diff
@@ -0,0 +1,48 @@
...
+
+    protected MultiBranchProjectFactoryDescriptor(Class<? extends MultiBranchProjectFactory> clazz) {
+        super(clazz);
+    }
+
+    /**
+     * Creates a factory instance with a default configuration.
+     * @return a default factory, or null to not registered this factory by default
+     */
+    public abstract @CheckForNull MultiBranchProjectFactory newInstance();
```
stephenc: style guide

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -0,0 +1,241 @@
...
+import jenkins.scm.api.SCMSourceObserver;
+import jenkins.scm.api.SCMSourceOwner;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+import org.kohsuke.stapler.StaplerRequest;
+import org.kohsuke.stapler.StaplerResponse;
+
+/**
+ * A folder-like collection of {@link MultiBranchProject}s, one per repository.
+ */
```
stephenc: `@since`

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -0,0 +1,241 @@
...
+public final class OrganizationFolder extends ComputedFolder<MultiBranchProject<?,?>> implements SCMSourceOwner, MultiBranchProjectFactory.SCMSourceOwnerHack {
+
+    private final DescribableList<SCMNavigator,SCMNavigatorDescriptor> navigators = new DescribableList<SCMNavigator, SCMNavigatorDescriptor>(this);
+    private final DescribableList<MultiBranchProjectFactory,MultiBranchProjectFactoryDescriptor> projectFactories = new DescribableList<MultiBranchProjectFactory,MultiBranchProjectFactoryDescriptor>(this);
+
+    public OrganizationFolder(ItemGroup parent, String name) {
+        super(parent, name);
+    }
+
+    @Override
```
stephenc: Where the contract is unmodified from the contract of the inherited method, please document that you have assessed this and record with `/** {@inheritDoc} */`

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -0,0 +1,241 @@
...
+    public SCMSource getSCMSource(String sourceId) {
+        return null;
+    }
+
+    @Override
+    public void onSCMSourceUpdated(SCMSource source) {
+        // possibly we should recheck whether this project remains valid
```
stephenc: Should this be a TODO?

src/main/java/jenkins/branch/SpecificDescriptor.java
```diff
@@ -0,0 +1,102 @@
...
+import hudson.model.TopLevelItem;
+import hudson.model.TopLevelItemDescriptor;
+import hudson.model.View;
+import jenkins.scm.api.SCMNavigatorDescriptor;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+
+/**
+ * Ensures that when we have a non-null {@link SCMNavigatorDescriptor#newInstance(String)}, a special <b>New Item</b> entry is displayed instead of the generic {@link OrganizationFolder.DescriptorImpl}.
+ */
```
stephenc: `@since`

src/main/java/jenkins/branch/SpecificDescriptor.java
```diff
@@ -0,0 +1,102 @@
...
+    @Override
+    public TopLevelItem newInstance(ItemGroup parent, String name) {
+        OrganizationFolder p = new OrganizationFolder(parent, name);
+        p.getNavigators().add(delegate.newInstance(name));
+        return p;
+    }
+
+    @SuppressWarnings("deprecation") // dynamic registration intentional here
+    @Initializer(after=InitMilestone.PLUGINS_STARTED, before=InitMilestone.EXTENSIONS_AUGMENTED)
+    public static void addSpecificDescriptors() {
```
stephenc: Interesting hack!

src/main/resources/jenkins/branch/OrganizationFolder/configure-entries.jelly
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?jelly escape-by-default='true'?>
```
stephenc: MIT license

src/main/resources/jenkins/branch/MultiBranchProjectFactory/config.jelly
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?jelly escape-by-default='true'?>
```
stephenc: MIT License

src/main/resources/jenkins/branch/OrganizationFolder/newInstanceDetail.jelly
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+The MIT License
```
stephenc: Please keep consistent with the existing license in jelly file styling

stephenc:
There are some methods in `MultiBranchProject` that look like they retain the contract of the super method and could be documented with `/** {@inheritDoc} */` to indicate.

Other than that, only a :bug: for the niggly bits needed to merge this in and prevent the code style from deteriorating

src/main/java/jenkins/branch/Branch.java
jglick: Well, that is handled by a name prefix given by the `SCMSource`. The idea here is a _user_-defined prefix that would enable you to add multiple sources with otherwise clashing names, potentially giving them distinct branch properties. TBD whether this is really necessary; @stephenc and I had an example in mind but on reflection this feature would neither be necessary nor usable in that scenario.

Anyway I was just recording a note that even though `BranchProjectFactory.newInstance` is now constrained to honor `Branch.name`—which the existing implementations already did, except `YamlProjectFactory` which is going to need revision anyway to provide a unified fix for JENKINS-30744—we are not prevented from having project names include prefixes, since there is no requirement that `Branch.name` equals `SCMHead.name`.

src/main/java/jenkins/branch/OrganizationFolder.java
jglick: Potentially yes. Basically it would be the union of the sources defined by the folder’s current children, I suppose.

src/main/java/jenkins/branch/Branch.java
jglick: It is subclassed in `YamlProjectFactoryTest`, but I think this is indeed possible using `@Restricted`, will try it.

jglick: Never mind, it is not possible, since the regular constructor is callable as `public` so we cannot create a `protected` constructor and restrict it.

src/main/java/jenkins/branch/OrganizationFolder.java
jglick: No, this class is `@Restricted` for now at least.

src/main/java/jenkins/branch/SpecificDescriptor.java
jglick: Not an API class.

src/main/java/jenkins/branch/SpecificDescriptor.java
jglick: Not one I am happy with. I was thinking for a while we needed a way to inject additional items into an `ExtensionList`, but in this case I think it would suffice for `DescriptorVisibilityFilter` to be permitted to _add_ descriptors as well as remove them, since this is purely a UI modification.

src/main/java/jenkins/branch/Branch.java
stephenc: I was thinking more something like `assert getClass().getPackage() == Branch.getClass().getPackage()` or otherwise asserting that the subclass is within the same jar as Branch.class

jglick: But that would cause test failures in `yaml-project` (assuming tests are run with assertions enabled). The test subclasses are harmless mocks. I think the Javadoc note suffices.

stephenc:
:bee:

PR #9, opened 2015-10-27
JobPropertyDescriptor.newInstance can return null

jglick:
Needed for proper _Snippet Generator_ behavior in https://github.com/jenkinsci/workflow-plugin/pull/209. Something like https://github.com/jenkinsci/jenkins/pull/1656 would have made this more natural.

@reviewbybees esp. @stephenc


stephenc:
:bee:

tfennelly:
:bee: 

I think a comment in the `newInstance` method would do no harm i.e. why it returns `null` when `prop.getThrottle()` returns null.

jglick:
Well you can always jump to the super method which explains it. Anyway I intend for the implementation to be replaced when a core API is ready.

PR #10, opened 2015-10-27
Removing misleading annotation

jglick:
In fact this method _is_ used by Java code, in the same source file in fact.

@reviewbybees esp. @stephenc


stephenc:
:bee:

recena:
:bee:

PR #11, opened 2015-10-29
Ensure that SCMSource.setOwner is called

jglick:
@recena noticed that in some `workflow-multibranch` tests, `AbstractGitSCMSource.retrieve` was getting null from `getCriteria()`. The reason is that nothing was calling `SCMSource.setOwner`. Normally this is called from `MultiBranchProject.onLoad` or `.submit`, but in the case of test code using `getSourcesList().add(…)` it was not. This patch just ensures that however the sources list is modified, `setOwner` will be called.

@reviewbybees


amuniz:
:bee:

recena:
:bee:

tfennelly:
:bee:

PR #12, opened 2015-11-02
[JENKINS-30252] Defining environment variable BRANCH_NAME

jglick:
[JENKINS-30252](https://issues.jenkins-ci.org/browse/JENKINS-30252)

Could have restricted this to Workflow multibranch, but seemed generally useful.

@reviewbybees esp. @stephenc

PR #13, opened 2015-11-02
[JENKINS-30595] Implemented API

jglick:
Downstream of https://github.com/jenkinsci/scm-api-plugin/pull/4.

@reviewbybees esp. @stephenc, @recena, @amuniz

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -108,7 +108,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>scm-api</artifactId>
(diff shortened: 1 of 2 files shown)


jglick:
Could also use this to simplify `BranchNameContributor` slightly, now that #12 has been merged, but does not make much difference. (_Could_ now move `BranchNameContributor` into `scm-api` but that seems somehow inappropriate.)

stephenc:
:bee: as I understand it

recena:
:bee:

PR #14, opened 2015-11-03
Suppress non-read View permissions on MultiBranchProject within OrganizationFolder

jglick:
Otherwise the user is allowed to effectively modify what is supposed to be a completely generated folder.

@reviewbybees esp. @stephenc, @kzantow


recena:
:bee:

stephenc:
:bee:

PR #16, opened 2015-11-04
Add PeriodicFolderTrigger by default

jglick:
Adding a one-day fallback trigger to newly created organization folders, since you normally want some kind of fallback even if there are webhooks.

Also adding it to the child multibranch projects. Ideally this would be something you could configure, somehow, but there is no mechanism for making such configurations currently, so we should at least have a sensible setting.

Downstream of https://github.com/jenkinsci/cloudbees-folder-plugin/pull/22.

@reviewbybees esp. @stephenc

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,7 +113,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>cloudbees-folder</artifactId>
-      <version>4.11-beta-1</version>
+      <version>5.1-SNAPSHOT</version>
     </dependency>
     <!-- jenkins dependencies -->
     <!-- test dependencies -->
--- a/src/main/java/jenkins/branch/OrganizationFolder.java
+++ b/src/main/java/jenkins/branch/OrganizationFolder.java
@@ -24,9 +24,11 @@
 
 package jenkins.branch;
 
+import antlr.ANTLRException;
 import com.cloudbees.hudson.plugins.folder.AbstractFolderDescriptor;
 import com.cloudbees.hudson.plugins.folder.computed.ChildObserver;
 import com.cloudbees.hudson.plugins.folder.computed.ComputedFolder;
+import com.cloudbees.hudson.plugins.folder.computed.PeriodicFolderTrigger;
 import hudson.Extension;
 import hudson.ExtensionList;
 import hudson.model.Descriptor;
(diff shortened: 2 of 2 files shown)


stephenc:
:bug: on the funky `assert false : x` much rather throw an illegal state exception wrapping the exception

Otherwise fine

jglick:
Seems like a legal use of assertions to me, but will change if you like.

stephenc:
@jglick so if you do

`assert false : "should never happen"` that would be ok and not funky

but when you do

`assert false : x` 

that is really 

`assert false : x == null ? null : x.toString()`

Which will throw away the stack trace of the exception... which is the funkyness I was referring to... namely that the second expression in the assert should be String-like (with coercion via `toString()`)
screen shot 2015-11-04 at 5 04 58 p m

So it takes a few minutes to grock what is trying to go on with that statement and in the end, if it ever triggers we don't even get to know what exactly triggered it as the stack trace will be discarded

jglick:
> when you do
> 
> `assert false : x`
> 
> that is really
> 
> `assert false : x == null ? null : x.toString()`

Not true.

```java
public class AssertDemo {
    public static void main(String[] args) {
        try {
            helper();
        } catch (Exception x) {
            assert false : x;
        }
    }
    static void helper() {
        throw new NullPointerException("oops");
    }
}
```

→ (with `-ea`)

```
Exception in thread "main" java.lang.AssertionError: java.lang.NullPointerException: oops
	at AssertDemo.main(AssertDemo.java:6)
Caused by: java.lang.NullPointerException: oops
	at AssertDemo.helper(AssertDemo.java:10)
	at AssertDemo.main(AssertDemo.java:4)
```

`javap` confirms that it compiles to

```
       7: getstatic     #4                  // Field $assertionsDisabled:Z
      10: ifne          22
      13: new           #5                  // class java/lang/AssertionError
      16: dup
      17: aload_1
      18: invokespecial #6                  // Method java/lang/AssertionError."<init>":(Ljava/lang/Object;)V
      21: athrow
```

which invokes

```java
    public AssertionError(Object detailMessage) {
        this(String.valueOf(detailMessage));
        if (detailMessage instanceof Throwable)
            initCause((Throwable) detailMessage);
    }
```

Anyway, do I have a bee or what? :-)

stephenc:
:bee: despite your JVM seeming to not follow the JLS strictly

amuniz:
:bee:

PR #18, opened 2015-11-06
[JENKINS-30744] Introduced Branch.encodedName

jglick:
[JENKINS-30744](https://issues.jenkins-ci.org/browse/JENKINS-30744)

Yes, this is incompatible, but I see no way around that.

@reviewbybees esp. @stephenc

### DIFF
--- a/src/main/java/jenkins/branch/Branch.java
+++ b/src/main/java/jenkins/branch/Branch.java
@@ -23,22 +23,13 @@
  */
 package jenkins.branch;
 
-import hudson.model.Descriptor;
-import hudson.model.Job;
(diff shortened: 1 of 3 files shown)


stephenc:
:bee: Yup I concur

amuniz:
:bee:
Not sure in which way is incompatible. Do you mean for jobs already containing a slash?

jglick:
> Not sure in which way is incompatible. Do you mean for jobs already containing a slash?

Specifically for users of existing plugin releases (mainly Literate and YAML Project) with multibranch projects involving slashy branch names who do not choose to upgrade to releases with my downstream PRs.

PR #19, opened 2015-11-06
[JENKINS-31432] NPE on SCMSourceOwner.getSCMSources() call

amuniz:
[JENKINS-31432](https://issues.jenkins-ci.org/browse/JENKINS-31432)

@reviewbybees

### DIFF
--- a/src/test/java/jenkins/branch/MultibranchImplementationTest.java
+++ b/src/test/java/jenkins/branch/MultibranchImplementationTest.java
@@ -27,6 +27,7 @@ package jenkins.branch;
 import java.io.IOException;
 
 import org.junit.Assert;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.jvnet.hudson.test.JenkinsRule;
@@ -45,6 +46,7 @@ public class MultibranchImplementationTest {
     public static boolean check = true;
 
     @Test
+    @Ignore("JENKINS-31432 neds to be fixed")
     public void createMultiBranchProjectWithListenerTest() throws IOException {
         j.jenkins.createProject(MultiBranchImpl.class, "test");
         Assert.assertTrue("SCMSourceOwner.getSCMSources() should never throw NullPointerException", check);
--- a/src/test/java/jenkins/branch/MultibranchImplementationTest.java
+++ b/src/test/java/jenkins/branch/MultibranchImplementationTest.java
@@ -1,0 +1,70 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2015 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package jenkins.branch;
+
+import java.io.IOException;
+
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+import org.jvnet.hudson.test.TestExtension;
+
+import hudson.model.Item;
+import hudson.model.listeners.ItemListener;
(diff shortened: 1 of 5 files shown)


src/test/java/jenkins/branch/MultibranchImplementationTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class MultibranchImplementationTest {
+
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
+
+    public static boolean check = true;
+
+    @Test
+    @Ignore("JENKINS-31432 neds to be fixed")
```
jglick: :ant: typo

src/test/java/jenkins/branch/MultibranchImplementationTest.java
```diff
@@ -0,0 +1,72 @@
...
+    public static class SCMSourceOwnerListener extends ItemListener {
+
+        @Override
+        public void onUpdated(Item item) {
+            if (item instanceof SCMSourceOwner) {
+                try {
+                    ((SCMSourceOwner) item).getSCMSources();
+                } catch (NullPointerException e) {
+                    check = false;
+                    throw e;
```
jglick: :ant: not a good test style, since this will make no sense after the bug is fixed, and will not catch unrelated problems. Better to set `check` to `false` initially, then set it to `true` after calling `getSCMSources` (with no `catch` block).

src/test/java/jenkins/branch/harness/MultiBranchImpl.java
```diff
@@ -0,0 +1,124 @@
...
+    public List<SCMSource> getSCMSources() {
+        if (getSourcesList() == null) {
+            // test code is generating a NullPointer when calling it from an ItemListener on the SCMSourceOwner
+            // It seems that the object is not fully initialized when the ItemListener uses it.
+            // Perhaps it needs to be reproduced and investigated in a branch-api test.
+            return new ArrayList<SCMSource>();
+        }
+        return super.getSCMSources();
+    }
+    */
```
jglick: Can just delete this comment block since after a fix, who is going to care?

src/test/java/jenkins/branch/harness/MultiBranchImpl.java
```diff
@@ -0,0 +1,124 @@
...
+
+    public static class BranchProjectFactoryImpl extends BranchProjectFactory<FreeStyleProject, FreeStyleBuild> {
+
+        @Override
+        public FreeStyleProject newInstance(Branch branch) {
+            FreeStyleProject job = new FreeStyleProject(getOwner(), branch.getName());
+            job.onCreatedFromScratch();
+            FreeStyleProject spied = spy(job);
+            // Do nothing.. Running the actual build is not desired/required (and not possible) in this tests.
+            when(spied.scheduleBuild()).thenReturn(false);
```
jglick: Could do this without mocks by using a custom `Project` subclass that overrides `scheduleBuild`.

src/test/java/jenkins/branch/harness/MultiBranchImpl.java
```diff
@@ -0,0 +1,124 @@
...
+        return super.getSCMSources();
+    }
+    */
+
+    public static class BranchProjectFactoryImpl extends BranchProjectFactory<FreeStyleProject, FreeStyleBuild> {
+
+        @Override
+        public FreeStyleProject newInstance(Branch branch) {
+            FreeStyleProject job = new FreeStyleProject(getOwner(), branch.getName());
+            job.onCreatedFromScratch();
```
jglick: :bug: this is not to be called by `newInstance` implementations as per documentation.

src/test/java/jenkins/branch/MultibranchImplementationTest.java
amuniz: Surprisingly `onUpdated` is called more than once within a single call to `j.jenkins.createProject` and only the first call is producing the null pointer. So if I remove the try-catch then `check` is set in the next call and the test passes (when it should not). Any idea to improve the test in other way?

amuniz: (and ItemListener is [eating](https://github.com/jenkinsci/jenkins/blob/e2dccc35572b64f5e602a6b1b3efb517c578872c/core/src/main/java/hudson/model/listeners/ItemListener.java#L169) the exception, but I'm sure you already knew that).

jglick: > `onUpdated` is called more than once within a single call to `j.jenkins.createProject`

The block in `MultiBranchProject.computeChildren` between the call to `BranchProjectFactory.newInstance` and the call to `scheduleBuild` might need to be wrapped in a `BulkChange`.

> Any idea to improve the test in other way?

Replace `check` with an `Exception` field initially null. The `catch` block would populate it. If non-null in the test case, rethrow it.

amuniz:
The `MultiBranchProject(ItemGroup parent, String name)` constructor uses the super constructor, which in turns calls `AbstractFolder.init` which leads in a `View.save` and finally a `AbstractFolder.save`, the thing is that the `AbstractFolder` which `save` is called on, is the same object that we are initializing in `MultiBranchProject` constructor.

As a result, `onUpdated` is being called on a object that it's not even fully constructed (!).

I think this is the root cause of the issue. I'm trying to fix it now.

jglick:
> `AbstractFolder.init` which leads in a `View.save`

This is what ought to be fixed, using `BulkChange.abort` if nothing better seems available.

src/test/java/jenkins/branch/MultibranchImplementationTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @TestExtension
+    public static class SCMSourceOwnerListener extends ItemListener {
+
+        @Override
+        public void onUpdated(Item item) {
+            if (item instanceof SCMSourceOwner) {
+                try {
+                    ((SCMSourceOwner) item).getSCMSources();
+                } catch (NullPointerException e) {
```
jglick: :ant: better to catch _any_ exception here, to make the test stronger.

src/test/java/jenkins/branch/MultibranchImplementationTest.java
```diff
@@ -0,0 +1,68 @@
...
+                    check = new Exception(e);
```
jglick: :ant:

```java
check = e;
```

would suffice.

stephenc:
:bee:

jglick:
:bee:

PR #17, opened 2015-11-06
Improved UI when no projects in Org folders or Multibranch

kzantow:
Better informative view when no repositories are present with Jenkinsfiles in Organization Folders, or no branches are available with Jenkinsfiles in multibranch projects.

https://issues.jenkins-ci.org/browse/JENKINS-31381

@reviewbybees 

Screenshots:

Multibranch:
welcome-multibranch

Org Folder:
welcome-org-folder

There is a related PR, adding specific information for the Workflow/Jenkinsfile project recognizer here: https://github.com/jenkinsci/workflow-plugin/pull/248

PR #22, opened 2015-12-22
[JENKINS-32178] Create a working mechanism for users to create additional views.

mjdetullio:
[JENKINS-32178](https://issues.jenkins-ci.org/browse/JENKINS-32178)

Introduce BranchListView, an extension of ListView that is only
available to MultiBranchProjects.  The view has an additional URL
binding for "branch" that is necessary for job links to work.  Standard
views do not have this binding, so links to branch jobs from standard
non-primary views end up broken.  This view is marked as
non-instantiable so it does not show outside of MultiBranchProjects.

Build the message from MultiBranchProjectWelcomeView into
BranchListView by overriding the inherited noJob.jelly.  This allows us
to remove the primary view override and also more accurately displays
configuration prompts to the user.

Add newView.jelly to MultiBranchProject, overriding the inherited view.
The new view now uses MultiBranchProject#getViewDescriptors() to
determine views that users are allowed to create.

Add a migration path for views that deletes all non-compatible view
types and adds a BranchListView if all views were deleted.

Override view initialization to use BranchListView instead of AllView.


jglick:
`BranchListView` seems like a bad idea. We do not want to limit users to only using list views. If there is no other way to make Stapler understand how to resolve `job/mbp/view/anything/branch/master/` (`StaplerFallback`?), then IMO it would be better to just delete the override of `getUrlChildPrefix` in `MultiBranchProject` (and deprecate this in `AbstractFolder`).

Need test coverage for whatever the ultimate solution is. Should be straightforward with `JenkinsRule.createWebClient`.

The `noJob` part clashes with the more featureful #17.

mjdetullio:
Yes, not a good solution, but demonstrates the problem.  This is more or less copy/paste from my other plugin.

The object needing ```StaplerFallback``` is the view, which would then have to fallback to the parent, which falls back to the primary view, which falls back to ... you see where I'm going with this.

The most stable solution is deleting the ```getUrlChildPrefix``` override.

The only other option I could think is to add an interface to the API for all views to optionally implement to make compatible with the prefix.  Core would then need this API as a dependency, and you could implement it in ```hudson.model.View```, but at that point why bother with those changes when you'd just make core changes to better support overrides of ```getUrlChildPrefix```.

jglick:
> The object needing `StaplerFallback` is the view

I was rather thinking `MultiBranchProject` getting some Stapler binding method so that `view/*/` returns an object implementing `StaplerFallback`, though this looks hard. A `getUrlChildPrefix` override deletion would be much less risky I think.

philippelabat:
Any update on this?

mjdetullio:
Closing in favor of https://github.com/jenkinsci/branch-api-plugin/pull/35

PR #23, opened 2015-12-22
[JENKINS-32179] Branch indexing always attempts to create a new project for conflicting branch names from multiple sources

mjdetullio:
[JENKINS-32179](https://issues.jenkins-ci.org/browse/JENKINS-32179)

observer.mayCreate is a slight misnomer, since it actually marks the
branch as observed and its return value is whether it has been observed.
Move this check to before updating/creating branches, since we don't
want to do either if the branch has been observed already.

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -299,6 +299,19 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
                     String rawName = branch.getName();
                     String encodedName = branch.getEncodedName();
                     P project = observer.shouldUpdate(encodedName);
+
+                    /*
+                     * Check observer.mayCreate:
+                     * 1. after observer.shouldUpdate so removal from the orphaned items list will still occur
+                     * 2. to check if this branch has been observed, and mark it as observed if it was not
+                     * 3. before verifying observer.shouldUpdate's retval, because if this branch has already been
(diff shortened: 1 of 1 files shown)


jglick:
I suppose this is about the _second_ indexing, when we already have a child with the duplicated name.

Besides the fact that you neglected to change `OrganizationFolder` in the parallel way, this patch looks wrong. I suspect it is not even in the right plugin: more likely `ComputedFolder.updateChildren`’s `shouldUpdate` should be calling `observed.add(name)`.

Any change of this kind requires test coverage. Probably `cloudbees-folder` should host tests for general `ComputedFolder` logic.

mjdetullio:
You mention instead fixing it in the Folders Plugin.

If this PR is deregarded, the only viable solution I can think is to add the following before this line https://github.com/jenkinsci/cloudbees-folder-plugin/blob/d5421010bac1ac623397777c348706cd8656e9c9/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java#L154

```
if (!observed.add(name)) {
    return null;
}
// then try to pop from orphaned set
```

jglick:
Something like that, yeah.

jglick:
I think I have the right fix prepared.

mjdetullio:
Thanks for hitting this one. I haven't had time to toy with the multi-branch stuff the past couple months.  It would help if I was dogfooding it at ${dayJob} but that's not the case at the moment.

PR #24, opened 2015-12-26
Remove useless MultiBranchProjectDescriptor#getSCMDescriptors() API.

mjdetullio:
`SingleSCMSource`'s config-detail.jelly uses
[`SingleSCMSource.DescriptorImpl#getSCMDescriptors(SCMSourceOwner)`](https://github.com/jenkinsci/scm-api-plugin/blob/7c99d7f705408a457065fa2d9ef89be8b900b805/src/main/java/jenkins/scm/impl/SingleSCMSource.java#L183-L204),
contrary to the Javadoc on the method removed by this commit.

If you were to implement the method removed by this commit, you could make the method always throw an `IllegalStateException` and it would never occur.

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProjectDescriptor.java
+++ b/src/main/java/jenkins/branch/MultiBranchProjectDescriptor.java
@@ -35,7 +35,20 @@ import java.util.ArrayList;
 import java.util.List;
 
 /**
- * The {@link Descriptor} for {@link MultiBranchProject}s.
+ * <p>The {@link Descriptor} for {@link MultiBranchProject}s.</p>
+ *
+ * <p>Compatible {@link hudson.scm.SCM}s displayed by {@link jenkins.scm.impl.SingleSCMSource} (via their
+ * {@link hudson.scm.SCMDescriptor}) can be defined by overriding {@link #isApplicable(Descriptor)}:</p>
+ * <pre>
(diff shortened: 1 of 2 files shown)


jglick:
See also [`yaml-project`](https://github.com/jenkinsci/yaml-project-plugin/blob/83a0d05bc7eb145190c4d25e4284adfaf365e334/src/main/java/com/google/jenkins/plugins/dsl/YamlMultiBranchProject.java#L202), [`literate`](https://github.com/jenkinsci/literate-plugin/blob/cead08b391a33f6408928d372fe761a7dfb02222/src/main/java/org/cloudbees/literate/jenkins/LiterateMultibranchProject.java#L289), and [`freestyle-multibranch`](https://github.com/jenkinsci/freestyle-multibranch-plugin/blob/4081dcdf05bea9e6f47982e37099d176dcc2be45/src/main/java/org/cloudbees/jenkins/multibranch/freestyle/FreeStyleMultibranchProject.java#L152) implementations.

jglick:
Confirmed in https://github.com/jenkinsci/workflow-plugin/issues/291 that `isApplicable` works for this purpose. Surely that should be documented somewhere in `MultiBranchProjectDescriptor`?

mjdetullio:
Added docs

jglick:
:+1:

PR #27, opened 2016-02-03
Make BranchSource.strategy use @DataBoundSetter

jglick:
@reviewbybees esp. @stephenc

### DIFF
--- a/src/main/java/jenkins/branch/BranchSource.java
+++ b/src/main/java/jenkins/branch/BranchSource.java
@@ -32,6 +32,7 @@ import jenkins.scm.api.SCMSourceDescriptor;
 import org.kohsuke.stapler.DataBoundConstructor;
 
 import java.util.List;
+import org.kohsuke.stapler.DataBoundSetter;
 
 /**
  * A source of branches, which consists of a source and a strategy for creating properties of the branches from this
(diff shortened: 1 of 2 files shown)


src/main/java/jenkins/branch/BranchSource.java
```diff
@@ -78,7 +77,12 @@ public SCMSource getSource() {
      * @return the strategy.
      */
     public BranchPropertyStrategy getStrategy() {
-        return strategy;
+        return strategy != null ? strategy : new DefaultBranchPropertyStrategy(new BranchProperty[0]);
```
andresrc: As `new DefaultBranchPropertyStrategy(new BranchProperty[0])` is effectively constant, couldn't we make it to `private static final` and use it as a default value?

andresrc:
:bee:

stephenc:
Not happy about removing the immutability of this class, but I understand the need :bee:

src/main/java/jenkins/branch/BranchSource.java
jglick: Could. The number of instances in the VM is probably only a handful so I saw no need to bother.

jglick:
> Not happy about removing the immutability of this class

I suppose in the future we could define a `@DataBoundBuilder`, if there were some standard implementation of the builder pattern to attach it to.

PR #28, opened 2016-02-18
[JENKINS-32749] Prevent NPE during unserialization of BranchSource

recena:
[JENKINS-32749](https://issues.jenkins-ci.org/browse/JENKINS-32749)

This PR prevents `NPE` during unserialization in case that a `BranchSource` has a null `SCMSource`. It can happen if an implementation of `SCMSource` is not present (disappears).

@reviewbybees specially @jglick, @amuniz

### DIFF
--- a/src/main/java/jenkins/branch/BranchSource.java
+++ b/src/main/java/jenkins/branch/BranchSource.java
@@ -34,6 +34,8 @@ import org.kohsuke.stapler.DataBoundConstructor;
 import java.util.List;
 import org.kohsuke.stapler.DataBoundSetter;
 
+import javax.annotation.Nonnull;
+
 /**
  * A source of branches, which consists of a source and a strategy for creating properties of the branches from this
  * source.
@@ -44,6 +46,7 @@ public class BranchSource extends AbstractDescribableImpl<BranchSource> {
     /**
      * The source.
      */
+    @Nonnull
     private final SCMSource source;
 
     /**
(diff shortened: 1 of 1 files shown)


amuniz:
Did you test that the `OldDataMonitor` do the right thing when dismissing old data?

recena:
@amuniz Sure! Anyway, I'll do some additional tests.

jglick:
:bee: assuming there is a test somewhere proving this works, probably using `@LocalData`.

recena:
@jglick Sure, I'm working on this test.

recena:
@stephenc Do you think we can cut a release with this PR? /cc @jglick

src/main/java/jenkins/branch/BranchSource.java
```diff
@@ -34,6 +34,8 @@
 import java.util.List;
 import org.kohsuke.stapler.DataBoundSetter;
 
+import javax.annotation.Nonnull;
```
stephenc: Prefer `edu.umd.cs.findbugs.annotations` over `javax.annotation` as:

1. https://docs.oracle.com/javase/7/docs/api/javax/annotation/package-summary.html does not have `Nonnull`
2. You are not supposed to use the `javax` package space outside of the supervision of a JSR
3. JSR-305 has not made any releases, therefore any artifacts *claiming* to be produced by JSR-305 are just rumours.
4. We do not even know when/if JSR-305 is revived whether they will pick `Nonnull` or `NonNull` or what semantics they will pick for it, so we risk LinkageErrors by using it.
5. The only reason to use `javax.annotation.Nonnull` is if overriding a method interface from Jenkins core that has incorrectly used `javax.annotation.Nonnull`.

As this is an internal usage only, this should - as plugin maintainer - be `edu.umd.cs.findbugs.annotations.NonNull`

stephenc:
I'd prefer if you could avoid using `javax.annotation` package classes that are not either part of the JVM or actually confirmed as released by a JSR. [JSR-305](https://jcp.org/en/jsr/detail?id=305) has not released any actual files and the only documentation on the JSR page indicates that they would use `NonNull` as opposed to `Nonnull`. IMHO without an official statement from the JSR the use of any annotations in the `javax.annotation` package is just wrong.

recena:
@stephenc I forgot your preferences about this. I agree with you. Done.

PR #29, opened 2016-03-08
[JENKINS-33309] Use ChangeRequestAction

jglick:
Downstream of https://github.com/jenkinsci/scm-api-plugin/pull/7.

Can use https://github.com/cloudbeers/PR-demo/pull/9 as a demo.

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/branch/BranchNameContributor.java
+++ b/src/main/java/jenkins/branch/BranchNameContributor.java
@@ -31,9 +31,13 @@ import hudson.model.ItemGroup;
 import hudson.model.Job;
 import hudson.model.TaskListener;
 import java.io.IOException;
+import java.net.URL;
+import jenkins.scm.api.SCMHead;
+import jenkins.scm.api.actions.ChangeRequestAction;
 
 /**
  * Defines the environment variable {@code BRANCH_NAME} for multibranch builds.
+ * Also defines {@code CHANGE_*} variables when {@link ChangeRequestAction} is present.
  */
 @Extension
 public class BranchNameContributor extends EnvironmentContributor {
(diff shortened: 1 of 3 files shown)


src/main/java/jenkins/branch/BranchNameContributor.java
```diff
@@ -46,9 +50,26 @@ public void buildEnvironmentFor(Job j, EnvVars envs, TaskListener listener) thro
         if (parent instanceof MultiBranchProject) {
             BranchProjectFactory projectFactory = ((MultiBranchProject) parent).getProjectFactory();
             if (projectFactory.isProject(j)) {
+                SCMHead head = projectFactory.getBranch(j).getHead();
                 // Note: not using Branch.name, since in the future that could be something different
                 // than SCMHead.name, which is what we really want here.
-                envs.put("BRANCH_NAME", projectFactory.getBranch(j).getHead().getName());
+                envs.put("BRANCH_NAME", head.getName());
+                ChangeRequestAction cr = head.getAction(ChangeRequestAction.class);
```
amuniz: Who is adding this action to `head`?

amuniz: IOW where is the `TransientActionFactory` implementation (as per https://github.com/jenkinsci/scm-api-plugin/pull/7)?

jglick: The current implementation is in `github-branch-source` (see linked PR) and uses an override rather than `TransientActionFactory`.

amuniz: Ok. So there are currently two ways to add actions to an `SCMHead`: extend it and override `getAllActions` or add an implementation of `TransientActionFactory`, right?

jglick: Correct. Similar to the situation for `Actionable`s, where you can override `getActions`, etc., from the inside; or use `TransientActionFactory` from the outside. `Actionable` also exposes `addAction` from the outside, though this presumes that there is mutable persisted action storage, an imposition I did not really want to apply to `SCMHead` (besides the other API problems with `Actionable` mentioned in a linked JIRA ticket).

amuniz:
:bee:

src/main/java/jenkins/branch/BranchNameContributor.java
amuniz: Ok, got it.

jglick:
@reviewbybees done

(@stephenc merge if you agree)

PR #33, opened 2016-04-05
[JENKINS-33808] Support for Item categorization

recena:
[JENKINS-33808](https://issues.jenkins-ci.org/browse/JENKINS-33808)

downstream of https://github.com/jenkinsci/scm-api-plugin/pull/9

@reviewbybees especially @jglick

### DIFF
--- a/src/main/java/jenkins/branch/CustomOrganizationFolderDescriptor.java
+++ b/src/main/java/jenkins/branch/CustomOrganizationFolderDescriptor.java
@@ -71,7 +71,7 @@ public class CustomOrganizationFolderDescriptor extends TopLevelItemDescriptor {
     }
 
     /**
-     * Needed if it wants CustomOrganizationFolderDescriptor implementations are categorized in Jenkins 2.x.
+     * Needed if it wants SCMNavigatorDescriptor implementations are categorized in Jenkins 2.x.
      *
      * TODO: Override when the baseline is upgraded to 2.x
      *
@@ -82,18 +82,18 @@ public class CustomOrganizationFolderDescriptor extends TopLevelItemDescriptor {
     }
 
     /**
-     * Needed if it wants CustomOrganizationFolderDescriptor implementations are categorized in Jenkins 2.x.
+     * Needed if it wants SCMNavigatorDescriptor implementations are categorized in Jenkins 2.x.
      *
      * TODO: Override when the baseline is upgraded to 2.x
      *
      * @return A string it represents a ItemCategory identifier.
      */
     public String getCategoryId() {
-        return delegate.getCategoryId();
+        return "nested-projects";
     }
 
     /**
-     * Needed if it wants CustomOrganizationFolderDescriptor implementations are categorized in Jenkins 2.x.
+     * Needed if it wants SCMNavigatorDescriptor implementations are categorized in Jenkins 2.x.
      *
      * TODO: Override when the baseline is upgraded to 2.x
      *
--- a/src/main/java/jenkins/branch/CustomOrganizationFolderDescriptor.java
+++ b/src/main/java/jenkins/branch/CustomOrganizationFolderDescriptor.java
@@ -89,7 +89,7 @@ public class CustomOrganizationFolderDescriptor extends TopLevelItemDescriptor {
      * @return A string it represents a ItemCategory identifier.
      */
     public String getCategoryId() {
-        return "nested-projects";
+        return delegate.getCategoryId();
     }
 
     /**
--- a/src/main/java/jenkins/branch/CustomOrganizationFolderDescriptor.java
+++ b/src/main/java/jenkins/branch/CustomOrganizationFolderDescriptor.java
@@ -70,6 +70,39 @@ public class CustomOrganizationFolderDescriptor extends TopLevelItemDescriptor {
         return delegate.getDisplayName();
     }
 
+    /**
+     * Needed if it wants CustomOrganizationFolderDescriptor implementations are categorized in Jenkins 2.x.
+     *
+     * TODO: Override when the baseline is upgraded to 2.x
+     *
+     * @return A string with the Item description.
+     */
+    public String getDescription() {
+        return delegate.getDescription();
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


src/main/java/jenkins/branch/CustomOrganizationFolderDescriptor.java
```diff
@@ -70,6 +70,39 @@ public String getDisplayName() {
...
+
+    /**
+     * Needed if it wants CustomOrganizationFolderDescriptor implementations are categorized in Jenkins 2.x.
+     *
+     * TODO: Override when the baseline is upgraded to 2.x
+     *
+     * @return A string it represents a ItemCategory identifier.
+     */
+    public String getCategoryId() {
+        return delegate.getCategoryId();
```
jglick: What happens when this is null? (3×)

src/test/java/jenkins/branch/CustomOrganizationFolderDescriptorTest.java
```diff
@@ -116,17 +116,32 @@ public void insideFolder() throws Exception {
...
+            return "MockNavigator description";
+        }
+
+        public String getCategoryId() {
+            return "MockNavigator categoryId";
+        }
+
+        public String getIconFilePathPattern() {
+            return "MockNavigator iconfilepathpattern";
+        }
```
jglick: Should not be necessary since these methods should not have been made `abstract`.

src/main/java/jenkins/branch/CustomOrganizationFolderDescriptor.java
jglick: And would we not just want to return `nested-projects` by default, since by definition these will all fall into that category?

recena: It can not be `null` since we decided to do that [here](https://github.com/jenkinsci/scm-api-plugin/pull/9/commits/f5d41fd0238c01a920b1be8ab45444c91afa3b7c)

src/test/java/jenkins/branch/CustomOrganizationFolderDescriptorTest.java
recena: @jglick you were right [here](https://github.com/jenkinsci/scm-api-plugin/pull/9#discussion_r58587652) but not these changes here, but because of we can not force to use specific versions of `scm-api` and `branch-api`.

Maybe we have here another reason to include both dependencies directly in **github-branch-source-plugin**.

src/main/java/jenkins/branch/CustomOrganizationFolderDescriptor.java
```diff
@@ -59,17 +59,50 @@
...
+
+    /**
+     * Needed if it wants SCMNavigator implementations are categorized in Jenkins 2.x.
+     *
+     * TODO: Override when the baseline is upgraded to 2.x
+     *
+     * @return A string it represents a ItemCategory identifier.
+     */
+    public String getCategoryId() {
+        return "nested-projects";
```
jglick: :bug: surely you meant `delegate.getCategoryId()`?

src/test/java/jenkins/branch/CustomOrganizationFolderDescriptorTest.java
```diff
@@ -116,17 +116,21 @@ public void insideFolder() throws Exception {
         public void visitSources(SCMSourceObserver observer) throws IOException, InterruptedException {}
     }
     private static abstract class MockNavigatorDescriptor extends SCMNavigatorDescriptor {
+
```
jglick: ?

jglick: I guess you did not know to

    git checkout master -- src/test/java/jenkins/branch/CustomOrganizationFolderDescriptorTest.java

src/main/java/jenkins/branch/CustomOrganizationFolderDescriptor.java
recena: No, I've decided that is this method who defines the item category.

recena: Now, I understand better [this comment](https://github.com/jenkinsci/scm-api-plugin/pull/9#discussion_r58925376)

jglick: Then why does `SCMNavigatorDescriptor.getCategoryId()` exist?

src/test/java/jenkins/branch/CustomOrganizationFolderDescriptorTest.java
recena: Perhaps you are underestimating too :smile:

src/main/java/jenkins/branch/CustomOrganizationFolderDescriptor.java
recena: `CustomOrganizationFolderDescriptor` is used for `SCMNavigator` implementations and we have decided that this kind of Item would be **nested-projects**.

Other option is [this|https://github.com/jenkinsci/scm-api-plugin/pull/9#discussion_r58925376].

IMO both are right.

jglick: It makes no sense to do both. Either you hardcode the category here, and delete the method from `SCMNavigatorDescriptor`; or you keep the method in `SCMNavigatorDescriptor`, changing its default value to `nested-projects`, and make this method delegate.

recena: When I said *both*, I meant that two options.

src/main/java/jenkins/branch/CustomOrganizationFolderDescriptor.java
jglick: (resolved)

jglick:
:bee:

recena:
@jglick Upgraded `scm-api` dependency. This PR is ready to merge /cc @stephenc

jglick:
:bee:

PR #36, opened 2016-04-26
[JENKINS-30206] Add scm and properties fields to Branch#equals(Object)

mjdetullio:
https://issues.jenkins-ci.org/browse/JENKINS-34304


mjdetullio:
ping @stephenc @jglick

stephenc:
👍 AIUI

recena:
👍  A test would be great.

mjdetullio:
There is a test.

recena:
@mjdetullio Great. I don't know why GitHub does not show me the rest of the files modified.

recena:
@mjdetullio Or Did you squash the commits?

mjdetullio:
It has always been one commit

recena:
@mjdetullio Thanks for your feedback.

mjdetullio:
Good to merge?

jglick:
Needs merge with `master` I think.

The problem here is that `SCM` and `BranchProperty` implementations generally do not override `equals`/`hashCode` from `Object`. So I think `property.getBranch().equals(branch)` will generally be false and you could just call `property.setBranch(branch)` unconditionally.

In other words, the proposed code is not so much wrong as misleading: it is pretending to compare `Branch`es by their contents, but really it would be more honest to check pointer equality on the `scm` and `properties` fields, since _any_ GUI reconfiguration will recreate those objects with different hash codes.

jglick:
Why was this closed? We still need a fix.

mjdetullio:
You are right, `SCM#equals(Object)` is not defined and not a good way to compare.  However, using `properties.equals(Object)` will compare the containing elements in the list.

The only other update I can think of is using `SCM#getKey()` to do the comparison (https://github.com/jenkinsci/jenkins/blob/jenkins-2.6/core/src/main/java/hudson/scm/SCM.java#L429).

Unless there is another clear fix I will be dropping this PR in favor of updating my `BranchProjectFactory` implementations to unconditionally call `property.setBranch(branch)` as suggested.

jglick:
`SCM.getKey` would work well for the `scm` part, I think. Comparing `properties` fully would require existing `BranchProperty` implementations to start overriding `equals`/`hashCode`.

As I was trying to say earlier, we can afford to have some false negatives on `Branch.equals`—at worst we just get a gratuitous configuration save. I mainly objected to there being code which looked like it was doing something subtle and really was just doing pointer equality.

https://github.com/jenkinsci/workflow-multibranch-plugin/pull/2 or similar could serve as an integration test.

mjdetullio:
But, how deep do we go with defining equals?  For example `BuildRetentionBranchProperty` holds a `BuildDiscarder` configuration, which would have to go all the way into core.  Each `BranchProperty` implementation and each implementation of whatever settings it holds would have to be examined.

It would make more sense to always update the `Branch` and remove this logic: https://github.com/jenkinsci/branch-api-plugin/blob/branch-api-1.8/src/main/java/jenkins/branch/MultiBranchProject.java#L308

Always updating the `Branch` also prompts this bug however: https://issues.jenkins-ci.org/browse/JENKINS-30089 -- making the bug applicable to any plugin implementing the Branch API that unconditionally updates the `Branch`.

jglick:
> which would have to go all the way into core

True. There are hacks to compare XStream output, but it _is_ looking impractical to detect when a `Branch` has been changed.

jglick:
> It would make more sense to always update the `Branch` and remove this logic

I am taking that approach in https://github.com/jenkinsci/workflow-multibranch-plugin/pull/12.

PR #38, opened 2016-04-28
[JENKINS-34259] Some links are broken in the Welcome page

recena:
[JENKINS-34259](https://issues.jenkins-ci.org/browse/JENKINS-34259)

- [x] **People** link
- [x] **Build History** link
- [x] **Welcome Page** `http://localhost:8080/jenkins/job/je2/view/Welcome`
- [x] Remove **Edit View** option

jenkins-34259-1

@reviewbybees especially, @kzantow since he worked on `BaseEmptyView`.

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -554,7 +554,7 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
     @Override
     public View getPrimaryView() {
         if (getItems().isEmpty()) {
-            // when there's no branches to show, switch to the special welcome view
+            // when there are no branches nor pull requests to show, switch to the special welcome view
             return getWelcomeView();
         }
         return super.getPrimaryView();
@@ -565,6 +565,11 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
      */
     protected View getWelcomeView() {
         return new MultiBranchProjectEmptyView(this);
+    }
+
+    @Override
+    public View getView(String name) {
+        return getPrimaryView();
     }
 
     /**
--- a/src/main/java/jenkins/branch/OrganizationFolder.java
+++ b/src/main/java/jenkins/branch/OrganizationFolder.java
@@ -252,7 +252,8 @@ public final class OrganizationFolder extends ComputedFolder<MultiBranchProject<
     }
 
     /**
-     * Will create an specialized view when there are no repositories found, which contain a Jenkinsfile
+     * Will create an specialized view when there are no repositories or branches found, which contain a Jenkinsfile
+     * or other MARKER file.
      */
     @Override
     public View getPrimaryView() {
@@ -260,6 +261,11 @@ public final class OrganizationFolder extends ComputedFolder<MultiBranchProject<
             return new OrganizationFolderEmptyView(this);
         }
         return super.getPrimaryView();
+    }
+
+    @Override
+    public View getView(String name) {
+        return getPrimaryView();
     }
 
     @Extension
--- a/src/main/java/jenkins/branch/BaseEmptyView.java
+++ b/src/main/java/jenkins/branch/BaseEmptyView.java
@@ -33,6 +33,14 @@ public class BaseEmptyView extends View {
     }
 
     /**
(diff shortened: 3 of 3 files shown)


recena:
I'm not confident with the idea of **People** and **Build History** options are present in the **Welcome Page**.

kzantow:
:bee: AFAICT

andresrc:
:bee:

Vlatombe:
🐝

recena:
@reviewbybees done

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -567,6 +567,11 @@ protected View getWelcomeView() {
         return new MultiBranchProjectEmptyView(this);
     }
 
+    @Override
+    public View getView(String name) {
+        return getPrimaryView();
```
jglick: Uh, what? Unless I am missing something, this is a :bug: since it prevents you from defining alternate views.

jglick: Should I think be

```java
if (name.equals("Welcome")) {
    return getWelcomeView();
} else {
    return super.getView(name);
}
```

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -262,6 +263,11 @@ public View getPrimaryView() {
         return super.getPrimaryView();
     }
 
+    @Override
+    public View getView(String name) {
+        return getPrimaryView();
```
jglick: similar :bug: here

jglick:
> People and Build History options are present in the Welcome Page

More likely a core bug: these should be suppressed for empty views.

jglick:
I wonder if the `BaseEmptyView` functionality, and corresponding `getView`/`getPrimaryView` tricks, should not be lifted up into `ComputedFolder`. Just as `AllView` for editable folders has a custom index page that guides the user to create & add items, so it would make sense for any newly created (or misconfigured) computed folder to have some sort of standardized UI guiding the user toward the `configure` page.

src/main/java/jenkins/branch/MultiBranchProject.java
recena: I think it would make sense if it was possible to add custom view.

captura el 2016-04-29 a las 22 16 01

recena:
> More likely a core bug: these should be suppressed for empty views.

I think so. Maybe we need a new API similar to `isEditable()` for the rest of the hardcode (in `sidepanel.jelly`) options.

> I wonder if the BaseEmptyView functionality, and corresponding getView/getPrimaryView tricks, should not be lifted up into ComputedFolder. Just as AllView for editable folders has a custom index page that guides the user to create & add items, so it would make sense for any newly created (or misconfigured) computed folder to have some sort of standardized UI guiding the user toward the configure page.

I did not make any proposal here because I thought this was thought-out and meditated but I agree with the idea to re-think it.

src/main/java/jenkins/branch/MultiBranchProject.java
jglick: You can. There are problems ([JENKINS-32178](https://issues.jenkins-ci.org/browse/JENKINS-32178)), but this mistake is making it worse.

jglick:
> Maybe we need a new API

Do not think so; for the two links mentioned, these just obviously make no sense for a `View` with no `Item`s in it, so that requires no new API. Maybe there are other cases which do.

src/main/java/jenkins/branch/MultiBranchProject.java
recena: The behavior described in that Jira issue assumes the user indexed some branches or pull requests. In this case, you don't have a empty view.

recena:
> Do not think so; for the two links mentioned, these just obviously make no sense for a View with no Items in it, so that requires no new API. Maybe there are other cases which do.

I think it is needed. It would be helpful to have something similar to `isEditable()` for *People and Build History* options.

src/main/java/jenkins/branch/MultiBranchProject.java
recena: @jglick I'm sorry, now I understand your bug. Let me solve it.

PR #41, opened 2016-05-11
[JENKINS-33819] OrphanedItemStrategy is propagated to the Multibranch projects

recena:
[JENKINS-33819](https://issues.jenkins-ci.org/browse/JENKINS-33819)

downstream of https://github.com/jenkinsci/cloudbees-folder-plugin/pull/61

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/branch/OrganizationFolder.java
+++ b/src/main/java/jenkins/branch/OrganizationFolder.java
@@ -163,6 +163,7 @@ public final class OrganizationFolder extends ComputedFolder<MultiBranchProject<
                                 PersistedList<BranchSource> sourcesList = existing.getSourcesList();
                                 sourcesList.clear();
                                 sourcesList.addAll(createBranchSources());
+                                existing.setOrphanedItemStrategy(getOrphanedItemStrategy());
                                 existing.scheduleBuild();
                                 return;
                             }
@@ -181,6 +182,7 @@ public final class OrganizationFolder extends ComputedFolder<MultiBranchProject<
                                     continue;
                                 }
                                 if (project != null) {
+                                    project.setOrphanedItemStrategy(getOrphanedItemStrategy());
                                     project.getSourcesList().addAll(createBranchSources());
                                     try {
                                         project.addTrigger(new PeriodicFolderTrigger("1d"));
--- a/pom.xml
+++ b/pom.xml
@@ -91,7 +91,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>cloudbees-folder</artifactId>
-      <version>5.1</version>
+      <version>5.10-SNAPSHOT</version>
     </dependency>
     <!-- jenkins dependencies -->
     <!-- test dependencies -->
--- a/pom.xml
+++ b/pom.xml
@@ -91,7 +91,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>cloudbees-folder</artifactId>
-      <version>5.10-SNAPSHOT</version>
+      <version>5.10</version>
     </dependency>
     <!-- jenkins dependencies -->
     <!-- test dependencies -->


src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -115,6 +116,11 @@ protected MultiBranchProject(ItemGroup parent, String name) {
         init2();
     }
 
+    protected MultiBranchProject(ItemGroup parent, String name, OrphanedItemStrategy strategy) {
```
jglick: As in the upstream PR, this would be simpler using a setter.

src/main/java/jenkins/branch/MultiBranchProjectFactory.java
```diff
@@ -59,7 +60,8 @@
      * @return a new uninitialized multibranch project (do not configure its {@link MultiBranchProject#getSourcesList} or call {@link MultiBranchProject#onCreatedFromScratch}), or null if unrecognized
      */
     @CheckForNull
-    public abstract MultiBranchProject<?,?> createProject(@Nonnull ItemGroup<?> parent, @Nonnull String name, @Nonnull List<? extends SCMSource> scmSources, @Nonnull Map<String,Object> attributes, @Nonnull TaskListener listener) throws IOException, InterruptedException;
+    public abstract MultiBranchProject<?,?> createProject(@Nonnull ItemGroup<?> parent, @Nonnull String name, @Nonnull List<? extends SCMSource> scmSources, @Nonnull OrphanedItemStrategy strategy,
+            @Nonnull Map<String,Object> attributes, @Nonnull TaskListener listener) throws IOException, InterruptedException;
```
jglick: :bug: this is not binary-compatible.

Again it would be simpler to have a setter for the strategy. Then no change to `MultiBranchProjectFactory` implementations would be required at all, unless we specifically wanted to customize the behavior on a per-factory basis (which I do not foresee a need for).

src/main/java/jenkins/branch/MultiBranchProjectFactory.java
```diff
@@ -92,10 +94,12 @@ public MultiBranchProjectFactoryDescriptor getDescriptor() {
          * @return a new project suitable for {@link #createProject}
          */
         @Nonnull
-        protected abstract MultiBranchProject<?,?> doCreateProject(@Nonnull ItemGroup<?> parent, @Nonnull String name, @Nonnull Map<String,Object> attributes);
+        protected abstract MultiBranchProject<?,?> doCreateProject(@Nonnull ItemGroup<?> parent, @Nonnull String name, @Nonnull OrphanedItemStrategy strategy, @Nonnull Map<String,Object> attributes);
```
jglick: :bug: not binary-compatible

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -173,7 +173,8 @@ public void complete() throws IllegalStateException, InterruptedException {
                             for (MultiBranchProjectFactory factory : projectFactories) {
                                 MultiBranchProject<?, ?> project;
                                 try {
-                                    project = factory.createProject(OrganizationFolder.this, projectName, sources, Collections.<String,Object>emptyMap(), listener);
+                                    project = factory.createProject(OrganizationFolder.this, projectName, sources, getOrphanedItemStrategy(),
```
jglick: :ant: as noted in JIRA, there is no reason to assume that a strategy which is appropriate for _repositories_ is also appropriate for _branches_. You might be forced to retain repositories you did not really want, just in order to avoid losing branches you did want.

src/main/java/jenkins/branch/MultiBranchProjectFactory.java
```diff
@@ -59,7 +60,8 @@
      * @return a new uninitialized multibranch project (do not configure its {@link MultiBranchProject#getSourcesList} or call {@link MultiBranchProject#onCreatedFromScratch}), or null if unrecognized
      */
     @CheckForNull
-    public abstract MultiBranchProject<?,?> createProject(@Nonnull ItemGroup<?> parent, @Nonnull String name, @Nonnull List<? extends SCMSource> scmSources, @Nonnull Map<String,Object> attributes, @Nonnull TaskListener listener) throws IOException, InterruptedException;
+    public abstract MultiBranchProject<?,?> createProject(@Nonnull ItemGroup<?> parent, @Nonnull String name, @Nonnull List<? extends SCMSource> scmSources, @Nonnull OrphanedItemStrategy strategy,
```
amuniz: I think backward compatibility should be kept here

src/main/java/jenkins/branch/MultiBranchProjectFactory.java
amuniz: Backward compatibility here too.

recena:
Well, binary compatibility issues are known, this was a way to start the review and I know this PoV https://github.com/jenkinsci/workflow-multibranch-plugin/pull/8#issuecomment-218466658

src/main/java/jenkins/branch/OrganizationFolder.java
recena: The idea was to resolve an urgent problem I found in several production environment. In fact, I started few day ago other approach based on provide a custom strategy for _branches_ and _repositories_ as you proposed in the Jira issue.

jglick: > The idea was to resolve an urgent problem I found in several production environment.

The change involves API additions, if not settings format changes, so it is not eligible for a hotfix. It has to be done right the first time.

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -174,6 +174,7 @@ public void complete() throws IllegalStateException, InterruptedException {
                                 MultiBranchProject<?, ?> project;
                                 try {
                                     project = factory.createProject(OrganizationFolder.this, projectName, sources, Collections.<String,Object>emptyMap(), listener);
+                                    project.setOrphanedItemStrategy(getOrphanedItemStrategy());
```
recena: Propagated when the project is created. We have to do the same when the `OrganizationFolder` is updated, _propagate again_.

jglick: Probably before line 166.

recena: @jglick I've moved this assignment because `project` can be `null.

jglick:
:bee:, good enough for now

amuniz:
I think this is going to fail following this steps:

1. Upgrade this plugin (including this change).
1. Set the orphaned item strategy to keep items for 1000 days (forever basically).
1. In a repository with only one branch, and a Jenkinsfile on it -already detected in a previous indexing-, remove the Jenkinsfile from the branch.
1. Run a re-index

What I would expect: nothing is removed
What I think is going to happen: the branch is going to be removed, as the strategy is not going to be set on jobs not `observe`d by the `SCMSource`.

It would be great If can do a manual test to see if I'm wrong or not.

recena:
@amuniz This PR propagates the `OrphanedItemStrategy` when a new `OrganizationFolder` is **created** or **updated**. This was my initial intention. In your step by step you are including saving the job configuration. It should work fine. At any rate, let me try it.

amuniz:
> In your step by step you are including saving the job configuration. It should work fine.

AFAICT this patch is not going to be executed on save time, but indexing. 

If you are going to check it, it's important the have **only one branch** in the repo with the Jenkinsfile.

recena:
@amuniz After doing several tests it seems this PR is working as expected.

If you configure:

1. Days to keep old items: **1000**
1. Max # of old items to keep: **0**

The orphaned jobs will be removed because both conditions are being verified. it is a _AND_ not **OR**. This should be clarified in the UI.

recena:
@amuniz As you see in my shared screen, this was the expected behavior. If you agree, we consider this PR as valid.

amuniz:
Ok. So I guess there is a bug somewhere else that needs file and fix.

:bee:

recena:
@amuniz After to dive into the source source it seems there is no a bug.

If you want to use only one of the conditions, you have to leave **empty** the other one or using **-1**.

recena:
@reviewbybees done

PR #1, opened 2014-09-12
[JENKINS-32396] Make triggering per-branch projects optional 

mattmoor:
Add an advanced option to allow users to keep branch indexing from triggering the projects it creates/updates.

PR #42, opened 2016-05-12
Switched from SCMTrigger.SCMTriggerCause to BranchIndexingCause

jglick:
`SCMTriggerCause` is not really appropriate for multibranch projects. For example, it tries to save a polling log, which in fact has bogus text. And the hyperlink is not useful.

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -44,7 +44,6 @@ import hudson.model.View;
 import hudson.scm.PollingResult;
 import hudson.security.ACL;
 import hudson.security.Permission;
-import hudson.triggers.SCMTrigger;
 import hudson.util.PersistedList;
 import jenkins.scm.api.SCMHead;
 import jenkins.scm.api.SCMHeadObserver;
@@ -282,8 +281,7 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
      * {@inheritDoc}
      */
     public void onSCMSourceUpdated(@NonNull SCMSource source) {
-        SCMTrigger.SCMTriggerCause cause = new SCMTrigger.SCMTriggerCause(source.getDescriptor().getDisplayName());
-        scheduleBuild(0, cause);
+        scheduleBuild(0, new BranchIndexingCause());
     }
 
     /**
@@ -363,7 +361,7 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
     }
 
     private void scheduleBuild(BranchProjectFactory<P,R> factory, final P item, SCMRevision revision, TaskListener listener, String name) {
-        if (ParameterizedJobMixIn.scheduleBuild2(item, 0, new CauseAction(new SCMTrigger.SCMTriggerCause("Branch indexing"))) != null) {
+        if (ParameterizedJobMixIn.scheduleBuild2(item, 0, new CauseAction(new BranchIndexingCause())) != null) {
             listener.getLogger().println("Scheduled build for branch: " + name);
             try {
                 factory.setRevisionHash(item, revision);
@@ -371,7 +369,7 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
                 e.printStackTrace(listener.error("Could not update last revision hash"));
             }
         } else {
-            listener.getLogger().println("Failed to schedule build for branch: " + name);
+            listener.getLogger().println("Did not schedule build for branch: " + name);
         }
     }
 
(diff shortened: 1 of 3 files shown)


recena:
I like the idea.

src/main/resources/jenkins/branch/BranchIndexingCause/description.jelly
```diff
@@ -0,0 +1,37 @@
...
+
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+    <j:set var="url" value="${it.indexingUrl}"/>
+    <j:choose>
+        <j:when test="${url != null}">
+            <a href="${rootURL}/${url}console">Branch indexing</a>
+        </j:when>
+        <j:otherwise>
+            Branch indexing
```
recena: Does it make sense to show in the left-menu a entry without link?

jglick: This is not in the sidebar.

recena: Where is used this _part of_ view? or How is used?

jglick: Read the Javadoc for `Cause`.

kwhetstone:
:bee: Makes sense; do we need some tests for this?

jglick:
Currently it is not used at any API level, just a minor UI feature, but I am looking into using it from a variant of #1, which would get test coverage.

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -363,15 +361,15 @@ public void observe(@NonNull SCMHead head, @NonNull SCMRevision revision) {
...
+        if (ParameterizedJobMixIn.scheduleBuild2(item, 0, new CauseAction(new BranchIndexingCause())) != null) {
             listener.getLogger().println("Scheduled build for branch: " + name);
             try {
                 factory.setRevisionHash(item, revision);
             } catch (IOException e) {
                 e.printStackTrace(listener.error("Could not update last revision hash"));
             }
         } else {
-            listener.getLogger().println("Failed to schedule build for branch: " + name);
+            listener.getLogger().println("Did not schedule build for branch: " + name);
```
jtnord: why the change?
to me "did not" means there was nothing to do whereas "failed" is I tried but couldn't.
"Failed" seems more appropriate here (Jenkins was shutting down perhaps?)

jglick: In the context of a possible alternative to #1, this message is more accurate.

A null return value could mean several things (unfortunately `ScheduleResult.Refused` does not differentiate them); the most innocuous would be a `QueueDecisionHandler` that returns `false` from `shouldSchedule` (again, unfortunately, it is not given a `TaskListener` to explain its decision, or offered to return a `CauseOfBlockage` or similar).

jtnord: thanks.

jtnord:
:bee:

src/main/resources/jenkins/branch/BranchIndexingCause/description.jelly
recena: @jglick From the Javadoc in `Cause`...

>  Cause object base class.  This class hierarchy is used to keep track of why a given build was started. This object encapsulates the UI rendering of the cause, as well as providing more useful information in respective subypes.

I don't still know who Jelly file includes to this one. Could you help me?

jglick: > From the Javadoc in `Cause`...

Read a little longer! :-)

> <h2>Views</h2>
> <dl>
> <dt>description.jelly
> <dd>Renders the cause to HTML. By default, it puts the short description.
> </dl>

So…

> I don't still know who Jelly file includes to this one.

`CauseAction/summary.jelly`

recena: @jglick If I'm asking is because it was not clear to me ;(

recena: Ah ok, this jelly file replaces to this one https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/model/Cause/description.jelly.

Thanks.

recena:
Belated 🐝

src/main/resources/jenkins/branch/BranchIndexingCause/description.jelly
jglick: Yes, the default prints plain text, but in this we would like a hyperlink.

src/main/java/jenkins/branch/MultiBranchProject.java
jglick: [JENKINS-35403](https://issues.jenkins-ci.org/browse/JENKINS-35403)

PR #32, opened 2016-03-15
[JENKINS-32670] Multibranch Pipelines have no applicable branch properties

jglick:
[JENKINS-32670](https://issues.jenkins-ci.org/browse/JENKINS-32670)

So, do not show any of the controls for setting branch properties (unless some other plugin adds some which do not get excluded).

@reviewbybees

PR #43, opened 2016-05-13
[JENKINS-32396] Option to suppress automatic SCM trigger

jglick:
[JENKINS-32396](https://issues.jenkins-ci.org/browse/JENKINS-32396)

Supersedes #1, from which it mostly borrows the branch property system (though with a different mechanism for actually suppressing the build).

branch-sources

I debated various approaches for extending the feature to organization folders and could not come up with a very satisfactory answer, so I went with this relatively straightforward system of using a folder property, which seems to work well enough.

org-folder

The question is whether there is a more general need for defining branch property strategies at the organization folder level. For example, you could imagine a `BranchPropertyStrategyFactory` bound to an `OrganizationFolder` which would define a `BranchPropertyStrategy` on each `BranchSource` in each `MultiBranchProject` child. A simple implementation could simply embed a `BranchPropertyStrategy` which it would apply to all children. This would not be hard to write. The problem is the GUI configuration: without rather complicated API changes (which other plugins would then need to implement), there is no way for this section in the configuration form of the `OrganizationFolder` to know which kinds of `MultiBranchProject`s and `SCMSource`s it might encounter after indexing. Thus it would have to show all `BranchPropertyDescriptor`s. This could be quite confusing; for example, if you had only `workflow-multibranch` installed and were expecting to use only `WorkflowMultiBranchProject`s, which previously accepted _no_ `BranchProperty`s and now will accept only `NoTriggerBranchProperty`, at the organization folder level you would still be prompted to configure things like `UntrustedBranchProperty` which will not work with multibranch Pipelines, as well as things like `BuildRetentionBranchProperty` which are not wanted (supplanted by the `properties` step).

Originally I considered using [JENKINS-34005](https://issues.jenkins-ci.org/browse/JENKINS-34005) and retrofitting the branch indexing trigger to be a true `Trigger`:

```groovy
properties([[$class: 'TriggersProperty', triggers: [[$class: 'BranchIndexingTrigger']]]])
```

or similar. But then every `Jenkinsfile` which you would want to build on commits would need to opt in, which seems weird (and would be incompatible). And it would be incompatible for other multibranch project types without some extra effort. Worst, the *first* build would not get run automatically for any given branch, which may be only an annoyance for `master` but would be a disaster for `PR-1234`.

Another option would be a `JobProperty` that suppresses the branch trigger, rather than a branch property, which would solve the default behavior issue and let you write something like

```groovy
if (!env.BRANCH_NAME.matches('release.*')) {
  properties([[$class: 'NoBranchIndexingTrigger']])
}
```

But this would suffer from the reverse first build problem: *all* branch projects would automatically run the first build, and the above snippet would only disable running *subsequent* builds upon commit. For some use cases this would be OK, but some users legitimately want to completely suppress all automatic builds of certain branches.

@reviewbybees esp. @stephenc

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,6 +113,29 @@
       <version>1.24</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>git</artifactId>
+      <version>2.4.4</version>
+      <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.httpcomponents</groupId>
+          <artifactId>httpcore</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.apache.httpcomponents</groupId>
+          <artifactId>httpclient</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency> <!-- TODO for GitSampleRepoRule; should this be pushed up into scm-api, or git, or jenkins-test-harness? -->
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-scm-step</artifactId>
+      <version>2.0</version>
+      <classifier>tests</classifier>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
 </project>
--- a/src/main/java/jenkins/branch/BranchIndexingCause.java
+++ b/src/main/java/jenkins/branch/BranchIndexingCause.java
@@ -42,7 +42,7 @@ public final class BranchIndexingCause extends Cause {
     }
 
     /**
-     * @inheritDoc
+     * {@inheritDoc}
      */
     @Override
     public void onAddedTo(Run build) {
@@ -53,7 +53,7 @@ public final class BranchIndexingCause extends Cause {
     }
 
     /**
-     * @inheritDoc
+     * {@inheritDoc}
      */
     @Override
     public void onLoad(Run<?,?> build) {
@@ -76,7 +76,7 @@ public final class BranchIndexingCause extends Cause {
     }
 
     /**
-     * @inheritDoc
+     * {@inheritDoc}
      */
     @Override
     public String getShortDescription() {
--- a/src/main/java/jenkins/branch/NoTriggerBranchProperty.java
+++ b/src/main/java/jenkins/branch/NoTriggerBranchProperty.java
@@ -1,0 +1,92 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2016 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package jenkins.branch;
+
+import hudson.Extension;
+import hudson.model.Action;
+import hudson.model.Cause;
+import hudson.model.CauseAction;
+import hudson.model.Job;
+import hudson.model.Queue;
+import hudson.model.Run;
+import java.util.List;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+/**
+ * Suppresses builds due to {@link BranchIndexingCause}.
+ */
+@Restricted(NoExternalUse.class)
+public class NoTriggerBranchProperty extends BranchProperty {
+
+    @DataBoundConstructor
+    public NoTriggerBranchProperty() {}
+
+    @Override
+    public <P extends Job<P, B>, B extends Run<P, B>> JobDecorator<P, B> jobDecorator(Class<P> clazz) {
+        return null;
+    }
+
+    @Extension
+    public static class DescriptorImpl extends BranchPropertyDescriptor {
+
+        @Override
+        public String getDisplayName() {
+            return Messages.NoTriggerBranchProperty_suppress_automatic_scm_triggering();
+        }
+
+    }
+
+    @Extension
+    public static class Dispatcher extends Queue.QueueDecisionHandler {
+
+        @SuppressWarnings({"unchecked", "rawtypes"}) // untypable
+        @Override
+        public boolean shouldSchedule(Queue.Task p, List<Action> actions) {
+            for (Action action :  actions) {
+                if (action instanceof CauseAction) {
+                    for (Cause c : ((CauseAction) action).getCauses()) {
+                        if (c instanceof BranchIndexingCause) {
+                            if (p instanceof Job) {
+                                Job j = (Job) p;
+                                if (j.getParent() instanceof MultiBranchProject) {
+                                    for (BranchProperty prop : ((MultiBranchProject) j.getParent()).getProjectFactory().getBranch(j).getProperties()) {
+                                        if (prop instanceof NoTriggerBranchProperty) {
+                                            return false;
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+            return true;
+        }
+
+    }
+
+}
--- a/src/main/java/jenkins/branch/NoTriggerOrganizationFolderProperty.java
+++ b/src/main/java/jenkins/branch/NoTriggerOrganizationFolderProperty.java
@@ -1,0 +1,116 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2016 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package jenkins.branch;
+
+import com.cloudbees.hudson.plugins.folder.AbstractFolderProperty;
+import com.cloudbees.hudson.plugins.folder.AbstractFolderPropertyDescriptor;
+import hudson.Extension;
+import hudson.model.Action;
+import hudson.model.Cause;
+import hudson.model.CauseAction;
+import hudson.model.Job;
+import hudson.model.Queue;
+import hudson.util.FormValidation;
+import java.util.List;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.QueryParameter;
+
+/**
+ * Defines {@link NoTriggerBranchProperty} on selected branches.
+ */
+@Restricted(NoExternalUse.class)
+public class NoTriggerOrganizationFolderProperty extends AbstractFolderProperty<OrganizationFolder> {
+
+    /** Regexp of branch names to trigger. */
+    private final String branches;
+
+    @DataBoundConstructor
+    public NoTriggerOrganizationFolderProperty(String branches) {
+        this.branches = branches;
+    }
+    
+    public String getBranches() {
+        return branches;
+    }
+
+    @Extension
+    public static class DescriptorImpl extends AbstractFolderPropertyDescriptor {
+
+        @Override
+        public String getDisplayName() {
+            return Messages.NoTriggerBranchProperty_suppress_automatic_scm_triggering();
+        }
+
+        public FormValidation doCheckBranches(@QueryParameter String value) {
+            try {
+                Pattern.compile(value);
+                return FormValidation.ok();
+            } catch (PatternSyntaxException x) {
+                return FormValidation.error(x.getMessage());
+            }
+        }
+
+    }
+
+    @Extension
+    public static class Dispatcher extends Queue.QueueDecisionHandler {
+
+        @SuppressWarnings({"unchecked", "rawtypes"}) // untypable
+        @Override
+        public boolean shouldSchedule(Queue.Task p, List<Action> actions) {
+            for (Action action :  actions) {
+                if (action instanceof CauseAction) {
+                    for (Cause c : ((CauseAction) action).getCauses()) {
+                        if (c instanceof BranchIndexingCause) {
+                            if (p instanceof Job) {
+                                Job j = (Job) p;
+                                if (j.getParent() instanceof MultiBranchProject) {
+                                    MultiBranchProject mbp = (MultiBranchProject) j.getParent();
+                                    if (mbp.getParent() instanceof OrganizationFolder) {
+                                        NoTriggerOrganizationFolderProperty prop = ((OrganizationFolder) mbp.getParent()).getProperties().get(NoTriggerOrganizationFolderProperty.class);
+                                        if (prop != null) {
+                                            // Not necessarily the same as j.getName(), which may be encoded:
+                                            String name = mbp.getProjectFactory().getBranch(j).getName();
+                                            if (!name.matches(prop.getBranches())) {
+                                                return false;
+                                            }
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+            return true;
+        }
+
+    }
+
+}
--- a/src/main/java/jenkins/branch/OrganizationFolder.java
+++ b/src/main/java/jenkins/branch/OrganizationFolder.java
@@ -146,7 +146,7 @@ public final class OrganizationFolder extends ComputedFolder<MultiBranchProject<
                             }
                             List<BranchSource> branchSources = new ArrayList<BranchSource>();
                             for (SCMSource source : sources) {
-                                // TODO we probably need to define a BranchPropertyStrategyFactory (cf. JENKINS-22242)
+                                // TODO do we want/need a more general BranchPropertyStrategyFactory?
                                 branchSources.add(new BranchSource(source));
                             }
                             sources = null; // make sure complete gets called just once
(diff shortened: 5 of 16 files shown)


src/main/java/jenkins/branch/BranchIndexingCause.java
```diff
@@ -42,7 +42,7 @@
     }
 
     /**
-     * @inheritDoc
+     * {@inheritDoc}
```
jglick: Left over from #42 but was blocking me from trying `mvn clean install` in this branch. Note that `BranchIndexingCause` is used by the newly added code.

src/main/java/jenkins/branch/NoTriggerOrganizationFolderProperty.java
```diff
@@ -0,0 +1,116 @@
...
+    public String getBranches() {
+        return branches;
+    }
+
+    @Extension
+    public static class DescriptorImpl extends AbstractFolderPropertyDescriptor {
+
+        @Override
+        public String getDisplayName() {
+            return Messages.NoTriggerBranchProperty_suppress_automatic_scm_triggering();
```
jglick: Does not appear in UI, since `config` views of folder properties are rendered inline, just like `JobProperty`.

src/main/resources/jenkins/branch/NoTriggerOrganizationFolderProperty/help-branches.html
```diff
@@ -0,0 +1,6 @@
+<div>
+    Allows you to control the SCM commit trigger coming from branch indexing.
+    Supply a regular expression of branch names, for example <code>(?!release.*).*</code> or <code>PR-\d+</code>.
```
jglick: Yes this is different from the syntax of `NamedExceptionsBranchPropertyStrategy.Named.name`, but regular expressions are more flexible, especially in this case where you may need to create a pattern in either the positive or negative sense.

src/test/java/jenkins/branch/NoTriggerOrganizationFolderPropertyTest.java
```diff
@@ -0,0 +1,115 @@
...
+        assertEquals(3, master.getNextBuildNumber());
+        assertEquals(1, release.getNextBuildNumber());
+        assertEquals(3, newfeature.getNextBuildNumber());
+        QueueTaskFuture<FreeStyleBuild> releaseBuild = release.scheduleBuild2(0);
+        assertNotNull("was able to schedule a manual build of the release branch", releaseBuild);
+        assertEquals(1, releaseBuild.get().getNumber());
+        assertEquals(2, release.getNextBuildNumber());
+    }
+
+    @TestExtension
```
jglick: A `@TestExtension` is scoped to a test suite, so this needed to be copied.

src/test/java/jenkins/branch/harness/BranchProjectFactoryImpl.java
```diff
@@ -0,0 +1,85 @@
...
+import jenkins.branch.Branch;
+import jenkins.branch.BranchProjectFactory;
+import jenkins.branch.BranchProjectFactoryDescriptor;
+import jenkins.branch.MultiBranchProject;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+public class BranchProjectFactoryImpl extends BranchProjectFactory<FreeStyleProject, FreeStyleBuild> {
+
+    @DataBoundConstructor
+    public BranchProjectFactoryImpl() {}
```
jglick: This and `config.jelly` were previously missing, preventing tests from using `configRoundtrip` on a `MultiBranchImpl`.

src/test/java/jenkins/branch/NoTriggerOrganizationFolderPropertyTest.java
```diff
@@ -0,0 +1,115 @@
...
+        sampleRepo.git("commit", "--all", "--message=newfeature");
+        sampleRepo.git("checkout", "-b", "release", "master");
+        sampleRepo.write("server", "mycorp.com");
+        sampleRepo.git("add", "server");
+        sampleRepo.git("commit", "--all", "--message=release");
+        OrganizationFolder top = r.jenkins.createProject(OrganizationFolder.class, "top");
+        r.configRoundtrip(top);
+        NoTriggerOrganizationFolderProperty prop = top.getProperties().get(NoTriggerOrganizationFolderProperty.class);
+        assertNotNull(prop);
+        assertEquals(".*", prop.getBranches());
```
jglick: Like `JobProperty`s (which are not `OptionalJobProperty`), `AbstractFolderProperty`s are saved by default in `configSubmit`, so we just make sure the default configuration settings produce behavior matching the previous state.

src/test/java/jenkins/branch/NoTriggerBranchPropertyTest.java
```diff
@@ -0,0 +1,107 @@
...
+        assertEquals(3, master.getNextBuildNumber());
+        assertEquals(1, release.getNextBuildNumber());
+        assertEquals(3, newfeature.getNextBuildNumber());
+        QueueTaskFuture<FreeStyleBuild> releaseBuild = release.scheduleBuild2(0);
+        assertNotNull("was able to schedule a manual build of the release branch", releaseBuild);
+        assertEquals(1, releaseBuild.get().getNumber());
+        assertEquals(2, release.getNextBuildNumber());
+    }
+
+    static void showComputation(@NonNull ComputedFolder<?> d) throws Exception {
```
jglick: Copied from `workflow-multibranch`. Could potentially be added to `ComputedFolder` itself, though it is useful only from tests.

src/test/java/jenkins/branch/NoTriggerBranchPropertyTest.java
```diff
@@ -0,0 +1,107 @@
...
+        sampleRepo.git("add", "stuff");
+        sampleRepo.git("commit", "--all", "--message=newfeature");
+        sampleRepo.git("checkout", "-b", "release", "master");
+        sampleRepo.write("server", "mycorp.com");
+        sampleRepo.git("add", "server");
+        sampleRepo.git("commit", "--all", "--message=release");
+        MultiBranchImpl stuff = r.jenkins.createProject(MultiBranchImpl.class, "stuff");
+        BranchSource branchSource = new BranchSource(new GitSCMSource(null, sampleRepo.toString(), "", "*", "", false));
+        branchSource.setStrategy(new NamedExceptionsBranchPropertyStrategy(new BranchProperty[0], new NamedExceptionsBranchPropertyStrategy.Named[] {
+            new NamedExceptionsBranchPropertyStrategy.Named("release*", new BranchProperty[] {new NoTriggerBranchProperty()})
```
jglick: Makes slightly more sense when you see the configuration screen.

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -146,7 +146,7 @@ public void addSource(SCMSource source) {
                             }
                             List<BranchSource> branchSources = new ArrayList<BranchSource>();
                             for (SCMSource source : sources) {
-                                // TODO we probably need to define a BranchPropertyStrategyFactory (cf. JENKINS-22242)
```
jglick: JENKINS-22242 was actually something unrelated.

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -146,7 +146,7 @@ public void addSource(SCMSource source) {
...
+                                // TODO do we want/need a more general BranchPropertyStrategyFactory?
```
jglick: …see PR description.

src/main/java/jenkins/branch/NoTriggerOrganizationFolderProperty.java
```diff
@@ -0,0 +1,116 @@
...
+                    for (Cause c : ((CauseAction) action).getCauses()) {
+                        if (c instanceof BranchIndexingCause) {
+                            if (p instanceof Job) {
+                                Job j = (Job) p;
+                                if (j.getParent() instanceof MultiBranchProject) {
+                                    MultiBranchProject mbp = (MultiBranchProject) j.getParent();
+                                    if (mbp.getParent() instanceof OrganizationFolder) {
+                                        NoTriggerOrganizationFolderProperty prop = ((OrganizationFolder) mbp.getParent()).getProperties().get(NoTriggerOrganizationFolderProperty.class);
+                                        if (prop != null) {
+                                            // Not necessarily the same as j.getName(), which may be encoded:
```
jglick: cf. `WorkflowBranchProjectFactoryTest.slashyBranches`

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
```diff
@@ -0,0 +1,92 @@
...
+ */
+@Restricted(NoExternalUse.class)
+public class NoTriggerBranchProperty extends BranchProperty {
+
+    @DataBoundConstructor
+    public NoTriggerBranchProperty() {}
+
+    @Override
+    public <P extends Job<P, B>, B extends Run<P, B>> JobDecorator<P, B> jobDecorator(Class<P> clazz) {
+        return null;
```
jglick: Cf. `BranchProperty` patch; we could perhaps make this method optional to implement (returning null by default).

pom.xml
```diff
@@ -113,6 +113,29 @@
...
+      <version>2.4.4</version>
+      <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>org.apache.httpcomponents</groupId>
+          <artifactId>httpcore</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.apache.httpcomponents</groupId>
+          <artifactId>httpclient</artifactId>
```
jglick: Maven dependency tree madness…we want the newest dependency, not the “closest”.

pom.xml
```diff
@@ -113,6 +113,29 @@
       <version>1.24</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>git</artifactId>
+      <version>2.4.4</version>
+      <scope>test</scope>
```
jglick: `GitSCMSource`

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
```diff
@@ -0,0 +1,92 @@
...
+            for (Action action :  actions) {
+                if (action instanceof CauseAction) {
+                    for (Cause c : ((CauseAction) action).getCauses()) {
+                        if (c instanceof BranchIndexingCause) {
+                            if (p instanceof Job) {
+                                Job j = (Job) p;
+                                if (j.getParent() instanceof MultiBranchProject) {
+                                    for (BranchProperty prop : ((MultiBranchProject) j.getParent()).getProjectFactory().getBranch(j).getProperties()) {
+                                        if (prop instanceof NoTriggerBranchProperty) {
+                                            return false;
```
jglick: It would be a lot nicer if a `QueueDecisionHandler` could return a `CauseOfBlockage` that would be stored in `ScheduleResult.Refused` and thus accessible to callers like `MultiBranchProject.scheduleBuild` which have a `TaskListener` handy.

pom.xml
```diff
@@ -113,6 +113,29 @@
...
+          <groupId>org.apache.httpcomponents</groupId>
+          <artifactId>httpcore</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.apache.httpcomponents</groupId>
+          <artifactId>httpclient</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency> <!-- TODO for GitSampleRepoRule; should this be pushed up into scm-api, or git, or jenkins-test-harness? -->
```
amuniz: I guess git-plugin (tests) would be the right place.

amuniz:
IIUC when the no-trigger property is set, `shouldSchedule` will return always false for branch jobs, even when a reindex is done (on incoming webhooks in `github-branch-source`, for example). Is that what we want? Or just to suppress the first build when a branch is detected?

pom.xml
jglick: Perhaps. It will be about 1Mb, most of which will be irrelevant, and `GitSampleRepoRule` is not used by `git-plugin` itself, which would be odd too.

jglick:
> Is that what we want?

Yes, the usual request is to suppress all automatic builds for the branch—both the first, and any subsequent following SCM changes.

amuniz:
Then :bee:

stephenc:
I have no clue what is going on here and a wall of text is not making it clear to me. This is the best review of this code I can provide at the minute

jglick:
Fair enough!

i386:
UX here looks good to me 🐝

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
```diff
@@ -0,0 +1,92 @@
...
+        @SuppressWarnings({"unchecked", "rawtypes"}) // untypable
+        @Override
+        public boolean shouldSchedule(Queue.Task p, List<Action> actions) {
+            for (Action action :  actions) {
+                if (action instanceof CauseAction) {
+                    for (Cause c : ((CauseAction) action).getCauses()) {
+                        if (c instanceof BranchIndexingCause) {
+                            if (p instanceof Job) {
+                                Job j = (Job) p;
+                                if (j.getParent() instanceof MultiBranchProject) {
```
kzantow: NIT: consolidate some of these checks to avoid this giant nesting

michaelneale:
🐝 on feature concept

src/main/java/jenkins/branch/NoTriggerOrganizationFolderProperty.java
```diff
@@ -0,0 +1,116 @@
...
+        @SuppressWarnings({"unchecked", "rawtypes"}) // untypable
+        @Override
+        public boolean shouldSchedule(Queue.Task p, List<Action> actions) {
+            for (Action action :  actions) {
+                if (action instanceof CauseAction) {
+                    for (Cause c : ((CauseAction) action).getCauses()) {
+                        if (c instanceof BranchIndexingCause) {
+                            if (p instanceof Job) {
+                                Job j = (Job) p;
+                                if (j.getParent() instanceof MultiBranchProject) {
```
kzantow: Off topic, Java needs an language feature for this, like: `j.getParent() as MultibranchProject (mbp) => {`

kzantow:
Looks good, approach makes sense :bee:

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
jglick: Yeah, I go back and forth on whether the nesting is more readable, or using `return`/`break`/`continue`.

src/main/java/jenkins/branch/NoTriggerOrganizationFolderProperty.java
jglick: That would be nice! Except for the fact that you cannot generify a SAM interface based on exception list (`throws` clause), you could have

```java
public static <T,R> as(Object arg, Class<T> type, R fallback, Function<T,R> function) {
    if (type.isInstance(arg)) {
        return function(type.cast(arg));
    } else {
        return falback;
    }
}
```

used as

```java
if (as(j.getParent(), MultiBranchProject.class, false, (mbp) -> mbp.matchesSomeCondition())) {
    return false;
}
```

jglick:
Note that [this test](https://github.com/jenkinsci/workflow-multibranch-plugin/blob/651b56627a5e5b3df99d31ebd2248c7b69a9c41a/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowMultiBranchProjectTest.java#L137-L139) should be updated after releasing this PR to assert a single `NoTriggerBranchProperty`. Note also that this PR effectively obsoletes #32: now that there is one branch property applicable to Pipeline (without installing any further plugins), the branch property selection UI will always be displayed.

(I had been hoping that we could totally bypass the branch property system for Pipeline-as-code, but as argued in the PR description, this seems to be one of the cases that just cannot be sanely handled in `Jenkinsfile` itself, since it pertains to the behavior of the job _before_ the first build. The other corner case is defining parameters. As shown [here](https://github.com/jenkinsci/workflow-multibranch-plugin/blob/651b56627a5e5b3df99d31ebd2248c7b69a9c41a/src/test/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStepTest.java#L102), it can be handled in `Jenkinsfile`, though it is currently a little awkward.)

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
jglick: Also auto:ant: it would be better to do the `Job`/`MultiBranchProject` typecheck at the start of `shouldSchedule` so as to short-circuit this method for unrelated project types. Probably negligible overhead in the context of queue maintenance generally, but still.

recena: you are awesome!

jglick:
Regarding the crude UI, I would have preferred to have `NoTriggerOrganizationFolderProperty` appear in the configuration screen near _Build Triggers_; but implementing that would require a lot of haphazard API changes to `cloudbees-folder`, since traditional Jenkins configuration screens are organized by a page inclusion metaphor that forces the DOM to align with internal structure, and `AbstractFolderProperty`s and `Trigger`s are each kept in their own sections. Could perhaps work around this by reconceiving the property as a dummy trigger, but this could be confusing: everything else under the trigger section controls whether and when the organization folder will be reindexed, whereas this is something else entirely (related to the triggers of the folder’s _grandchildren_).

HRMPW:
This should cover the scenario described here: 
As a user I want to be able to specify whether a branch uses SCM triggers or not.
I have a project with the following SCM branches:

- master
- feature-1
- release-xyz

Let's say the requirements below apply to release* branches.
When I first set up the multi-branch project, Jobs should be created for master, feature-1 and release-xyz branches. A new build is triggered for master and feature-1 jobs

- When I create a new branch release-abc, the Job is created, but no build is triggered
- When I commit to existing feature-1 branch, a new build is triggered for the feature-1 branch
- When I commit to existing release-abc branch nothing happens

🐝

src/main/java/jenkins/branch/NoTriggerOrganizationFolderProperty.java
kzantow: True, of course I consider checked exceptions a bug in Java. But that's a whole different topic... sorry for polluting your PR with this nonsense. :)

recena: *I consider checked exceptions a bug in Java* ready? I need beers to discuss this one 😄

jglick:
@stephenc points out that at some point we may want the organization folder configuration to have a richer set of options, for example to suppress the trigger on a certain branch name pattern but only in repositories whose names also match a certain pattern. But he agrees that this could be done later without major compatibility issues in a variety of ways.

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
jglick: [https://issues.jenkins-ci.org/browse/JENKINS-35403](JENKINS-35403)

arnaud-deprez:
Hi,

I don't understand how this issue should solve https://issues.jenkins-ci.org/browse/JENKINS-32396 in any way.

The issue says: 

> There should be an option to create branch projects automatically wherever a Jenkinsfile is detected, yet *suppress the automatic SCM trigger which fires when branch indexing detects* a new branch head. Manual or *other automatic scheduled builds should still be possible*.

It seems that it is all or nothing: 
When I use `Suppress automatic SCM triggering` option, it disables the job trigger *during scanning AND on webhook*

What I would like is to have a way to not trigger jobs when scanning is executed, but trigger a job when a webhook is received (so a change).

Is there a way to do it with this ?

Cheers,

stephenc:
Scanning and webhook are the same thing. All that the webhook does is trigger a scan (although where possible we only do a narrow targeted scan on just the rumoured branch that was changed)

arnaud-deprez:
Hi @stephenc 
Thank you for your fast reply.

I understand that the current implementation consider it as a same thing but IMOH, it should not.
I have exactly the same issue than here: https://issues.jenkins-ci.org/browse/JENKINS-33909?focusedCommentId=334232&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-334232.

Basically, we use groovy script and some job dsl to configure the master and our jobs (that was before `configuration-as-code` plugin was released) on a kubernetes cluster.
So in order to guarantee that latest configuration remains and that no old configuration is kept, we clean the whole workspace to start from fresh.
And so when we start our latest instance with our latest change, jenkins start getting crazy and rebuild everything in the same time.

In short IMHO, there are 2 distincts needs and one should not rely on the other: 
- the scan is used to detect branch and configure jobs and eventually configure the webhook
- the webhook or the UI are used to trigger the jobs

I think the initial intent of this issue was to solve that (at least from the description, not the implementation), so either we need to reopen it or create a new issue but I would put exactly the same description.
I also think that this need will increase with the adoption of `configuration-as-code` plugin.

stephenc:
The specific impl added in this pr is imho an anti-pattern.

There is the branch build strategies that allow for more correct behavioural choices.

Ultimately the suppression added in this PR is slated for removal... I just am focused away from Jenkins plugins at present. TBD when the removal gets actioned

jglick:
> I think the initial intent of this issue was to solve [the two distinct needs]

No. As @stephenc wrote, webhooks are merely an optimization to avoid the need to wait for a full rescan. The behavior should be identical other than timing. @arnaud-deprez I think your issue comes from a misuse of `job-dsl`, which already supports deleting items that the latest script run did not ask to create.

> There is the branch build strategies

Maybe something added in SCM API 2.x after this PR?

arnaud-deprez:
@jglick thanks for your reply.
I'm interested in this feature `deleting items that the latest script run did not ask to create`. Where can I find this (I tried to google it a bit without success so far...) ?
This might solve my issue indeed :+1: 

In general, it would be nicer if jenkins would be a bit more stateless without getting crazy and only rely on its storage for a cache purpose, but that's out of the scope of this issue.

jglick:
> Where can I find this

IIRC this is a standard feature of the `job-dsl` plugin. But I neither work on that nor use it, so please consult the appropriate forum.

jcoyne:
Leaving a trail for anyone who gets here trying to figure this out. The approach outlined on this ticket is deprecated. See https://issues.jenkins-ci.org/browse/JENKINS-56231

jglick:
@jcoyne if you had to get this far, there is an unfiled UX buglet somewhere. Please consider taking a moment to file a PR to update some inline help, jenkins.io reference material, etc. pointing people to the current recommendations.

jcoyne:
@jglick I'd love to, but I still haven't figured it out yet. 😢

PR #44, opened 2016-06-08
[JENKINS-32253] replace slash symbols ('/') with double underscores ('__')

dbelyaev:
replace slash symbols ('/') with double underscores ('__') in branch names to fix problems with encoded slashes in pathname (%2F) that corrupt msbuild and some other command line tools

It is a quick workaround that worked for me. Not sure, if it follows development process, if not, please give me a link to guide.

### DIFF
--- a/src/main/java/jenkins/branch/Branch.java
+++ b/src/main/java/jenkins/branch/Branch.java
@@ -99,7 +99,15 @@ public class Branch {
      * @since 0.2-beta-7
      */
     public String getEncodedName() {
-        return Util.rawEncode(getName());
+        return this.encodeBranchName(getName());
+    }
+
+    /**
+     * Get a branch name with backslach symbol replaced with "__"
+     * @return with URL-unsafe characters escaped and slash symbols replaced with underscores
+     */
+    protected String encodeBranchName(String branchName) {
+        return Util.rawEncode(branchName).replace("%2F","___");
     }
 
     /**


jglick:
I think this is dangerous. Have other ideas.

brianfroelund:
Came here because i have a project that will not compile exactly because of the encoded slashes in pathname that are quite common when using git-flow or similar.

dbelyaev:
@jglick Yes, it can be dangerous on some installation or platforms. Do we have a proper and more generic solution?

mjdetullio:
I've pondered this many times over.  The only stable and reasonable solution is encoding branch names to base64, but this does not give pretty or identifiable project names (display name is still set however).  

It would yield a URL such as `http://localhost:8080/jenkins/job/multibranch-project/job/bWFzdGVy` instead of `http://localhost:8080/jenkins/job/multibranch-project/job/master`.

I also have not tested whether views will sort names by display name or the new encoded base64 name.  If display names are not used, sorting by branch name in the UI will not work properly.  Autocomplete for configuration forms needs to be tested as well.

jglick:
:-1: as described in JIRA, like #46

AlberTajuelo:
Are there any way to resolve this issue? Thanks!

dbelyaev:
closing this PR as it seems to not satisfy all the developers (can break existing installation).

PR #46, opened 2016-07-07
Jenkins 34564 - Provide alternate branch encodings

deece:
The default branch encoding method adds '%' to the directory names.

This is problematic as this character has special meanings in some situations (eg. % is a wildcard in makefiles).

This pull request adds alternate encoding methods which do not add special characters:
- Base64 encoding
- Strip encoding, URL encoded, forward slashes replaced with underscore, % replaced with double underscore, underscore replaced with triple underscore

### DIFF
--- a/src/main/java/jenkins/branch/Branch.java
+++ b/src/main/java/jenkins/branch/Branch.java
@@ -23,22 +23,42 @@
  */
 package jenkins.branch;
 
+import hudson.Extension;
+import hudson.ExtensionPoint;
 import hudson.Util;
+import hudson.model.AbstractDescribableImpl;
+import hudson.model.Descriptor;
+import hudson.model.Descriptor.FormException;
 import hudson.scm.NullSCM;
 import hudson.scm.SCM;
+import hudson.scm.SCMDescriptor;
+
+import javax.xml.bind.DatatypeConverter;
+
+import java.io.Serializable;
+import java.io.UnsupportedEncodingException;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
+
+import jenkins.model.GlobalConfiguration;
+import jenkins.model.Jenkins;
 import jenkins.scm.api.SCMHead;
 import jenkins.scm.impl.NullSCMSource;
+import net.sf.json.JSONObject;
+
+import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.export.Exported;
 import org.kohsuke.stapler.export.ExportedBean;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
 
 /**
  * A source code branch.
  * Not to be subclassed outside this plugin.
  */
 @ExportedBean
-public class Branch {
+public class Branch extends AbstractDescribableImpl<Branch> {
 
     /**
      * The ID of our {@link jenkins.scm.api.SCMSource}
@@ -67,6 +87,7 @@ public class Branch {
      * @param head     the name of the branch.
      * @param scm      the {@link SCM} for the branch.
      */
+    @DataBoundConstructor
     public Branch(String sourceId, SCMHead head, SCM scm, List<? extends BranchProperty> properties) {
         this.sourceId = sourceId;
         this.head = head;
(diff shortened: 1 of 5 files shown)


michaelneale:
Interesting. 

This won't fix it for users until they find that setting and then all new projects use one of the alt encodings though.

deece:
Correct, the intent is to do things in a backwards-compatible way. Since this plugin does not deal with the directory itself (maybe it should?), it can't rename existing directories when they have to change.

Simply changing the default encoding will result in users losing their build history, as the consumers of the API will delete it as an orphaned branch.

mjdetullio:
Could a `hudson.model.listeners.SaveableListener` be used to migrate the branch projects when the setting is changed?  What about if a concurrent indexing happens while said migration is running?

deece:
I'm considering migration/renaming as out of scope for this PR, but due to the complexity & care required (as Matt points out above) could be doable as a separate PR.

The PR as it stands let's the user know (via help) the implications of changing the encoding method, which is an improvement over the workflow that is currently required, which involves manually patching and rebuilding the Branch API plugin, and (possibly unexpectedly) losing the history.

Is there a way we can prompt the user directly for confirmation of the change? I'm not that happy with having the warning is the help section as there is no guarantee the user will see it.

superboum:
I'm also interested by this PR.

I've internally modified the branch-api-plugin for my own use here: https://github.com/superboum/branch-api-plugin/tree/features/md5-encoded-name

I'm using the first 8 characters of the md5 hash of the branch. Even if we can have collision, I personally think chances are sufficiently low.

Reasons are:
 * msbuild doesn't like % in the path name and just crash
 * windows `PATH TOO LONG EXCEPTION`, as we have some branch named `bugs/9874-fix-this-very-and-long-bug-that-we-discovered-yesterday`.

Would be happy to have some advice/contribute to merge this fix in upstream without breaking compatibility or everything else.

Thanks :D

mjdetullio:
Actually, MD5 collisions are highly probable and taking a substring of the hash makes it worse to a high order of magnitude.  I'd avoid that at all costs.  Base64 encoding is safe and solves your problem.

michaelneale:
I bed padding of "=" in base64 will upset other tools...

michaelneale:
@deece agree that the migration probably is best done as another PR

mjdetullio:
Padding would be stripped out.  Many decoders do not even require it, and if it does it can easily be re-added.

```
while (encoded.length() % 4 != 0) {
    encoded += "=";
}
String decoded = base64Decode(encoded);
```

superboum:
@mjdetullio Unfortunately it doesn't solve all my problem, as I can't change the workspace used by the multibranch plugin when it clones the repository to fetch the Jenkinsfile.

During the git checkout, git crashes because of the Path too long exception. I've mapped my JENKINS_HOME to a network drive to limit the length of the path, but if the branch is very long my problem is not solved. Indeed, the base64 encoding is proportional to the length of my branch name.

Here is an example:
`bugs/9874-fix-this-very-and-long-bug-that-we-discovered-yesterday`
will become in base64:
`YnVncy85ODc0LWZpeC10aGlzLXZlcnktYW5kLWxvbmctYnVnLXRoYXQtd2UtZGlzY292ZXJlZC15ZXN0ZXJkYXk=`

I understand that my solution is not ideal and very debatable, but many industry projects are made under Windows and use very long path. As an example, the [Unreal Engine](https://answers.unrealengine.com/questions/300284/pathtoolongexception-when-running-setupbat-for-the.html). So it's really important to save any possible characters.

Still the MD5 can be replaced by a SHA1 or another hashing algorithm. To keep more data, it's possible to encode the data of this hash as base64 instead of hexadecimal. Finally we can keep more characters, maybe 12 as it's a fixed length. We should also estimate the risks of collision, I think they are not so high. Especially if we let the user choose its branch encoding strategy as a fix if the other one doesn't work for him.

Finally, I'm not a Windows person, it's just my working OS currently and don't really have the choice to use anything else.

jglick:
:-1: as described in JIRA

dbelyaev:
So, no resolution? And all the developers  who are using "Git Workflow" and msbuild compiler cannot use it?

jglick:
Addressed in #50.

> many industry projects are made under Windows and use very long path

Windows only really cares about the length of single path _components_, so long as you use modern tools. In the case of Git, IIRC there is a patch to make it turn on the magic flag to work properly.

PR #48, opened 2016-08-17
[JENKINS-37219] Add a job property for overriding branch index triggers

abayer:
[JENKINS-37219](https://issues.jenkins-ci.org/browse/JENKINS-37219)

Works in both directions - disables branch index triggers if set to
false, enables them if set to true and either there's a
NoTriggerBranchProperty applying to the branch or there's a
NoTriggerOrganizationFolderProperty applying to the organization.

Theoretically, the config shouldn't show up for any job type but
multibranch jobs, where it'll only show up in the Snippet Generator anyway.

https://github.com/jenkinsci/workflow-multibranch-plugin/pull/26 will be replaced with an integration test of this in the `properties` step.

cc @reviewbybees esp @jglick

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
   </scm>
 
   <properties>
-    <jenkins.version>1.625.1</jenkins.version>
+    <jenkins.version>1.642.3</jenkins.version>
   </properties>
 
   <repositories>
@@ -92,6 +92,11 @@
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>cloudbees-folder</artifactId>
       <version>5.10</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>structs</artifactId>
+      <version>1.4-SNAPSHOT</version>
     </dependency>
     <!-- jenkins dependencies -->
     <!-- test dependencies -->
--- a/pom.xml
+++ b/pom.xml
@@ -96,7 +96,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>structs</artifactId>
-      <version>1.4-SNAPSHOT</version>
+      <version>1.4</version>
     </dependency>
     <!-- jenkins dependencies -->
     <!-- test dependencies -->
--- a/src/main/java/jenkins/branch/NoTriggerBranchProperty.java
+++ b/src/main/java/jenkins/branch/NoTriggerBranchProperty.java
@@ -71,19 +71,19 @@ public class NoTriggerBranchProperty extends BranchProperty {
                     for (Cause c : ((CauseAction) action).getCauses()) {
                         if (c instanceof BranchIndexingCause) {
                             if (p instanceof Job) {
-                                Job j = (Job) p;
+                                Job<?,?> j = (Job) p;
                                 if (j.getParent() instanceof MultiBranchProject) {
 
-                                    OverrideIndexTriggersJobProperty overrideProp =
-                                            (OverrideIndexTriggersJobProperty) j.getProperty(OverrideIndexTriggersJobProperty.class);
-
-                                    for (BranchProperty prop : ((MultiBranchProject) j.getParent()).getProjectFactory().getBranch(j).getProperties()) {
-                                        if (prop instanceof NoTriggerBranchProperty
-                                                && (overrideProp == null || !overrideProp.getEnableTriggers())) {
-                                            return false;
+                                    OverrideIndexTriggersJobProperty overrideProp = j.getProperty(OverrideIndexTriggersJobProperty.class);
+                                    if (overrideProp != null) {
+                                        return overrideProp.getEnableTriggers();
+                                    } else {
+                                        for (BranchProperty prop : ((MultiBranchProject) j.getParent()).getProjectFactory().getBranch(j).getProperties()) {
+                                            if (prop instanceof NoTriggerBranchProperty) {
+                                                return false;
+                                            }
                                         }
                                     }
-
                                 }
                             }
                         }
--- a/src/main/java/jenkins/branch/NoTriggerOrganizationFolderProperty.java
+++ b/src/main/java/jenkins/branch/NoTriggerOrganizationFolderProperty.java
@@ -90,6 +90,9 @@ public class NoTriggerOrganizationFolderProperty extends AbstractFolderProperty<
                         if (c instanceof BranchIndexingCause) {
                             if (p instanceof Job) {
                                 Job j = (Job) p;
+                                OverrideIndexTriggersJobProperty overrideProp =
+                                        (OverrideIndexTriggersJobProperty) j.getProperty(OverrideIndexTriggersJobProperty.class);
+
                                 if (j.getParent() instanceof MultiBranchProject) {
                                     MultiBranchProject mbp = (MultiBranchProject) j.getParent();
                                     if (mbp.getParent() instanceof OrganizationFolder) {
@@ -97,7 +100,8 @@ public class NoTriggerOrganizationFolderProperty extends AbstractFolderProperty<
                                         if (prop != null) {
                                             // Not necessarily the same as j.getName(), which may be encoded:
                                             String name = mbp.getProjectFactory().getBranch(j).getName();
-                                            if (!name.matches(prop.getBranches())) {
+                                            if (!name.matches(prop.getBranches()) &&
+                                                    (overrideProp == null || !overrideProp.getEnableTriggers())) {
                                                 return false;
                                             }
                                         }
--- a/src/main/java/jenkins/branch/NoTriggerOrganizationFolderProperty.java
+++ b/src/main/java/jenkins/branch/NoTriggerOrganizationFolderProperty.java
@@ -89,20 +89,22 @@ public class NoTriggerOrganizationFolderProperty extends AbstractFolderProperty<
                     for (Cause c : ((CauseAction) action).getCauses()) {
                         if (c instanceof BranchIndexingCause) {
                             if (p instanceof Job) {
-                                Job j = (Job) p;
-                                OverrideIndexTriggersJobProperty overrideProp =
-                                        (OverrideIndexTriggersJobProperty) j.getProperty(OverrideIndexTriggersJobProperty.class);
+                                Job<?,?> j = (Job) p;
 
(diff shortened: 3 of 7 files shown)


abayer:
So the config properly doesn't show up normally! The problem is it only shows up with the help and nothing else in multibranch snippet generator. D'oh. Working on it.

abayer:
@jglick So...how would I distinguish between when the snippet generator is for a multibranch job rather than for a non-multibranch job? 'cos otherwise, it just doesn't show up in the snippet generator ever.

abayer:
ping @jglick

src/main/java/jenkins/branch/OverrideIndexTriggersJobProperty.java
```diff
@@ -0,0 +1,104 @@
...
+
+    public boolean getEnableTriggers() {
+        return enableTriggers;
+    }
+
+    @Extension
+    @Symbol("overrideIndexTriggers")
+    public static class DescriptorImpl extends JobPropertyDescriptor {
+
+        public boolean isOwnerMultibranch(Job<?,?> owner) {
```
jglick: `it` could be any `Item` (cf. `Snippetizer/index.jelly` and `getItem(StaplerRequest)`). To account for _Pipeline Syntax_ (i.e., `Snippetizer`) being displayed from various contexts—`OrganizationFolder`, `MultiBranchProject`, `WorkflowJob`—you could perhaps check

```java
return item instanceof MultiBranchProject || item instanceof OrganizationFolder || item.getParent() instanceof MultiBranchProject;
```

Try something along those lines.

pom.xml
```diff
@@ -93,6 +93,11 @@
       <artifactId>cloudbees-folder</artifactId>
       <version>5.10</version>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>structs</artifactId>
+      <version>1.4-SNAPSHOT</version>
```
jglick: :bug: update to 1.4

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
```diff
@@ -73,11 +73,17 @@ public boolean shouldSchedule(Queue.Task p, List<Action> actions) {
                             if (p instanceof Job) {
                                 Job j = (Job) p;
                                 if (j.getParent() instanceof MultiBranchProject) {
+
+                                    OverrideIndexTriggersJobProperty overrideProp =
+                                            (OverrideIndexTriggersJobProperty) j.getProperty(OverrideIndexTriggersJobProperty.class);
```
jglick: BTW if you declare `j` to be a `Job<?,?>` you do not need the cast.

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
```diff
@@ -73,11 +73,17 @@ public boolean shouldSchedule(Queue.Task p, List<Action> actions) {
...
+
                                     for (BranchProperty prop : ((MultiBranchProject) j.getParent()).getProjectFactory().getBranch(j).getProperties()) {
-                                        if (prop instanceof NoTriggerBranchProperty) {
+                                        if (prop instanceof NoTriggerBranchProperty
+                                                && (overrideProp == null || !overrideProp.getEnableTriggers())) {
```
jglick: :ant: easier, clearer, and (ever so slightly) faster to write

```java
OverrideIndexTriggersJobProperty overrideProp = j.getProperty(OverrideIndexTriggersJobProperty.class);
if (overrideProp != null) {
    return overrideProp.getEnableTriggers();
} else {
    for (BranchProperty prop : ((MultiBranchProject) j.getParent()).getProjectFactory().getBranch(j).getProperties()) {
        if (prop instanceof NoTriggerBranchProperty) {
            return false;
        }
    }
}
```

src/main/java/jenkins/branch/NoTriggerOrganizationFolderProperty.java
```diff
@@ -90,14 +90,18 @@ public boolean shouldSchedule(Queue.Task p, List<Action> actions) {
                         if (c instanceof BranchIndexingCause) {
                             if (p instanceof Job) {
                                 Job j = (Job) p;
+                                OverrideIndexTriggersJobProperty overrideProp =
+                                        (OverrideIndexTriggersJobProperty) j.getProperty(OverrideIndexTriggersJobProperty.class);
```
jglick: :ant: ditto

src/main/java/jenkins/branch/NoTriggerOrganizationFolderProperty.java
```diff
@@ -90,14 +90,18 @@ public boolean shouldSchedule(Queue.Task p, List<Action> actions) {
...
                                 if (j.getParent() instanceof MultiBranchProject) {
                                     MultiBranchProject mbp = (MultiBranchProject) j.getParent();
                                     if (mbp.getParent() instanceof OrganizationFolder) {
                                         NoTriggerOrganizationFolderProperty prop = ((OrganizationFolder) mbp.getParent()).getProperties().get(NoTriggerOrganizationFolderProperty.class);
                                         if (prop != null) {
                                             // Not necessarily the same as j.getName(), which may be encoded:
                                             String name = mbp.getProjectFactory().getBranch(j).getName();
-                                            if (!name.matches(prop.getBranches())) {
+                                            if (!name.matches(prop.getBranches()) &&
+                                                    (overrideProp == null || !overrideProp.getEnableTriggers())) {
```
jglick: :ant: ditto

src/main/java/jenkins/branch/OverrideIndexTriggersJobProperty.java
```diff
@@ -0,0 +1,104 @@
...
+import org.kohsuke.stapler.StaplerRequest;
+
+import java.util.List;
+
+/**
+ * Allows overriding indexing triggers for an individual job - either by enabling when the multibranch or org is set to
+ * suppress them, or disabling if they're otherwise enabled.
+ */
+public class OverrideIndexTriggersJobProperty extends JobProperty<Job<?,?>> {
+    private boolean enableTriggers;
```
jglick: :ant: `final`

src/main/java/jenkins/branch/OverrideIndexTriggersJobProperty.java
```diff
@@ -0,0 +1,104 @@
...
+        @Override
+        public boolean shouldSchedule(Queue.Task p, List<Action> actions) {
+            for (Action action : actions) {
+                if (action instanceof CauseAction) {
+                    for (Cause c : ((CauseAction) action).getCauses()) {
+                        if (c instanceof BranchIndexingCause) {
+                            if (p instanceof Job) {
+                                Job j = (Job) p;
+                                OverrideIndexTriggersJobProperty overrideProp =
+                                        (OverrideIndexTriggersJobProperty) j.getProperty(OverrideIndexTriggersJobProperty.class);
```
jglick: :ant: ditto

src/main/resources/jenkins/branch/OverrideIndexTriggersJobProperty/config.jelly
```diff
@@ -0,0 +1,38 @@
...
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+-->
+
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:f="/lib/form">
+    <j:if test="${descriptor.isOwnerMultibranch(it)}">
+        <f:block>
```
jglick: :bug: I think you need to delete the `f:block` wrapper. Configuration forms are generally in the context of a rowset; `f:optionalBlock` expects to be in a rowset and invokes its body in a rowset; `f:block` expects to be in a rowset and invokes its body as inline content.

Even if a particular browser is forgiving, HtmlUnit etc. may not be.

src/main/resources/jenkins/branch/OverrideIndexTriggersJobProperty/help.html
```diff
@@ -0,0 +1,11 @@
+<div>
```
jglick: You might want to make this `help-enableTriggers.html`. Not sure which looks better in practice.

src/test/java/jenkins/branch/OverrideIndexTriggersJobPropertyTest.java
```diff
@@ -0,0 +1,226 @@
...
+import static org.junit.Assert.assertNotNull;
+
+public class OverrideIndexTriggersJobPropertyTest {
+
+    @Rule
+    public JenkinsRule r = new JenkinsRule();
+    @Rule
+    public GitSampleRepoRule sampleRepo = new GitSampleRepoRule();
+    @Rule
+    public GitSampleRepoRule orgSampleRepo = new GitSampleRepoRule();
```
jglick: :ant: Why the two rules? You do not appear to be using both from a single test case.

src/main/resources/jenkins/branch/OverrideIndexTriggersJobProperty/help.html
abayer: Something went weird when I did it as `help-enableTriggers.html`, but can't remember what off the top of my head.

src/test/java/jenkins/branch/OverrideIndexTriggersJobPropertyTest.java
abayer: ...because I forgot they reset between test cases.

src/main/resources/jenkins/branch/OverrideIndexTriggersJobProperty/config.jelly
abayer: Ok.

jglick:
:bee:

e0003315:
Hi, i am new to jenkinspipeline. Do you have an example of how i can use this to prevent automatic scm trigger in jenkinsfile? it looks complicated. I was hoping for something like properties[autoSCMTrigger: false]

PR #50, opened 2016-09-17
[JENKINS-34564] Safer workspace names for branch projects

jglick:
[JENKINS-34564](https://issues.jenkins-ci.org/browse/JENKINS-34564)

As opposed to #46, this only touches the workspace paths, not the actual `Item.fullName`; guarantees unique names; and bounds path lengths too.

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
+++ b/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
@@ -24,7 +24,7 @@
 
 package jenkins.branch;
 
-import com.google.common.hash.Hashing;
+import com.google.common.base.Charsets;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.model.Item;
@@ -32,12 +32,15 @@ import hudson.model.Node;
 import hudson.model.Slave;
 import hudson.model.TopLevelItem;
 import hudson.model.listeners.ItemListener;
-import hudson.remoting.Base64;
 import java.io.IOException;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import jenkins.model.Jenkins;
 import jenkins.slaves.WorkspaceLocator;
+import org.apache.commons.codec.binary.Base32;
+import org.apache.commons.lang.StringUtils;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 
@@ -52,6 +55,7 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
     private static final Logger LOGGER = Logger.getLogger(WorkspaceLocatorImpl.class.getName());
 
     /** The most characters to allow in a workspace directory name, relative to the root. Zero to disable altogether. */
+    // TODO 2.4+ use SystemProperties
     @SuppressWarnings("FieldMayBeFinal")
     private static /* not final */ int PATH_MAX = Integer.getInteger(WorkspaceLocatorImpl.class.getName() + ".PATH_MAX", 80);
 
@@ -75,17 +79,20 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
     }
 
     static String uniqueSuffix(String name) {
-        return Base64.encode(Hashing.sha256().hashString(name).asBytes()).
-            replace('/', '_').
-            replace('+', '.').
-            replaceFirst("=+$", "");
+        // TODO still in beta: byte[] sha256 = Hashing.sha256().hashString(name).asBytes();
+        byte[] sha256;
+        try {
+            sha256 = MessageDigest.getInstance("SHA-256").digest(name.getBytes(Charsets.UTF_16LE));
+        } catch (NoSuchAlgorithmException x) {
+            throw new AssertionError("https://docs.oracle.com/javase/7/docs/technotes/guides/security/StandardNames.html#MessageDigest", x);
+        }
+        return new Base32(0).encodeToString(sha256).replaceFirst("=+$", "");
     }
 
     static String minimize(String name) {
-        int maxMnemonic = Math.max(PATH_MAX - 44, 1);
-        String result = name.replaceAll("(%[0-9A-F]{2}|[^a-zA-Z0-9-_.])+", "_").
-            replaceFirst(".*?(.{0," + maxMnemonic + "}$)", "$1") +
-            "-" + uniqueSuffix(name);
+        String mnemonic = name.replaceAll("(%[0-9A-F]{2}|[^a-zA-Z0-9-_.])+", "_");
+        int maxMnemonic = Math.max(PATH_MAX - /* ceil(256 / lg(32)) + length("-") */53, 1);
+        String result = StringUtils.right(mnemonic, maxMnemonic) + "-" + uniqueSuffix(name);
         assert result.length() <= PATH_MAX : result + " does not fit inside " + PATH_MAX;
         return result;
     }
--- a/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
+++ b/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
@@ -49,7 +49,8 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
         if (node instanceof Jenkins) {
             return ((Jenkins) node).getRootPath().child("workspace/" + minimized);
         } else if (node instanceof Slave) {
-            return ((Slave) node).getWorkspaceRoot().child(minimized);
+            FilePath root = ((Slave) node).getWorkspaceRoot();
+            return root != null ? root.child(minimized) : null;
         } else { // ?
             return null;
         }
--- a/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
+++ b/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
@@ -1,0 +1,63 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2016 CloudBees, Inc.
+ *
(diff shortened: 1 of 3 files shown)


abayer:
I'm wary of the length of the workspace names. Admittedly, I'm wary of pretty much anything that makes workspace paths longer, but...that's really long, and there are things out there (i.e., some older versions of python) that just choke out with >256 character paths, so...yeah. Is there any way we can do shorter than this?

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -0,0 +1,132 @@
...
+ */
+@Restricted(NoExternalUse.class)
+@Extension(ordinal = -100)
+public class WorkspaceLocatorImpl extends WorkspaceLocator {
+
+    private static final Logger LOGGER = Logger.getLogger(WorkspaceLocatorImpl.class.getName());
+
+    /** The most characters to allow in a workspace directory name, relative to the root. Zero to disable altogether. */
+    @SuppressWarnings("FieldMayBeFinal")
+    private static /* not final */ int PATH_MAX = Integer.getInteger(WorkspaceLocatorImpl.class.getName() + ".PATH_MAX", 80);
```
stephenc: TODO on switching to `SystemProperties`

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -0,0 +1,132 @@
...
+    public FilePath locate(TopLevelItem item, Node node) {
+        if (PATH_MAX == 0) {
+            return null;
+        }
+        if (!(item.getParent() instanceof MultiBranchProject)) {
+            return null;
+        }
+        String minimized = minimize(item.getFullName());
+        if (node instanceof Jenkins) {
+            return ((Jenkins) node).getRootPath().child("workspace/" + minimized);
```
stephenc: is there no accessor for this?

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -0,0 +1,132 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package jenkins.branch;
+
+import com.google.common.hash.Hashing;
```
stephenc: Is this a stable API? I've been badly burned by google commons before... I have little trust and we probably could just use the helper methods in `Util`

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -0,0 +1,132 @@
...
+        return Base64.encode(Hashing.sha256().hashString(name).asBytes()).
+            replace('/', '_').
+            replace('+', '.').
+            replaceFirst("=+$", "");
+    }
+
+    static String minimize(String name) {
+        int maxMnemonic = Math.max(PATH_MAX - 44, 1);
+        String result = name.replaceAll("(%[0-9A-F]{2}|[^a-zA-Z0-9-_.])+", "_").
+            replaceFirst(".*?(.{0," + maxMnemonic + "}$)", "$1") +
```
stephenc: Seems like a lot of regex magic for what could be a simple `StringUtils.left(safeName,maxMnemonic) + "-" + uniqueSuffix(name)`

Obfuscates exactly what is being done

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -0,0 +1,132 @@
...
+        } else if (node instanceof Slave) {
+            FilePath root = ((Slave) node).getWorkspaceRoot();
+            return root != null ? root.child(minimized) : null;
+        } else { // ?
+            return null;
+        }
+    }
+
+    static String uniqueSuffix(String name) {
+        return Base64.encode(Hashing.sha256().hashString(name).asBytes()).
```
stephenc: We probably want Base36 encoding so that on case insensitive filesystems we have less risk of collision (i.e. macOS and Windows)

src/test/java/jenkins/branch/harness/BranchProjectFactoryImpl.java
```diff
@@ -42,7 +42,7 @@ public BranchProjectFactoryImpl() {}
 
     @Override
     public FreeStyleProject newInstance(Branch branch) {
-        FreeStyleProject job = new FreeStyleProject(getOwner(), branch.getName());
+        FreeStyleProject job = new FreeStyleProject(getOwner(), branch.getEncodedName());
```
stephenc: perhaps we should be setting the display name to `branch.getName()` if we are switching to encoded name

jglick:
> I'm wary of the length of the workspace names.

Well that is exactly the point here: some people were running into problems because the _existing_ workspace names were too  long. This caps them. See JIRA.

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
jglick: Not that I could find, no. See JIRA—the standard workspace determination for master builds is complicated (as well as buggy).

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
jglick: Good point. Very unlikely, but more likely than an actual cryptographic collision.

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
jglick: Can check if that works, thanks.

src/test/java/jenkins/branch/harness/BranchProjectFactoryImpl.java
jglick: We [already are](https://github.com/jenkinsci/branch-api-plugin/blob/78766a821ebd05eb62c8eadc7d5d7bd135fa1bcd/src/main/java/jenkins/branch/MultiBranchProject.java#L348-L354). This patch just affects the workspace determination.

abayer:
I'm still wary, but the argument seems reasonable, so :bee:

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -0,0 +1,139 @@
...
+            throw new AssertionError("https://docs.oracle.com/javase/7/docs/technotes/guides/security/StandardNames.html#MessageDigest", x);
+        }
+        return new Base32(0).encodeToString(sha256).replaceFirst("=+$", "");
+    }
+
+    static String minimize(String name) {
+        String mnemonic = name.replaceAll("(%[0-9A-F]{2}|[^a-zA-Z0-9-_.])+", "_");
+        int maxMnemonic = Math.max(PATH_MAX - /* ceil(256 / lg(32)) + length("-") */53, 1);
+        String result = StringUtils.right(mnemonic, maxMnemonic) + "-" + uniqueSuffix(name);
+        assert result.length() <= PATH_MAX : result + " does not fit inside " + PATH_MAX;
```
jglick: Note that this will trigger if you set `0 < PATH_MAX < 54` for some reason, though assertions would be off by default anyway.

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * Cleans up workspace when an orphaned branch project is deleted.
+     * @see "JENKINS-2111"
+     */
+    @Extension
+    public static class Deleter extends ItemListener {
+
+        @Override
+        public void onDeleted(Item item) {
+            if (item.getParent() instanceof MultiBranchProject) {
```
jglick: Could perhaps skip this block if `PATH_MAX == 0`, to save a bit of time; though it is still potentially useful in case the feature was active in a previous session and workspaces were created.

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
stephenc: as we use this logic in more than one place, could we at least refactor it into a static helper method so that it is easier to fix later

jglick: In `onDeleted` you mean. Yes that would make sense, if there is ever any reason to touch that code.

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -0,0 +1,139 @@
...
+                String suffix = uniqueSuffix(item.getFullName());
+                Jenkins jenkins = Jenkins.getActiveInstance();
+                cleanUp(suffix, jenkins.getRootPath().child("workspace"), jenkins);
```
orrc: I think this won't work as expected when the "Workspace Root Directory" setting (`Jenkins#workspaceDir`) has been customised in the global config.

joeydong:
Looks like there is a character limit somewhere else in the system which is having a problem with these longer names. See the discussion here https://issues.jenkins-ci.org/browse/JENKINS-38706

Techcadia:
This pull request is breaking my windows builds as mentioned in the ticket above. I only had 30ish characters to spare from my 260 limit and this pull took them all

PR #54, opened 2016-10-28
Clarify that CHANGE_* properties may or may not be set

jglick:
@reviewbybees esp. @daniel-beck who requested this


abayer:
:bee:

PR #56, opened 2016-11-04
[JENKINS-39520] ExtensionList.removeAll is unimplemented

jglick:
[JENKINS-39520](https://issues.jenkins-ci.org/browse/JENKINS-39520)

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/branch/CustomOrganizationFolderDescriptor.java
+++ b/src/main/java/jenkins/branch/CustomOrganizationFolderDescriptor.java
@@ -150,7 +150,9 @@ public class CustomOrganizationFolderDescriptor extends TopLevelItemDescriptor i
             }
         }
         LOGGER.log(Level.FINE, "clearing {0}", old);
-        all().removeAll(old);
+        for (CustomOrganizationFolderDescriptor d : old) {
+            all().remove(d);
+        }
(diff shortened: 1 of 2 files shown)


recena:
🐝

src/main/java/jenkins/branch/CustomOrganizationFolderDescriptor.java
```diff
@@ -150,7 +150,9 @@ private static void doAddSpecificDescriptors() {
             }
         }
         LOGGER.log(Level.FINE, "clearing {0}", old);
-        all().removeAll(old);
+        for (CustomOrganizationFolderDescriptor d : old) {
+            all().remove(d);
```
stephenc: :bug: can you show that this will not cause an infinite loop of death as you are now calling https://github.com/jenkinsci/jenkins/blob/6718a71f4be8cda7c575ae40ea0befec5c2ff1cf/core/src/main/java/hudson/ExtensionList.java#L197-L205

jglick: Yes but on a different extension list.

stephenc: ok (bee)

PR #57, opened 2016-11-17
[JENKINS-39816] Fix PCT against >= 2.16

varyvol:
Add plugin dependency as first level dependency.

@reviewbybees @stephenc

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -135,6 +135,12 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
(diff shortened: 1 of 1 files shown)


jglick:
Why should the plugin express a dependency on (an ancient version of) a plugin which it does not use in tests? What concrete problem is this purporting to solve? JIRA contains no explanation.

stephenc:
No clue. I wonder can we delete this dependency. (As we cannot depend on latest - circular dependency) and PCT will just have to solve the actual problem rather than adding this hack l

PR #59, opened 2016-12-09
More tests

stephenc:
I suspect we should have some test failures as these are new test cases based on some manual testing I have been doing

@reviewbybees


rsandell:
:bee: although semi blind on a galloping horse.

PR #61, opened 2016-12-16
Make compilable using JDK 9 javac

jglick:
@reviewbybees


src/main/java/jenkins/branch/LocalizedViewsTabBar.java
```diff
@@ -52,7 +52,7 @@
     @Restricted(NoExternalUse.class)
     @SuppressWarnings("unused") // invoked from stapler view
     public List<View> sort(@Nonnull List<? extends View> views) {
-        List<View> result = new ArrayList<>(views);
+        List<View> result = new ArrayList<View>(views);
```
stephenc: seriously, java 9 cannot use diamond? wtf

jglick: It can—but the inferred type here would have been something based on `? extends View` rather than plain `View`. Subtle fix to generics logic.

PR #62, opened 2017-01-01
[JENKINS-34564] Allow workspace paths to be less than 54 characters.

creste:
jenkinsci/branch-api-plugin#50 added `PATH_MAX`, which defaults to 80.  I assumed that I could set `PATH_MAX` to 20 to enforce a maximum workspace path of 20 characters.  I need somewhere between 15 and 20 character workspace paths due to Windows path limitations documented in [JENKINS-34564](https://issues.jenkins-ci.org/browse/JENKINS-34564).  However, due to [the way the minimize() function uses `PATH_MAX`](https://github.com/jenkinsci/branch-api-plugin/blob/4fdf888d62848cbf6d91255607d39dd71b536653/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java#L95), the effective minimum for `PATH_MAX` is actually 54 characters, which is too many for me.

This pull request reduces the effective minimum of `PATH_MAX` to 12 characters: one for the last character of the mnemonic, one for the "-", and 10 for the hash.  The unit tests show examples of workspace path names for various `PATH_MAX` values.  I chose 10 characters for the hash by computing the [probability of a collision with 100 workspaces checked out](https://www.wolframalpha.com/input/?i=1+-+e%5E((-100*(100-1))%2F(2*1125899906842624))). 4*(10^-12) seems reasonably low to me.

The default value of `PATH_MAX` is still 80, so this should not affect users who are still using the default value.  This also shouldn't affect users who have `PATH_MAX` set to 54 or higher.  This should only affect users with `PATH_MAX` between 1 and 53, inclusive.

According to @jglick, the original reason the entire 52-character hash was included in the workspace path was to reduce the probability of ["**malicious** collisions, which are easily constructed with weakened hashes"](https://issues.jenkins-ci.org/browse/JENKINS-34564?focusedCommentId=274302&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-274302).  If that is a valid concern then users can keep `PATH_MAX` at 54 or higher.  If that is not a valid concern for some users, or if some users are forced to use shorter paths due to more important concerns and they understand the security risk, then they can reduce `PATH_MAX` below 54.

My understanding is that a ["proper" solution to this problem is being implemented for JENKINS-2111](https://issues.jenkins-ci.org/browse/JENKINS-34564?focusedCommentId=274302&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-274302).  From @jglick:

>The real fix is to maintain a per-node database of workspaces, as discussed in JENKINS-2111. Then each workspace can be given a guaranteed unique, human-readable name of guaranteed maximum length—as well as solving other longstanding issues not limited to branch projects.

However, I don't know when that work will be finished.  That is why I am creating this pull request that does the minimum number of changes to fix my problem instead of waiting an indeterminate amount of time for JENKINS-2111 to be fixed.

### DIFF
--- a/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
+++ b/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
@@ -56,7 +56,7 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
 
     /** The most characters to allow in a workspace directory name, relative to the root. Zero to disable altogether. */
     // TODO 2.4+ use SystemProperties
-    private static /* not final */ int PATH_MAX = Integer.getInteger(WorkspaceLocatorImpl.class.getName() + ".PATH_MAX", 80);
+    static /* not final */ int PATH_MAX = Integer.getInteger(WorkspaceLocatorImpl.class.getName() + ".PATH_MAX", 80);
 
     @Override
     public FilePath locate(TopLevelItem item, Node node) {
@@ -75,10 +75,6 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
         } else { // ?
             return null;
         }
-    }
-
-    static void setPathMax(int pathMax) {
-        PATH_MAX = pathMax;
     }
 
     static String uniqueSuffix(String name) {
@@ -101,6 +97,7 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
             // and the mnemonic to fit inside PATH_MAX.  The mnemonic always gets
             // at least one character.  The suffix always gets 10 characters plus
             // the "-".  The rest of PATH_MAX is split evenly between the two.
+            LOGGER.log(Level.WARNING, "WorkspaceLocatorImpl.PATH_MAX is small enough that workspace path collisions are more likely to occur");
             final int minSuffix = 10 + /* length("-") */ 1;
             maxMnemonic = Math.max((int)((PATH_MAX - minSuffix) / 2), 1);
             maxSuffix = Math.max(PATH_MAX - maxMnemonic, minSuffix);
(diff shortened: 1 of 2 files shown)


creste:
I'm not sure what the appropriate amount of time is to wait for input on PRs in the repository, so I'm sorry if I'm out of line by asking, but do you (@jglick or @stephenc) have any thoughts on this?

Roemer:
Is this going to be merged anytime soon?

creste:
@Roemer - I haven't heard anything from @jglick or @stephenc about this PR or the comments I added to the Jenkins issues.  It doesn't look like this will be merged any time soon.

Roemer:
Guess they are fairly busy with the SCM API 2.0 regression problems, I still hope this finds its way in because we have a lot of problems with long paths and msbuild and deepely nested projects...

stephenc:
I would much rather migrate to using the `NameMangler` to generate the workspace names as that will also give uniqueness (at least as far as the branch name is concerned) with length capped out at 32 characters (through perhaps we could tweak the algorithm in `NameMangler` to make that shorter if needed)

In any case this is not going to be merged until we can get 2.0.x back into the general update center and are in bug fixing mode... plus I would prefer to see some addressing for the clean-up of any workspaces created with a different PATH_MAX otherwise an admin changing the PATH_MAX could end up out of disk space as they ratchet down.

creste:
@stephenc, thank you for the response.

I glanced at `NameMangler` and I see that its length is capped at 32 characters like you said.  For my purposes, anything you could do to get that length down to 20 would be greatly appreciated.

Can you confirm that if I upgrade to 2.0.x with `NameMangler` then all of my multibranch pipeline builds will have workspace names restricted to `NameMangler::MAX_SAFE_LENGTH`?  Or is there some other change needed to make multibranch piplines use `NameMangler` for workspace paths?  I'd be happy to help if you can provide a little more guidance on the areas of code that need to be tweaked.

Also, do you have a timeline for when 2.0.x will be available in the general update center?  This has been a blocking issue for me for several months.

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -78,6 +77,10 @@ public FilePath locate(TopLevelItem item, Node node) {
         }
     }
 
+    static void setPathMax(int pathMax) {
```
jglick: Just make the field package access.

creste:
Thank you, @jglick!  I pushed changes to address the comments you made and re-ran the unit tests to confirm it still works.

PR #63, opened 2017-01-04
BranchStatusColumn should not be exposed to users

stephenc:
@reviewbybees (found during some other stuff)

### DIFF
--- a/src/main/java/jenkins/branch/BranchStatusColumn.java
+++ b/src/main/java/jenkins/branch/BranchStatusColumn.java
@@ -26,6 +26,8 @@ package jenkins.branch;
 
 import hudson.Extension;
 import hudson.model.BallColor;
+import hudson.model.Descriptor;
+import hudson.model.DescriptorVisibilityFilter;
 import hudson.model.Item;
 import hudson.model.Job;
 import hudson.views.ListViewColumn;
@@ -93,7 +95,7 @@ public class BranchStatusColumn extends StatusColumn {
          */
         @Override
         public String getDisplayName() {
-            return Messages.DescriptionColumn_displayName();
+            return Messages.BranchStatusColumn_displayName();
         }
 
         /**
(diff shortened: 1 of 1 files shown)


src/main/java/jenkins/branch/BranchStatusColumn.java
```diff
@@ -104,4 +106,19 @@ public boolean shownByDefault() {
...
+     */
+    @Extension
+    public static class DescriptorVisibilityFilterImpl extends DescriptorVisibilityFilter {
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public boolean filter(Object context, Descriptor descriptor) {
+            return !(descriptor instanceof DescriptorImpl) || context instanceof MultiBranchProjectViewHolder.ViewImpl;
```
rsandell: won't `context instanceof MultiBranchProjectViewHolder.ViewImpl` hide all descriptors if the view is of that type?

rsandell: Or maybe that is the intention?

stephenc: either it is not `BranchStatusColumn.DescriptorImpl`, in which case we take no action... or it is  `BranchStatusColumn.DescriptorImpl`, in which case we hide it from everywhere except the `MultiBranchProjectViewHolder.ViewImpl` (which is non-user configurable so we could hide it here also, except there may be other things that rewrite and apply the filters, so to be safe we allow it for that non-user configurable view)

rsandell:
:bee:

raul-arabaolaza:
AFAIK :bee:

src/main/java/jenkins/branch/BranchStatusColumn.java
jglick: Might be clearer if written as

```java
return !(descriptor instanceof DescriptorImpl &&
         !(context instanceof MultiBranchProjectViewHolder.ViewImpl));
```

or even

```java
if (descriptor instanceof DescriptorImpl &&
    !(context instanceof MultiBranchProjectViewHolder.ViewImpl)) {
    return false;
} else {
    return true;
}
```

Of course clearer and more efficient still would have been if `@Extension` had never been invented and SezPoz had been used as intended (`-source 8` for brevity):

```java
@DescriptorVisibilityFilterRegistration(type=DescriptorImpl.class) public static final DescriptorVisibilityFilter HIDE_ME_ELSEWHERE =
    (context, descriptor) -> context instanceof MultiBranchProjectViewHolder.ViewImpl;
```

PR #66, opened 2017-01-05
[JENKINS-40829] Implement extension point for JENKINS-40829 support]

stephenc:
See [JENKINS-40829](https://issues.jenkins-ci.org/browse/JENKINS-40829)

Downstream of https://github.com/jenkinsci/scm-api-plugin/pull/23

@reviewbybees


src/main/java/jenkins/branch/BranchProjectFactory.java
```diff
@@ -281,4 +282,27 @@ public SCMHead getHead(Item item) {
 
     }
 
+    @Restricted(DoNotUse.class)
```
rsandell: why?

rsandell: As in do not use directly I guess?

rsandell:
:bee:

PR #68, opened 2017-01-06
[FIXED JENKINS-40865] Org folders should encode child project names

stephenc:
- Also switches to name mangling for multi-branch projects

See [JENKINS-40865](https://issues.jenkins-ci.org/browse/JENKINS-40865)

@reviewbybees esp @jglick


jglick:
Also:

```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertNotNull(Assert.java:712)
	at org.junit.Assert.assertNotNull(Assert.java:722)
	at jenkins.branch.WorkspaceLocatorImplTest.locate(WorkspaceLocatorImplTest.java:78)
```

stephenc:
`Util.rawEncode("..").equals("..")` FTW

stephenc:
> The unit test here does not reproduce an actual problem, just tests behavior of the helper function.

That is because I already added the additional tests in 26505ade48353f6ebffdba3d36af04a569096e55 and 6219ef9ef106001259c18387cc7c213c4addf31d

stephenc:
@jglick ping

jglick:
> I already added the additional tests

And those demonstrated some sort of user-visible error which this change corrects?

jglick:
And did you check that this does not break `workflow-multibranch` downstream tests?

src/main/java/jenkins/branch/Branch.java
```diff
@@ -153,7 +153,7 @@ public String getName() {
      * @since 0.2-beta-7
      */
     public String getEncodedName() {
-        return Util.rawEncode(getName());
+        return NameMangler.apply(getName());
```
jglick: The main concern I have about this is that it means the `Branch.name` is no longer reconstructible from the `Job.name`. So long as nobody is trying to do that, and is using `BranchProjectFactory.getBranch` correctly, then I guess it does not matter. But you must be sure that there can be no (accidental or malicious) collision between `encodedName`s.

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -161,6 +161,21 @@ protected MultiBranchProject(ItemGroup parent, String name) {
     public void onLoad(ItemGroup<? extends Item> parent, String name) throws IOException {
         super.onLoad(parent, name);
         init2();
+        // migrate any non-mangled names
+        ArrayList<P> items = new ArrayList<>(getItems());
+        for (P item : items) {
+            String itemName = item.getName();
+            String mangledName = NameMangler.apply(itemName);
```
jglick: Should this not rather be applying `NameMangler` to the original branch name? Here you are applying it to the encoded name, which seems like it might produce a different result.

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -161,6 +161,21 @@ protected MultiBranchProject(ItemGroup parent, String name) {
...
+            if (!itemName.equals(mangledName)) {
+                if (getItem(mangledName) == null) {
+                    item.renameTo(mangledName);
+                    if (item.getDisplayNameOrNull() == null) {
+                        item.setDisplayName(itemName);
+                        item.save();
+                    }
+                }
```
jglick: And what if there is already an item using the mangled name? Something is going to break, right? Perhaps we should just `item.delete()` in that case to be on the safe side.

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -538,9 +553,17 @@ public P getItem(String name) {
             if (item != null) {
                 return item;
             }
+            item = super.getItem(NameMangler.apply(decoded));
```
jglick: Does not seem to make sense. Should this not be un-mangling, and doing so on `name` rather than `decoded`? Hard to follow what all this logic really is doing; does it have full test coverage?

src/main/java/jenkins/branch/NameMangler.java
```diff
@@ -0,0 +1,172 @@
...
+        }
+        StringBuilder buf = new StringBuilder(name.length() + 16);
+        for (char c : name.toCharArray()) {
+            if (isSafe(c)) {
+                buf.append(c);
+            } else if (c == '/' || c == '\\' || c == ' ') {
+                buf.append('_');
+            } else if (c <= 0xff) {
+                buf.append('%');
+                buf.append(StringUtils.leftPad(Integer.toHexString(c & 0xff), 2, '0'));
```
jglick: IIUC `%` itself is unsafe, so deliberately creating a branch name with a `%` in it will not cause problems—you will just wind up with `%25` in there. Right?

src/main/java/jenkins/branch/NameMangler.java
```diff
@@ -0,0 +1,172 @@
...
+                    dd.append(toDigit(data & 0x1f));
+                    bits -= 5;
+                    data = data >> 5;
+                }
+                data = data | ((b & 0xff) << bits);
+                bits += 8;
+            }
+            digest = dd.toString();
+        } catch (NoSuchAlgorithmException e) {
+            throw new IllegalStateException("MD5 not installed", e);    // impossible
```
jglick: You mean SHA-1?

src/main/java/jenkins/branch/NameMangler.java
```diff
@@ -0,0 +1,172 @@
...
+        }
+        if (buf.length() <= MAX_SAFE_LENGTH - MIN_HASH_LENGTH - 1) {
+            // we have room to add the min hash
```
jglick: If this is false, why bother computing the `digest`?

src/main/java/jenkins/branch/NameMangler.java
```diff
@@ -0,0 +1,172 @@
...
+            buf.append('@');
+            buf.append(StringUtils.right(digest, MIN_HASH_LENGTH));
```
jglick: This is going to give a rather short portion of the hash, so any cryptographic protections are very weak. Meaning someone could easily create a branch name which maliciously mangles to the same text as a legitimate name. I am not sure if this could be used to mount an attack but it seems concerning.

OTOH `WorkspaceLocatorImpl` is already on the hook for guaranteeing that workspaces do not clash, which is important for preventing one job from accessing the workspace of another on a given node, if the build configuration limits access outside the workspace path (as for example `docker.image(…).inside {…}` does, assuming an administrator enforces its use by some means).

(If every build runs on a one-shot containerized executor then you need not care about such things, though you then pay the performance penalty for agent initialization on every build.)

src/main/java/jenkins/branch/NameMangler.java
```diff
@@ -0,0 +1,172 @@
...
+                data = data | ((b & 0xff) << bits);
+                bits += 8;
+            }
+            digest = dd.toString();
+        } catch (NoSuchAlgorithmException e) {
+            throw new IllegalStateException("MD5 not installed", e);    // impossible
+        }
+        if (buf.length() <= MAX_SAFE_LENGTH - MIN_HASH_LENGTH - 1) {
+            // we have room to add the min hash
+            buf.append('@');
```
jglick: Could use

```java
System.getProperty(WorkspaceList.class.getName(), "@")
```

since apparently this character is problematic on some systems. OTOH `WorkspaceLocatorImpl` may already be handling that, according to the test—a little hard to follow.

src/test/java/jenkins/branch/WorkspaceLocatorImplTest.java
```diff
@@ -74,11 +74,11 @@ public void locate() throws Exception {
         stuff.scheduleBuild2(0).getFuture().get();
         r.waitUntilNoActivity();
         showComputation(stuff);
-        FreeStyleProject master = r.jenkins.getItemByFullName("stuff/dev%2Fflow", FreeStyleProject.class);
+        FreeStyleProject master = r.jenkins.getItemByFullName("stuff/"+NameMangler.apply("dev/flow"), FreeStyleProject.class);
```
jglick: Prefer to use a string constant here so a reader of the test can immediately see what the actual item name will be—and we will notice if some change to `NameMangler` potentially might be causing compatibility issues.

src/test/java/jenkins/branch/WorkspaceLocatorImplTest.java
```diff
@@ -74,11 +74,11 @@ public void locate() throws Exception {
...
         assertNotNull(master);
-        assertEquals(r.jenkins.getRootPath().child("workspace/stuff_dev_flow-L5GKER67QGVMJ2UD3JCSGKEV2ACON2O4VO4RNUZ27HGUY32SYVXQ"), r.jenkins.getWorkspaceFor(master));
+        assertEquals(r.jenkins.getRootPath().child("workspace/stuff_"+ NameMangler.apply("dev/flow").replace('@','_')+"-HEAQQA3VKMJWWMJ2LGY5PBAKURNMTX2PMQ3FG7E2R5NWAUUN4SZQ"), r.jenkins.getWorkspaceFor(master));
```
jglick: Ditto here, so we can see what the final workspace path will look like.

src/main/java/jenkins/branch/MultiBranchProject.java
stephenc: If you do getItem("foo%2fbar") as a historical link, then `foo%2fbar` will be decoded into `foo/bar` and then we check if there is an item with the name `NameMangler.apply("foo/bar")`, if there is then that was the item the old link was trying to access

src/main/java/jenkins/branch/NameMangler.java
stephenc: Correct, we use % to identify the hash and we prevent creating names that match a mangled name

src/main/java/jenkins/branch/NameMangler.java
stephenc: ACK

src/main/java/jenkins/branch/NameMangler.java
stephenc: To ensure no collisions, e.g. the hash ensures that `foo bar`, `foo/bar` and `foo\bar` all get unique names

src/main/java/jenkins/branch/NameMangler.java
stephenc: We can move to a different character, but workspace locator is doing its own mangling IIUC

src/main/java/jenkins/branch/NameMangler.java
stephenc: Well it's not quite so easy... we prefix with the start of the name... in fact on this branch we prefix with the entire encoded name and then append a hash of the original name... so you only get to play with the characters that have been mapped to `_`... of which the most you can have is 24 of them... and they only have 3 possible values... so that's 3^24 or ~ 38 bits which is 7 characters... so maybe 32 potential collisions for somebody who's branch is called `////////////////////////` if they have a branch with at least one non-flattened character, e.g. `a_______________________` then there are only 3^23 potential names that have the same prefix and that is a space of ~ 6 characters which is the exact amount of hash that we use.

Now for longer names, you are constrained in that the start and end of the name have to match because we insert the hash in the middle for those cases.

Where the name is between 24 and 32 characters long you only get to play with the characters mapping to `_` in the prefix and postfix and the characters in the middle (overwritten by the hash) but at least for those you get to use a wider set (basically `0-9A-Za-z` otherwise it will be `%` encoded and the length could grow and affect how much of the postfix is output) So there is scope there for collisions in a 6 character hash, but you would need to brute-force it *and* have write permission to the repository

After that we start growing the hash length until it is 12 characters of hash

src/main/java/jenkins/branch/MultiBranchProject.java
stephenc: covered by getItem(mangledName) which will be non-null if the target item exists already

stephenc: So the orphaned item strategy will remove them next time around as they will not have been indexed

stephenc:
@jglick The only change required in workflow-multibranch is

```
diff --git a/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactoryTest.java b/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactoryTest.java
index a4ed9fe..407d045 100644
--- a/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactoryTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactoryTest.java
@@ -69,7 +69,7 @@ public class WorkflowBranchProjectFactoryTest {
                 WorkflowRun b1 = p.getLastBuild();
                 assertEquals(1, b1.getNumber());
                 story.j.assertLogContains("branch=dev/main", b1);
-                story.j.assertLogContains("workspace=dev_main-ZFNHWJSHKH4HUVOQUPOQV6WFX7XUPIKIAQAQ3DV7CCAGIXQW7YSA", b1);
+                story.j.assertLogContains("workspace=dev_main_k31kdj-4YKHANIWHM4KZZRNCZZDALTFKAX6WROU4PEFYWTQXHNMM3PHZQEA", b1);
                 verifyProject(p);
                 sampleRepo.write("Jenkinsfile", script.replace("branch=", "Branch="));
             }
```

other than that the tests all pass as we resolve the old urls via fall-back

src/main/java/jenkins/branch/NameMangler.java
stephenc: Hmmm so `@` would be problematic on z/OS... might have to move to another character (which would require making `_` or `-` unsafe)

recena:
I like the idea behind of this PR. 👍

src/main/java/jenkins/branch/NameMangler.java
```diff
@@ -0,0 +1,172 @@
...
+        if (name.length() <= MAX_SAFE_LENGTH) {
+            boolean unsafe = false;
+            for (char c : name.toCharArray()) {
+                if (!isSafe(c)) {
+                    unsafe = true;
+                    break;
+                }
+            }
+            // See https://msdn.microsoft.com/en-us/library/aa365247 we need to consistently reserve names across all OS
+            if (!unsafe) {
```
oleg-nenashev: IMHO the code should also consult with [ProjectNamingStrategy](https://wiki.jenkins-ci.org/display/JENKINS/Extension+points#Extensionpoints-jenkins.model.ProjectNamingStrategy)

stephenc: <iframe width="560" height="315" src="https://www.youtube.com/embed/dSIKBliboIo" frameborder="0" allowfullscreen></iframe>

I'm sorry Oleg I'm afraid I can't do that

stephenc: If we consult the ProjectNamingStrategy then the decision to encode or not is non-deterministic.

In any case what could we do if the ProjectNamingStrategy objects to our encoded name

recena:
🐝

stephenc:
Hmmm we may have to switch escape characters and reduce the safe set

recena:
🐝

src/main/java/jenkins/branch/NameMangler.java
jglick: Cannot claim to follow that, but my asking the question seems to have forced you to think about implications a bit more, so I will stick with that!

stephenc:
You would have thought that european names would not have big issues... but oh boy... `Éireann` and `España` were perfect at finding some strange bugs!

stephenc:
What `Éireann` and `España` show is that the normaization of names is architecture dependent. So depending on how the branch source returns the names with non-ascii characters and then how the user's filesystem handles those same names... it may be that you end up with some builds being lost.

Not much we can do about that as when we need to do migration we do not know what the `SCMNavigator`'s canonical encoding is and hence cannot tell if the name from the filesystem matches (hence the need for `ProjectNameProperty` going forward)

jglick:
> normaization of names is architecture dependent

Why is that? Are we not just using UTF-8/

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -563,9 +584,36 @@ public P getItem(String name) {
...
+     *
+     * @param branchName the name of the branch.
+     * @return the child job or {@code null} if no such job exists or if the requesting user does ave permission to
+     * view it.
+     * @since 2.0.0
+     */
+    @CheckForNull
+    public P getItemByBranchName(@NonNull String branchName) {
+        BranchProjectFactory<P, R> factory = getProjectFactory();
+        for (P p: getItems()) {
```
jglick: If this is performance-sensitive, could try `getItem(branchName)` first optimistically.

src/main/java/jenkins/branch/ProjectNameProperty.java
```diff
@@ -0,0 +1,65 @@
...
+import hudson.model.Descriptor;
+import net.sf.json.JSONObject;
+import org.kohsuke.stapler.StaplerRequest;
+
+/**
+ * Holds the unmangled project name of a {@link MultiBranchProject} within a {@link OrganizationFolder}.
+ *
+ * @since 2.0.0
+ */
+public class ProjectNameProperty extends AbstractFolderProperty<MultiBranchProject<?,?>> {
```
jglick: I am concerned the class name here could be confusing. After all we might expect “project name” to be `Job.name`, whereas this is intentionally different.

This _is_ the terminology used by `SCMSourceObserver`, so perhaps just clarify in Javadoc that this may be distinct from the `Job.name`.

stephenc:
> Why is that? Are we not just using UTF-8

One would think.... but some filesystems normalize to NFC and others to NFD... I'm sure the test may fail on a FS that normalizes to NFKC or NFKD

jglick:
> some filesystems normalize to NFC and others to NFD

Where are we using filesystem normalization here? We are just taking whatever text `SCMSource` or `SCMNavigator` gives us. Now if the particular SCM happens to implement scanning by inspecting local files, well that is just a problem in the SCM, right?

jglick:
(Mercurial does do this BTW. I long argued against it, but the developers insisted they wanted the tool to be completely neutral about the meaning of things like file paths and branch names, so they are just handled as octet streams, and the local OS do what it will with that.)

stephenc:
> Where are we using filesystem normalization here?

Migrating old data... because we never stored the `projectName` we have to rely on it being inferred from the name in the file system.

PR #69, opened 2017-01-11
[JENKINS-40876] Fixes to display of object metadata

stephenc:
When [JENKINS-40876](https://issues.jenkins-ci.org/browse/JENKINS-40876) is fixed, it becomes handy to try displaying the ObjectDisplayName and fall back to the Object Description

The following screenshot shows the PR's title being visible as a tool-tip

screen shot 2017-01-11 at 11 22 26

Note: It was really *fun* trying to determine how to handle escaping or not of the value. In the end as this is being directly applied to a HTML attribute, it turns out that Jelly does the required escaping for us. If we were providing this as a HTML value then we might need to escape

### DIFF
--- a/src/main/java/jenkins/branch/ItemColumn.java
+++ b/src/main/java/jenkins/branch/ItemColumn.java
@@ -25,8 +25,6 @@
 package jenkins.branch;
 
 import hudson.Extension;
-import hudson.Util;
-import hudson.markup.MarkupFormatter;
 import hudson.model.Actionable;
 import hudson.model.Descriptor;
 import hudson.model.DescriptorVisibilityFilter;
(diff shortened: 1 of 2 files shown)


stephenc:
@reviewbybees

src/main/java/jenkins/branch/ItemColumn.java
```diff
@@ -65,17 +64,26 @@ public boolean isPrimary(Object job) {
...
-     * Gets the description of a job.
+     * Gets the tool-tip title of a job.
      *
      * @param job the job.
-     * @return the description.
+     * @return the tool-tip title unescaped for use in an attribute.
      */
     @SuppressWarnings("unused") // used via Jelly EL binding
-    public String getDescription(Object job) {
+    public String getTitle(Object job) {
```
oleg-nenashev: It is a part of the 2.0 release, hence the compatibility breakage is acceptable. Right?

stephenc: correct, plus jelly only

oleg-nenashev: Make it restricted then?

stephenc: ItemColumn is Restricted

PR #71, opened 2017-01-17
[JENKINS-41121] This is the best hack fix I can come up with

stephenc:
See the FUD in [JENKINS-41121](https://issues.jenkins-ci.org/browse/JENKINS-41121)

- Will only apply to jobs that are already existing
- Will only apply if the job does not have a `state.xml` (which is introduced in 2.0.0)
- Will keep being applied until a full scan
- Once a full scan has completed, will defang
- Will not affect event triggered builds
- Will not affect non-deterministic revision builds - they trigger polling which should be cheaper

Side effects:

- If there are any changes during the restart of the master, those changes will be ignored *forever*
- We are slowing down job loading by one file existence check

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -35,8 +35,6 @@ import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.BulkChange;
 import hudson.Extension;
-import hudson.PluginManager;
-import hudson.PluginWrapper;
 import hudson.Util;
 import hudson.XmlFile;
 import hudson.console.HyperlinkNote;
@@ -61,7 +59,6 @@ import hudson.security.ACL;
 import hudson.security.Permission;
 import hudson.util.LogTaskListener;
 import hudson.util.PersistedList;
-import hudson.util.VersionNumber;
 import hudson.util.io.ReopenableRotatingFileOutputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
@@ -91,7 +88,6 @@ import jenkins.scm.api.SCMEventListener;
 import jenkins.scm.api.SCMHead;
 import jenkins.scm.api.SCMHeadEvent;
 import jenkins.scm.api.SCMHeadObserver;
-import jenkins.scm.api.SCMNavigator;
 import jenkins.scm.api.SCMRevision;
 import jenkins.scm.api.SCMSource;
 import jenkins.scm.api.SCMSourceCriteria;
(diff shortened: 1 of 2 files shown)


stephenc:
Oh and FYI I am not sure that this is testable by anything other than manual testing

stephenc:
To Highlight the trade-off in this:

If there are *ANY* changes that happen in source control in between shutting down the master for the upgrade and restarting the master and the first full index then those changes will *NEVER* trigger a build. *NEVER FULL STOP*. You will need to find them some how and manually trigger a build. The manually triggered build will helpfully trigger the exact correct revision though.

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -1751,7 +1775,22 @@ public void observe(@NonNull SCMHead head, @NonNull SCMRevision revision) {
...
                 boolean needSave = !branch.equals(origBranch) || !branch.getActions().equals(origBranch.getActions());
                 _factory.decorate(_factory.setBranch(project, branch));
-                if (rebuild) {
+                if (rebuild
+                        && event == null // JENKINS-41121 doesn't apply to events, they are new builds
+                        && legacySourceIds != null // JENKINS-41121 doesn't apply to new projects or after first index
+                        && !(legacySourceIds.contains(source.getId())) // JENKINS-41121 doesn't apply if source retained
+                        && legacySourceIds.contains(origBranch.getSourceId())
+                        && revision.isDeterministic()) {
+                    // JENKINS-41121 suppress automatic rebuild for branches being re-assoicated with the correct
```
jglick: typo

stephenc:
@jglick I have been able to narrow the condition... but that does raise the potential that there could be more of a loading performance impact

stephenc:
Can probably move the plugin version check to organisation folder. Which would mean the loading performance is not as badly affected

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -195,6 +202,25 @@ public void onLoad(ItemGroup<? extends Item> parent, String name) throws IOExcep
         if (state == null) {
             state = new State(this);
         }
+        if (getParent() instanceof OrganizationFolder) {
+            if (new File(getRootDir(), ".jenkins-41121").isFile()) {
```
abayer: So...what's going on here? Can you explain this in comments, please?

abayer: Also, better naming, please. I don't like the idea of having even a temporary file have a bug id as its name.

stephenc:
This PR addresses the issue in JENKINS-41255 but the fix in https://github.com/jenkinsci/branch-api-plugin/pull/74 is *much much better*. Closing in favour of #74

PR #74, opened 2017-01-20
[JENKINS-41255] Propagate SCMSource id changes when the new SCMSource is identical to the old

stephenc:
See [JENKINS-41255](https://issues.jenkins-ci.org/browse/JENKINS-41255)

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -58,12 +58,15 @@ import hudson.security.ACL;
 import hudson.security.Permission;
 import hudson.util.LogTaskListener;
 import hudson.util.PersistedList;
+import hudson.util.XStream2;
 import hudson.util.io.ReopenableRotatingFileOutputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.UnsupportedEncodingException;
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -72,6 +75,7 @@ import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.ListIterator;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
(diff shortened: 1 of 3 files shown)


src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -358,6 +362,146 @@ public synchronized void setProjectFactory(BranchProjectFactory<P, R> projectFac
...
+                if (!removals.contains(removal.getSource().getId())) {
+                    continue;
+                }
+                if (!equalButForId(removal.getSource(), addition.getSource())) {
+                    continue;
+                }
+                changedIds.put(removal.getSource().getId(), addition.getSource().getId());
+                // now take this one out of consideration
+                removals.remove(removal.getSource().getId());
+                break;
```
rsandell: I'm not sure I'm following this one, you are only interested in the first removal not matching?

stephenc: If I have paired an addition with a removal, we don't want to check either half of the pair again

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -358,6 +362,149 @@ public synchronized void setProjectFactory(BranchProjectFactory<P, R> projectFac
...
+        Map<String,String> changedIds = new HashMap<>();
+        Set<String> additions = new HashSet<>(newIds);
+        additions.removeAll(oldIds);
+        Set<String> removals = new HashSet<>(oldIds);
+        removals.removeAll(newIds);
+
+        // we will be manipulating the sources, so make sure we own the backing list impl
+        sources = new ArrayList<>(sources);
+        for (ListIterator<BranchSource> i = sources.listIterator(); i.hasNext(); ) {
+            BranchSource addition = i.next();
```
jglick: Why not a simple

```java
for (BranchSource addition : sources) {
```

?

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -358,6 +362,149 @@ public synchronized void setProjectFactory(BranchProjectFactory<P, R> projectFac
...
+
+    private boolean equalButForId(SCMSource a, SCMSource b) {
+        if (!a.getClass().equals(b.getClass())) {
+            return false;
+        }
+        boolean fallBack = false;
+        Class<?> clazz = a.getClass();
+        REFLECTION:
+        while (clazz != null && SCMSource.class.isAssignableFrom(clazz) && clazz != SCMSource.class) {
+            for (Field field : clazz.getDeclaredFields()) {
```
jglick: Surely easier to use the XStream form for the whole `SCMSource` and just white out the `<id>…</id>`.

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -358,6 +362,149 @@ public synchronized void setProjectFactory(BranchProjectFactory<P, R> projectFac
...
+                int modifiers = field.getModifiers();
+                if (Modifier.isStatic(modifiers)) {
+                    continue;
+                }
+                if (Modifier.isTransient(modifiers)) {
+                    continue;
+                }
+                // TODO Will Java 9 modules lock setAccessible permissions? Likely would break XStream too!
```
jglick: Nothing is going to run on Java 9, forget about it…

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -358,6 +362,149 @@ public synchronized void setProjectFactory(BranchProjectFactory<P, R> projectFac
...
+                if (aField == null || bField == null) {
+                    // we know they are different
+                    return false;
+                }
+                if (aField.equals(bField)) {
+                    // next field
+                    continue;
+                }
+                if (aField instanceof Boolean || aField instanceof Number || aField instanceof String) {
+                    // no need to compare primatives by XML, we know they are different
```
jglick: typo

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -358,6 +362,149 @@ public synchronized void setProjectFactory(BranchProjectFactory<P, R> projectFac
...
+
+            clazz = clazz.getSuperclass();
+        }
+        return !fallBack || SOURCE_ID_OMITTED_XSTREAM.toXML(a).equals(SOURCE_ID_OMITTED_XSTREAM.toXML(b));
+    }
+
+    private final static XStream2 SOURCE_ID_OMITTED_XSTREAM = new XStream2();
+
+    static {
+        SOURCE_ID_OMITTED_XSTREAM.omitField(SCMSource.class, "id");
```
jglick: So if you can do this anyway, why the giant pile of reflection crap?

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -358,6 +362,149 @@ public synchronized void setProjectFactory(BranchProjectFactory<P, R> projectFac
     }
 
     /**
+     * Offers direct access to set the configurable list of branch sources <strong>while</strong> preserving
+     * branch source id associations for sources that are otherwise unmodified
+     * @param sources
+     * @throws IOException
+     */
+    public void setSourcesList(List<BranchSource> sources) throws IOException {
```
stephenc: If somebody wants to build an alternative to org folders but retaining multibranch, they need this method to handle legacy navigator implementations

src/main/java/jenkins/branch/MultiBranchProject.java
stephenc: I had in my mind to replace parts of the list as I iterated... for the life of me I cannot fathom why. Will change back (and remove the ArrayList 3 lines above)

src/main/java/jenkins/branch/MultiBranchProject.java
stephenc: Yep. It's premature optimisation... from before I found `omitField` and was doing crazy things to ensure a bare ID would end up escaped the same way as in the whole object... I guess I'll put the reflection code back in my rainy day library

src/main/java/jenkins/branch/MultiBranchProject.java
stephenc: Reflection code was fun to write... ok no it wasn't but it was less scary that the evil reflex stuff I was doing before I found omitField

PR #75, opened 2017-01-20
[JENKINS-41121] Use the SCMHeadMigration API

stephenc:
Downstream of https://github.com/jenkinsci/scm-api-plugin/pull/24

See [JENKINS-41121](https://issues.jenkins-ci.org/browse/JENKINS-41121)

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -76,6 +76,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.logging.Level;
+import java.util.logging.LogRecord;
 import java.util.logging.Logger;
 import javax.annotation.Nonnull;
 import javax.servlet.ServletException;
@@ -86,6 +87,7 @@ import jenkins.scm.api.SCMEvent;
 import jenkins.scm.api.SCMEventListener;
 import jenkins.scm.api.SCMHead;
 import jenkins.scm.api.SCMHeadEvent;
+import jenkins.scm.api.SCMHeadMigration;
 import jenkins.scm.api.SCMHeadObserver;
 import jenkins.scm.api.SCMRevision;
 import jenkins.scm.api.SCMSource;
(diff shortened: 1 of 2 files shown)


src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -201,6 +203,48 @@ public void onLoad(ItemGroup<? extends Item> parent, String name) throws IOExcep
...
+                    LOGGER.log(Level.INFO, "Job {0}: a plugin upgrade is requesting migration of branch from {1} to {2}",
+                            new Object[]{item.getFullName(), oldHead.getClass(), newHead.getClass()}
+                    );
+                    try {
+                        Branch newBranch = new Branch(oldBranch.getSourceId(), newHead, oldBranch.getScm(),
+                                oldBranch.getProperties());
+                        newBranch.setActions(oldBranch.getActions());
+                        factory.setBranch(item, newBranch);
+                        SCMRevision revision = factory.getRevision(item);
+                        factory.setRevisionHash(item, SCMHeadMigration.readResolveSCMRevision(source, revision));
```
jglick: Wait, so `readResolveSCMRevision` will never be called unless `readResolveSCMHead` is implemented to modify the head? If so, I do not recall this being emphasized in the API.

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -201,6 +203,48 @@ public void onLoad(ItemGroup<? extends Item> parent, String name) throws IOExcep
...
+                    } catch (IOException | RuntimeException e) {
+                        LogRecord lr = new LogRecord(Level.WARNING,
+                                "Job {0}: Could not complete migration of branch from type {1} to {2}. "
+                                        + "The side-effect of this is that the next index may trigger a rebuild "
+                                        + "of the job (after which the issue will be resolved)");
+                        lr.setThrown(e);
+                        lr.setParameters(new Object[]{item.getFullName(), oldHead.getClass(), newHead.getClass()});
+                        LOGGER.log(lr);
```
jglick: Overkill, easier to use `log(Level, String, Throwable)`.

src/main/java/jenkins/branch/MultiBranchProject.java
stephenc: if you change the head you most likely need to change the revision too (as the revision includes the head)... thus if the head is changed we change the revision...

I guess we could modify the contract to resolve the revision too for future fix-up

PR #76, opened 2017-01-24
[JENKINS-41171] All OrganizationFolder instances should be single origin

stephenc:
See [JENKINS-41171](https://issues.jenkins-ci.org/browse/JENKINS-41171)

See [Blue Ocean embedding hard-coded assumption of single origin](https://github.com/jenkinsci/blueocean-plugin/blob/166b8f4e2a90037f212efac4ae4973cb53d825d7/blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubOrganizationFolder.java#L24)

While the test cases seemed to imply that multi-organization folders were an explicit design decision in Branch API 1.x, it seems to be more an accident from mirroring Multibranch Projects rather than a deliberate design decision.

Given that there are other plugins making the assumption of single origin, rather than expose the multi-origin use case, it seems more pragmatic to lock down the single origin for all Organization Folder instances that are single origin.

e.g. here is what a single origin GitHub Organization Folder will look like with this change:

screen shot 2017-01-24 at 14 55 42

Any users that have been using the multi-origin will still see the multi-origin behaviour as before:

screen shot 2017-01-24 at 14 53 57

But with a warning advising that the configuration is not recommended. As soon as the configuration is changed to single origin and the organization folder saved, the preferred locked behaviour will be applied.

It is likely that we may, at some future point in time, re-enable the multi-origin functionality for example:

* to allow for preparation and testing of changes before committing to non-DVCS SCM systems or SCM systems where it is non-trivial to offer in-place editing of the files from Blue Ocean; 

* to allow layering on code review systems such as [ReviewBoard](https://www.reviewboard.org/) on top of non-DVCS SCM Hosting solutions such as [Trac](https://trac.edgewall.org/wiki/TracSubversion) - though whether those specific projects are active enough is a separate question. The idea would be that one navigator is responsible for "hosting" the subversion / SCM repositories and the second system is responsible for "hosting" the change requests. A custom `SCM` implementation would delegate to the backing `SCM` to check out the baseline files and then use `patch` to apply the `.patch` that had been uploaded to the code review system. *Obviously DVCS is superior in that you do not have to manually upload `.patch` files and you can just push the proposed changes to the review server*

However, at this time there is no use-case driving the need for multi-origin and it is more likely to confuse users and create issues in downstream plugins.

@reviewbybees esp @i386 and @HRMPW


HRMPW:
👎  This change serves no purpose

vivek:
Code changes LGTM 🐝 From Blueocean perspective, UX is going to be more consistent and these are  UI changes so do not see any side effect as APIs remain untouched.

kshultzCB:
My migration tests have all passed. 🐝

PR #79, opened 2017-02-07
[JENKINS-41795] Report event origins

stephenc:
Downstream of https://github.com/jenkinsci/scm-api-plugin/pull/27

See [JENKINS-41795](https://issues.jenkins-ci.org/browse/JENKINS-41795)

Reports the origin of events, the event logs will now look something like:

screen shot 2017-02-07 at 11 56 09

And the tooltip for a branch event will include the origin if known:

screen shot 2017-02-07 at 11 57 40

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/branch/BranchEventCause.java
+++ b/src/main/java/jenkins/branch/BranchEventCause.java
@@ -26,6 +26,7 @@
 package jenkins.branch;
 
 import edu.umd.cs.findbugs.annotations.CheckForNull;
+import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.model.Cause;
 import hudson.model.ItemGroup;
 import hudson.model.Run;
@@ -42,6 +43,7 @@ public final class BranchEventCause extends Cause {
     private transient MultiBranchProject<?, ?> multiBranchProject;
 
     private final long timestamp;
+    @NonNull
     private final String origin;
 
     BranchEventCause(SCMEvent<?> event) {
@@ -64,6 +66,7 @@ public final class BranchEventCause extends Cause {
         return new Date(timestamp);
     }
 
+    @NonNull
     public String getOrigin() {
         return origin;
     }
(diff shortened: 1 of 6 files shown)


KostyaSha:
Exposing debug information to end users looks like overkill. Who cares about this IPs?

src/main/java/jenkins/branch/BranchEventCause.java
```diff
@@ -62,6 +64,10 @@ public Date getTimestamp() {
         return new Date(timestamp);
     }
 
+    public String getOrigin() {
```
jglick: `@CheckForNull`

src/main/resources/jenkins/branch/BranchEventCause/description.properties
```diff
@@ -1 +1,2 @@
 blurb=Branch event at {0,time} on {0,date}
+origin=Origin: {0}
```
jglick: Will this show `Origin: ?` if the origin is unknown?

stephenc: yes

stephenc:
@KostyaSha said:

> Exposing debug information to end users looks like overkill. Who cares about this IPs?

When the user is trying to determine why they have strange extra builds being triggered the IP address of the origin of the requests is vital.

There is no value in hiding the information from users as it will be in the logs that anyone with Item/READ can access anyway, and we cannot hide the origin in the logs from non-admins (plus the origin is expected to normally be the SCM system - given that the SCM links will be pointing to that system, its IP address is not exactly top secret)

The origin is only available as a tool-tip so not exactly in your face either

src/main/java/jenkins/branch/BranchEventCause.java
stephenc: `@NonNull` because the event will substitute `?`

PR #81, opened 2017-02-10
[FIXED JENKINS-41927] Orphaned branched should have name in strikethrough

stephenc:
See [JENKINS-41927](https://issues.jenkins-ci.org/browse/JENKINS-41927)

@reviewbybees 

screen shot 2017-02-10 at 15 12 00

### DIFF
--- a/src/main/resources/jenkins/branch/ItemColumn/column.jelly
+++ b/src/main/resources/jenkins/branch/ItemColumn/column.jelly
@@ -27,6 +27,13 @@ THE SOFTWARE.
          xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:i="jelly:fmt">
   <td style="${indenter.getCss(job)}">
     <j:choose>
+      <j:when test="${it.isOrphaned(job)}">
+        <s>
+          <a href="${jobBaseUrl}${job.shortUrl}" class='model-link inside' title="${it.getTitle(job)}">
+            <l:breakable value="${h.getRelativeDisplayNameFrom(job, itemGroup)}"/>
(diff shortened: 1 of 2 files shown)


myoung34:
@stephenc in the API is there a way to return this in the response for the job information?

stephenc:
@myoung34 if you get the config.xml of the job you can check the source id. There is a special null source for dead branches:

```
<properties>
<org.jenkinsci.plugins.workflow.multibranch.BranchJobProperty plugin="workflow-multibranch@2.12">
<branch plugin="branch-api@2.0.3-SNAPSHOT">
<sourceId>
::jenkins.scm.impl.NullSCMSource::
</sourceId>
```

Also the type should be `Branch.Dead`

src/main/resources/jenkins/branch/ItemColumn/column.jelly
```diff
@@ -27,6 +27,13 @@ THE SOFTWARE.
          xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:i="jelly:fmt">
   <td style="${indenter.getCss(job)}">
     <j:choose>
+      <j:when test="${it.isOrphaned(job)}">
+        <s>
+          <a href="${jobBaseUrl}${job.shortUrl}" class='model-link inside' title="${it.getTitle(job)}">
```
jtnord: am I missing something - shouldn't this have a new class or something to set the strikethough?

stephenc: `<s>`

PR #82, opened 2017-02-13
Allow implementers to throw IOException and InterruptedException

batmat:
@reviewbybees esp. @jglick


stephenc:
Approve in principle but need to consider if this needs a version bump to 2.1.0

jglick:
> need to consider if this needs a version bump

Why would it? The only caller is within the same source file, so it is compatible.

PR #84, opened 2017-02-14
[JENKINS-42009] Fix PCT for 2.32.2

varyvol:
[JENKINS-42009](https://issues.jenkins-ci.org/browse/JENKINS-42009)

- Bump plugin versions for PCT.
- Make some tests more realistic.

@reviewbybees @stephenc

### DIFF
--- a/src/test/java/integration/EventsTest.java
+++ b/src/test/java/integration/EventsTest.java
@@ -883,9 +883,9 @@ public class EventsTest {
             prj.scheduleBuild2(0).getFuture().get();
             r.waitUntilNoActivity();
             assertThat("Indexing applies dead branch cleanup",
-                    prj.getItems(), containsInAnyOrder(master));
-            assertThat("Indexing applies dead branch cleanup",
-                    prj.getItems(), not(containsInAnyOrder(feature)));
+                    prj.getItems(), allOf(
+                    containsInAnyOrder(master),
+                    not(containsInAnyOrder(feature))));
         }
     }
 
--- a/src/test/java/integration/EventsTest.java
+++ b/src/test/java/integration/EventsTest.java
@@ -794,7 +794,7 @@ public class EventsTest {
             FreeStyleProject master = prj.getItem("master");
             assertThat("We now have the master branch", master, notNullValue());
             FreeStyleProject feature = prj.getItem("feature");
-            assertThat("We now have the feature branch", master, notNullValue());
+            assertThat("We now have the feature branch", feature, notNullValue());
             assertThat("We only have the master and feature branches",
                     prj.getItems(), containsInAnyOrder(master, feature));
             r.waitUntilNoActivity();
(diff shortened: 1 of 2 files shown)


src/test/java/integration/EventsTest.java
```diff
@@ -883,7 +884,8 @@ public void given_multibranchWithSourcesAndTwoBranches_when_indexing_then_deadBr
             r.waitUntilNoActivity();
             assertThat("Indexing applies dead branch cleanup",
                     prj.getItems(), containsInAnyOrder(master));
-
+            assertThat("Indexing applies dead branch cleanup",
+                    prj.getItems(), not(containsInAnyOrder(feature)));
```
jglick: or

```java
assertThat("Indexing applies dead branch cleanup",
    prj.getItems(), allOf(
        containsInAnyOrder(master),
        not(containsInAnyOrder(feature))));
```

src/test/java/integration/EventsTest.java
```diff
@@ -1199,7 +1201,6 @@ public void given_orgFolder_when_someReposMatch_then_scanCreatesMatchingProjects
                     prj.getItems(),
                     Matchers.<MultiBranchProject<?, ?>>containsInAnyOrder(foo));
             assertThat("The matching branch exists", foo.getItem("master"), notNullValue());
-
```
jglick: gratuitous

src/test/java/integration/EventsTest.java
varyvol: Thanks @jglick

jglick:
You forgot to `import` it though.

varyvol:
@jglick right, sorry. It's fixed now (and passing on CI).

kwhetstone:
Have you checked if this would work with the most recent version of the PCT?  The failure I ran into was a missing `GitSampleRepoRule` class which required inclusion of `git`'s `tests` classifier.  This fixes that issue?

varyvol:
@kwhetstone this "PCT" consists only in changing the baseline version. The only problem here was that newer versions of some plugins were required, nothing similar to that you describe.

kwhetstone:
:bee:  I have some changes for the PCT as well to fix the issues when the plugin is run through the new PCT.

src/test/java/integration/EventsTest.java
```diff
@@ -882,8 +884,9 @@ public void given_multibranchWithSourcesAndTwoBranches_when_indexing_then_deadBr
             prj.scheduleBuild2(0).getFuture().get();
             r.waitUntilNoActivity();
             assertThat("Indexing applies dead branch cleanup",
-                    prj.getItems(), containsInAnyOrder(master));
```
stephenc: Reverting this as `containsInAnyOrder(master)` checks that the contents are `master` and nothing else so the `allOf` is redundant

varyvol: @stephenc sorry but could you point me to the place where that is being checked? Just so that I am aware.

stephenc: http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#containsInAnyOrder(T...)

> Creates an order agnostic matcher for Iterables that matches when a single pass over the examined Iterable yields a series of items, each logically equal to one item anywhere in the specified items. For a positive match, the examined iterable must be of the same length as the number of specified items.
> N.B. each of the specified items will only be used once during a given examination, so be careful when specifying items that may be equal to more than one entry in an examined iterable.


So try 

`assertThat(Arrays.asList("foo", "bar"), containsInAnyOrder("bar")))` 

and watch it fail

`assertThat(Arrays.asList("foo", "bar"), containsInAnyOrder("bar","foo")))` 

Will pass, while

`assertThat(Arrays.asList("foo", "bar"), containsInAnyOrder("manchu", "bar", "foo")))` 

will fail.

So you were over-complicating the reported failure in the event of a mismatch.

stephenc: The reason for using `containsInAnyOrder(_some_single_item_)` rather than `is(Collections.singleton(_some_single_item_))` is that you hit a lot of trouble trying to mash the generic collection types together, whereas `containsInAnyOrder(_some_single_item_)` doesn't care what type of collection you have been given

varyvol: @stephenc thanks for the clarification. Noted.

PR #85, opened 2017-02-14
[JENKINS-42000] Communication errors in organization scanning should not delete all everything

stephenc:
See [JENKINS-42000](https://issues.jenkins-ci.org/browse/JENKINS-42000)

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.19</version>
+    <version>2.21</version>
     <relativePath />
   </parent>
 
@@ -66,7 +66,8 @@
 
   <properties>
     <jenkins.version>1.642.3</jenkins.version>
-    <scm-api.version>2.0.3</scm-api.version>
+    <scm-api.version>2.0.4-20170214.145531-1</scm-api.version>
+    <hpi-plugin.version>1.121</hpi-plugin.version> <!-- TODO remove with the timestamp -SNAPSHOT -->
   </properties>
 
   <repositories>
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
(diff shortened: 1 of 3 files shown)


stephenc:
Downstream of https://github.com/jenkinsci/scm-api-plugin/pull/29

KostyaSha:
So simple fix requires changing 3 plugins???

stephenc:
@KostyaSha no, I technically didn't need to fix GHBS or BBBS but it is better to keep them on up to date dependencies, so for them I am fixing the compiler error proactively

stephenc:
@KostyaSha and when you separate the API from the implementation, a change to the API contract will always have downstream impact. 

In this case the only place where it is *runtime* required to fix the plugin is in the `OrganizationFolder`. In fact I could probably have hacked by either throwing a `RuntimeException` and then catching it again or by using an `AtomicReference` to stash the first exception... but other consumers of the SCM API would then have to repeat the "hack"

pom.xml
```diff
@@ -66,7 +66,8 @@
 
   <properties>
     <jenkins.version>1.642.3</jenkins.version>
-    <scm-api.version>2.0.3</scm-api.version>
+    <scm-api.version>2.0.4-20170214.145531-1</scm-api.version>
+    <hpi-plugin.version>1.121</hpi-plugin.version> <!-- TODO remove with the timestamp -SNAPSHOT -->
```
jglick: Just update the parent to 2.22.

pom.xml
```diff
@@ -66,7 +66,8 @@
 
   <properties>
     <jenkins.version>1.642.3</jenkins.version>
-    <scm-api.version>2.0.3</scm-api.version>
+    <scm-api.version>2.0.4-20170214.145531-1</scm-api.version>
```
jglick: recommend linking to the upstream PR

src/test/java/integration/EventsTest.java
```diff
@@ -1204,6 +1204,111 @@ public void given_orgFolder_when_someReposMatch_then_scanCreatesMatchingProjects
     }
 
     @Test
```
jglick: `@Issue`?

PR #88, opened 2017-02-21
[FIXED JENKINS-42234] Add the isMatch guard

stephenc:
See [JENKINS-42234](https://issues.jenkins-ci.org/browse/JENKINS-42234)

@reviewbybees


stephenc:
> BTW, should be the FINE/FINEST logging for both if and else cases?

translates as: "Would it make sense to add FINE/FINEST logging for both if and else cases?"

My answer is that it would only make sense if we are facing a lot of issues diagnosing why events not triggering jobs that should be triggered and the lack of trigger is because of a failed match against the repository name. I do not currently see this as a risk, but permission is hereby given for @oleg-nenashev to do the dance of I told you so when I am proved wrong

Dance of I told you so

PR #89, opened 2017-02-27
[JENKINS-37364] Tabs should indicate the number of items they have

stephenc:
See [JENKINS-37364](https://issues.jenkins-ci.org/browse/JENKINS-37364)

Before:
before

After:
after

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProjectViewHolder.java
+++ b/src/main/java/jenkins/branch/MultiBranchProjectViewHolder.java
@@ -223,7 +223,7 @@ public class MultiBranchProjectViewHolder extends AbstractFolderViewHolder {
          */
         @Override
         public String getDisplayName() {
-            return category.getDisplayName().toString();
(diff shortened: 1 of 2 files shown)


src/main/java/jenkins/branch/MultiBranchProjectViewHolder.java
```diff
@@ -223,7 +223,7 @@ public ViewImpl(ViewGroup owner, @NonNull SCMHeadCategory category) {
          */
         @Override
         public String getDisplayName() {
-            return category.getDisplayName().toString();
+            return category.getDisplayName().toString() + " (" + getItems().size() + ")";
```
jglick: Is there any significant cost to calling `getItems()` here, particularly for a tab we are not currently displaying?

stephenc: Should not be, as all items are in memory. There is allocation of an array and deduplication, so not the kind of thing you'd want to do in a tight loop, but in a UI request it should be reasonably cheap

PR #94, opened 2017-03-07
[JENKINS-42511] Add a test to verify root cause

stephenc:
See [JENKINS-42511](https://issues.jenkins-ci.org/browse/JENKINS-42511)

Downstream of https://github.com/jenkinsci/cloudbees-folder-plugin/pull/92

@reviewbybees

### DIFF
--- a/src/test/java/integration/EventsTest.java
+++ b/src/test/java/integration/EventsTest.java
@@ -28,14 +28,20 @@ package integration;
 import com.cloudbees.hudson.plugins.folder.computed.DefaultOrphanedItemStrategy;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import hudson.AbortException;
+import hudson.Functions;
+import hudson.model.Computer;
+import hudson.model.Executor;
 import hudson.model.FreeStyleProject;
+import hudson.model.OneOffExecutor;
+import hudson.model.Queue;
 import hudson.model.Result;
 import hudson.model.TopLevelItem;
+import hudson.model.queue.QueueTaskFuture;
 import integration.harness.BasicMultiBranchProject;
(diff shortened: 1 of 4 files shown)


src/test/java/integration/EventsTest.java
```diff
@@ -778,6 +780,122 @@ public void given_multibranchWithSources_when_createEventForExistingBranch_then_
     }
 
     @Test
```
jglick: Presumably you want to `@Ignore` it until the fix arrives.

src/test/java/integration/EventsTest.java
```diff
@@ -778,6 +780,122 @@ public void given_multibranchWithSources_when_createEventForExistingBranch_then_
...
+            FreeStyleProject master = prj.getItem("master");
+            waitUntilNoActivityIgnoringThreadDeathUpTo(10000);
+
+            assertThat(prj.getComputation().getEventsFile().lastModified(), greaterThan(lastModified));
+            assertThat("The master branch was built once only", master.getLastBuild().getNumber(), is(1));
+            assertThat(master.getLastBuild().getResult(), is(Result.SUCCESS));
+            List<Throwable> deaths = new ArrayList<Throwable>();
+            for (Computer comp : r.jenkins.getComputers()) {
+                for (Executor e : comp.getExecutors()) {
+                    if (e.getCauseOfDeath() != null) {
```
jglick: I would suggest just taking #92 first, which is simple and I think uncontroversial on its own (would defend against all manner of other mistakes), and then adapting this to verify that the revised lesser symptoms do not occur—namely, a failed indexing run.

stephenc:
@jglick do you want to re-review now that I have the fix

PR #96, opened 2017-03-16
[JENKINS-41736] Leverage the new event description API to expose event descriptions

stephenc:
@reviewbybees


src/main/resources/jenkins/branch/BranchEventCause/description.properties
```diff
@@ -1,2 +1,2 @@
-blurb=Branch event at {0,time} on {0,date}
+blurb={0} at {1,time} on {1,date}
```
jglick: Fortunately there are no localizations yet—if there were, this would be incompatible.

PR #97, opened 2017-04-10
[JENKINS-43433] Add more tests now that we can do merge PRs

stephenc:
Downstream of https://github.com/jenkinsci/scm-api-plugin/pull/35

@reviewbybees

Todo:

- [ ] Update to non-SNAPSHOT dependency once scm-api plugin released


stephenc:
Subsumed by #104

PR #98, opened 2017-04-13
[JENKINS-38837] Add a configuration option for workspaces on master

rsandell:
[JENKINS-38837](https://issues.jenkins-ci.org/browse/JENKINS-38837)

Because of the fix for [JENKINS-34564](https://issues.jenkins-ci.org/browse/JENKINS-34564) multi branch workspace names don't really fit into how normally workspace paths on the master are calculated.
So instead this introduces the option to configure where to put them as that would solve the original reporters issue with needing to put them on another drive/path than JENKINS_HOME.

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
+++ b/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
@@ -26,23 +26,28 @@ package jenkins.branch;
 
 import com.google.common.base.Charsets;
 import hudson.Extension;
+import hudson.ExtensionList;
 import hudson.FilePath;
-import hudson.model.Item;
-import hudson.model.Node;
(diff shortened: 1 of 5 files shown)


rsandell:
Test failed because it ran our of perm gen.

jglick:
Switched it to JDK 8.

rsandell:
Well, I've tried a couple of approaches, but haven't been able to come up with a solution that doesn't reintroduce JENKINS-34564 or pollutes the jobs directory.

jglick:
I am looking into whether it is feasible to use the existing core configuration.

jglick:
#103 is my counterproposal.

stephenc:
Closing in favour of @jglick 's proposal

PR #99, opened 2017-04-17
Remove final from OrganizationFolder declaration

i386:
This is so we can make `mock(OrganizationFolder.class)` possible as you cannot reliably mock final classes using Mockito.

We want to write some tests for `io.jenkins.blueocean.blueocean_github_pipeline.GithubPipelineCreateRequest` without having to spin up `JenkinsRule` and incur overhead when running these tests.

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/branch/OrganizationFolder.java
+++ b/src/main/java/jenkins/branch/OrganizationFolder.java
@@ -115,7 +115,7 @@ import static jenkins.scm.api.SCMEvent.Type.UPDATED;
  */
 @Restricted(NoExternalUse.class) // not currently intended as an API
 @SuppressWarnings({"unchecked", "rawtypes"}) // mistakes in various places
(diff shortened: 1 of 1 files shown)


i386:
@jglick could put a runtime check in the constructor?

jglick:
No, that would then fail under Mockito I am afraid.

For code which we expect to be called only from tests we can use

```java
@Restricted(NoExternalUse.class)`
```

but unfortunately this does not work to prevent subclassing, as there is no good way in Java to define distinct constructors for subclasses calling `super` vs. outside code instantiating the original type. (This general weakness is more commonly seen in the difficulty marking one of the two uses, but not the other, `@Deprecated`.)

jtnord:
@i386 - Check out powermock - works with mockito and will most final and static methods.

i386:
@jtnord ahh that would be good. Ill try that and reopen this PR if we have any trouble

PR #100, opened 2017-06-05
JENKINS-44676 support for env TAG_NAME when head type is tag

tanskann:
https://issues.jenkins-ci.org/browse/JENKINS-44676

Support for TAG_NAME env variable when head is TagSCMHead.

I'm working on building tags/releases with branch source plugin and this is needed so that pipeline is populated with TAG_NAM env variable.

### DIFF
--- a/src/main/java/jenkins/branch/BranchNameContributor.java
+++ b/src/main/java/jenkins/branch/BranchNameContributor.java
@@ -33,6 +33,7 @@ import hudson.model.TaskListener;
 import java.io.IOException;
 import jenkins.scm.api.mixin.ChangeRequestSCMHead;
 import jenkins.scm.api.SCMHead;
+import jenkins.scm.api.mixin.TagSCMHead;
 import jenkins.scm.api.actions.ChangeRequestAction;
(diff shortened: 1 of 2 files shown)


nfalco79:
Exception for:
* code formating (spaces vs tab and spaces between command)
* no a unit test (have a look to `EnvironmentTest`) to test this changes

I'm waiting this feature by a lot of time.

tanskann:
Added test and converted tabs to spaces.

jglick:
Is this actually necessary? The same information is already available as `$BRANCH_NAME`.

nfalco79:
I have to check again but in my last pipeline script I needed to know if the build was triggered due a tag or not and the `$BRANCH_NAME` was not set or its value was null. So actually I had duplicate the multibranch job with a pipeline job and add a StringParameter $TAG_NAME to run build on tags.

tanskann:
Yes same info is available in `BRANCH_NAME` but if the pipeline would contain `TAG_NAME` it would be easier to do things on the pipeline based on that info.

PR #103, opened 2017-06-12
[JENKINS-38837] Handle custom master workspace directory

jglick:
Unlike #98 this adds no UI and does not require users to make any configuration changes. It simply interprets `${ITEM_FULLNAME}` as a sanitized path rather than `Item.fullName`.

Also includes a fix for workspace deletion with custom `PATH_MAX` which I noticed; I really needed to fix that before implementing this anyway.

In the longer term, in JENKINS-2111 I have a proposal for a unified system.

@reviewbybees


jglick:
@reviewbybees done but would be good for @rsandell to agree that this can supersede #98.

jglick:
> workspaces from older versions of Jenkins will linger around after an upgrade

This is a general problem for all sorts of changes to workspace handling. My semi-proposal in JENKINS-2111 would handle this kind of issue better I think.

PR #105, opened 2017-06-28
Display stack traces in logs in a more readable order

jglick:
Would make the error in [JENKINS-45142](https://issues.jenkins-ci.org/browse/JENKINS-45142) more legible.

@reviewbybees


jglick:
@reviewbybees done

(looks like you need a `Jenkinsfile`!)

PR #106, opened 2017-07-13
[JENKINS-45322] Prevent event processing when disabled

stephenc:
See [JENKINS-45322](https://issues.jenkins-ci.org/browse/JENKINS-45322)

Downstream of https://github.com/jenkinsci/cloudbees-folder-plugin/pull/101

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -548,7 +548,9 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
      * {@inheritDoc}
      */
     public void onSCMSourceUpdated(@NonNull SCMSource source) {
-        scheduleBuild(0, new BranchIndexingCause());
+        if (isBuildable()) {
+            scheduleBuild(0, new BranchIndexingCause());
+        }
     }
 
     /**
@@ -644,6 +646,11 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
 
     private void scheduleBuild(BranchProjectFactory<P, R> factory, final P item, SCMRevision revision,
                                TaskListener listener, String name, Cause[] cause, Action... actions) {
+        if (!isBuildable()) {
+            listener.getLogger().printf("Did not schedule build for branch: %s (%s is disabled)%n",
+                    name, getDisplayName());
+            return;
+        }
         Action[] _actions = new Action[actions.length + 1];
         _actions[0] = new CauseAction(cause);
         System.arraycopy(actions, 0, _actions, 1, actions.length);
@@ -982,11 +989,22 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
      * {@inheritDoc}
      */
     @Override
+    protected boolean supportsMakeDisabled() {
+        if (sources == null || sources.isEmpty()) {
+            return false;
+        }
+        return super.supportsMakeDisabled();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
     public boolean isBuildable() {
         if (sources == null) {
             return false; // still loading
         }
-        return !sources.isEmpty();
+        return super.isBuildable() && !sources.isEmpty();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -982,11 +989,22 @@ public void run() {
      * {@inheritDoc}
      */
     @Override
+    protected boolean supportsMakeDisabled() {
+        if (sources == null || sources.isEmpty()) {
+            return false;
```
oleg-nenashev: Hmm, not sure why. But should not matter since it's empty anyway then

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -1671,6 +1705,16 @@ public void onSCMSourceEvent(SCMSourceEvent<?> event) {
                     try {
                         for (MultiBranchProject<?, ?> p : Jenkins.getActiveInstance()
                                 .getAllItems(MultiBranchProject.class)) {
+                            if (!p.isBuildable()) {
+                                LOGGER.log(Level.FINER,
```
oleg-nenashev: 🐜 Wrap by `LOGGER.isLoggable(Level.FINER)`, there are calculations inside

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -926,6 +945,16 @@ public void onSCMHeadEvent(SCMHeadEvent<?> event) {
                 if (CREATED == event.getType() || UPDATED == event.getType()) {
                     try {
                         for (OrganizationFolder p : Jenkins.getActiveInstance().getAllItems(OrganizationFolder.class)) {
+                            if (!p.isBuildable()) {
+                                LOGGER.log(Level.FINER,
```
oleg-nenashev: Check for isLoggable

src/main/java/jenkins/branch/MultiBranchProject.java
stephenc: not really calculations, only property access

src/main/java/jenkins/branch/OrganizationFolder.java
stephenc: property access only

src/main/java/jenkins/branch/MultiBranchProject.java
stephenc: Hide the enable / disable UI if no sources

src/main/java/jenkins/branch/MultiBranchProject.java
oleg-nenashev: `p.getFullName()` is not just a property access. The name is not even cached

stephenc: existing code already calls that method for similar logging statements

src/main/java/jenkins/branch/MultiBranchProject.java
jglick: This is actually bad because it means that you cannot mark a folder disabled in its `configure` screen when initially creating it (since the `sources` have not yet been saved), which is a desirable thing to be able to do if you want to get started with the configuration before actually making anything live. So I would suggest removing this override. Same in `OrganizationFolder`.

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -329,6 +332,22 @@ public boolean isHasEvents() {
      * {@inheritDoc}
      */
     @Override
+    protected boolean supportsMakeDisabled() {
+        return !navigators.isEmpty() && !projectFactories.isEmpty();
```
jglick: See above: undesirable.

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -1312,7 +1343,9 @@ public void complete() throws IllegalStateException, IOException, InterruptedExc
                                 } finally {
                                     bc.commit();
                                 }
-                                existing.scheduleBuild(cause());
+                                if (isBuildable()) {
```
jglick: see above

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -1350,7 +1383,9 @@ public void complete() throws IllegalStateException, IOException, InterruptedExc
                                 bc.commit();
                             }
                             observer.created(project);
-                            project.scheduleBuild(cause());
+                            if (isBuildable()) {
```
jglick: see above

PR #108, opened 2017-08-10
fix duration calculation

sbartram:
Calculation was incorrectly using division instead of multiplication


sbartram:
Sorry, I forgot to update the unit test - will fix now

sbartram:
after looking at the unit tests, I realize my reading of the documentation was wrong.  I'll suggest updates to clarify

PR #109, opened 2017-08-15
[JENKINS-37491] Fixing OverrideIndexTriggersJobProperty's ability to override event triggers

SpencerMalone:
Ticket: https://issues.jenkins-ci.org/browse/JENKINS-37491

The addition of BranchEventCause as a filtering cause for the NoTrigger rules accidentally made OverrideIndexTriggersJobProperty do more than just override index triggers. Initially when I did this, I had created a whole UI option before discovering OverrideIndexTriggersJobProperty. The UI approach didn't fit well w/ the property system found here, and the logic for when a build triggered immediately turned into a rat's nest.

In the end, even though it's technically a breaking API change for anyone that relied on OverrideIndexTriggersJobProperty to override index triggers AND event triggers, I think this is the best way forward, and makes the most logical sense.

Opinions?


SpencerMalone:
Note, this is missing tests at the moment. Still feeling out the test harness. Was hoping for some feedback before I really dive into that whole process.

jglick:
As discussed in JIRA, I doubt this approach is correct. The issue is not triggers, it is what is considered an interestingly distinct merge revision that deserves a fresh build.

SpencerMalone:
Yea, I'll be updating this PR shortly to support plugins that will allow custom behavior in the interim period until we have the idea of "interesting revision" implemented.

PR #110, opened 2017-08-23
[JENKINS-37491] Splitting OverrideIndexTriggersJobProperty up to support plugins

SpencerMalone:
Extending OverrideTriggerProperty will now let you stack multiple of these fancy properties together.

@jglick - Thoughts on this approach and then creating plugins as outlined here: https://issues.jenkins-ci.org/browse/JENKINS-37491?focusedCommentId=310597&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-310597

It was my first time fooling around w/ the extension annotation, and I struggled to define an extension that used the descriptor pattern and was a child of another extension (JobProperty). This is what I ended up with instead?

Todo list:

- [x]  Create the OverrideTriggerProperty
- [x]  Implement in the existing triggers for multibranch + pipeline jobs
- [ ]  Implement in subclasses of OverrideTriggerProperty
- [ ]  Unit tests for OverrideTriggerProperty

### DIFF
--- a/src/main/java/jenkins/branch/NoTriggerBranchProperty.java
+++ b/src/main/java/jenkins/branch/NoTriggerBranchProperty.java
@@ -76,14 +76,13 @@ public class NoTriggerBranchProperty extends BranchProperty {
                                 Job<?,?> j = (Job) p;
                                 if (j.getParent() instanceof MultiBranchProject) {
 
-                                    OverrideIndexTriggersJobProperty overrideProp = j.getProperty(OverrideIndexTriggersJobProperty.class);
-                                    if (overrideProp != null) {
-                                        return overrideProp.getEnableTriggers();
-                                    } else {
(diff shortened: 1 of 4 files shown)


stephenc:
I am inclined to reject this PR. Fundamentally the `NoTriggerOrganizationFolderProperty`/`OverrideIndexTriggersJobProperty` thing is a hack, and not even a good hack. The correct implementation should have been through `BranchBuildStrategy`. The block on removing this is that I need to find a way to move people using `NoTriggerOrganizationFolderProperty` to a concrete `BranchBuildStrategy` without forcing new and existing users of branch-api (that do not use this feature) to be forced to install a `branch-build-standard-strategies` plugin.

I think I have a way to achieve that, but it will need some experimenting. As such I would much rather remove this completely rather than building more features on top of this tower.

If you think there is a strong case for continuing with this PR, please state your case, otherwise unless I see comments I intend to close this PR some time after 1 week from now (depending on when I next review PRs)

PR #113, opened 2017-10-06
[FIXED JENKINS-47311] Fix the broken form submission and add the missing form support for org folders

stephenc:
See [JENKINS-43711](https://issues.jenkins-ci.org/browse/JENKINS-47311)

@reviewbybees

### DIFF
--- a/src/main/java/jenkins/branch/OrganizationFolder.java
+++ b/src/main/java/jenkins/branch/OrganizationFolder.java
@@ -130,6 +130,12 @@ public final class OrganizationFolder extends ComputedFolder<MultiBranchProject<
      * Our project factories.
      */
     private final DescribableList<MultiBranchProjectFactory,MultiBranchProjectFactoryDescriptor> projectFactories = new DescribableList<MultiBranchProjectFactory,MultiBranchProjectFactoryDescriptor>(this);
+    /**
+     * The rules for automatic building of branches.
+     *
+     * @since 2.0.12
+     */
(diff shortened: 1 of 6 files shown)


src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -153,6 +159,13 @@
     private transient String facDigest;
 
     /**
+     * The {@link #buildStrategies} digest used to detect if we need to trigger a rescan on save.
+     *
+     * @since 2.0.12
+     */
+    private transient String bbsDigest;
```
svanoort: Might be better to have an equals/hashCode operation defined for the buildStrategies and use that for comparison, but this gets the job done.

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -282,13 +300,24 @@ public boolean isSingleOrigin() {
     }
 
     /**
+     * The {@link BranchBuildStrategy}s to apply.
+     *
+     * @return The {@link BranchBuildStrategy}s to apply.
+     * @since 2.0.12
+     */
+    public DescribableList<BranchBuildStrategy, BranchBuildStrategyDescriptor> getBuildStrategies() {
```
svanoort: 🐜  @CheckForNull due to deserializing old format.

stephenc: nope. Top level project type, so constructor is always called and then unmarshal on top

PR #114, opened 2017-10-21
Add Chinese translation

LinuxSuRen:


stephenc:
Random spot-check with google translate appears to validate these... merging

PR #118, opened 2017-12-04
[FIXED JENKINS-48090] Merge CauseActions when scheduling

stephenc:
- Should have been specified that SCMSource.fetchActions should not return CauseAction, but we missed that so let's just handle the merge as it is not a major crime

See [JENKINS-48090](https://issues.jenkins-ci.org/browse/JENKINS-48090)

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -653,9 +653,35 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
                     name, getDisplayName());
             return;
         }
-        Action[] _actions = new Action[actions.length + 1];
-        _actions[0] = new CauseAction(cause);
-        System.arraycopy(actions, 0, _actions, 1, actions.length);
+        // JENKINS-48090 see Queue.Item.getCauses() which only operates on the first CauseAction
+        // We need to merge any additional causes with the causes we are supplying
+        int causeCount = 0;
+        for (Action action : actions) {
+            if (action instanceof CauseAction) {
+                causeCount++;
+            }
+        }
+        Action[] _actions;
+        if (causeCount == 0) {
+            // no existing causes, insert new one at start
+            _actions = new Action[actions.length + 1];
+            _actions[0] = new CauseAction(cause);
+            System.arraycopy(actions, 0, _actions, 1, actions.length);
+        } else {
+            // existing causes, filter out and insert aggregate at start
+            _actions = new Action[actions.length + 1 - causeCount];
+            int i = 1;
+            List<Cause> causeList = new ArrayList<>();
+            Collections.addAll(causeList, cause);
+            for (Action a: actions) {
+                if (a instanceof CauseAction) {
+                    causeList.addAll(((CauseAction) a).getCauses());
+                } else {
+                    _actions[i++] = a;
+                }
+            }
+            _actions[0] = new CauseAction(causeList);
+        }
         if (ParameterizedJobMixIn.scheduleBuild2(item, 0, _actions) != null) {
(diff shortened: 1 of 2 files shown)


KostyaSha:
@stephenc events can contribute own causes, while retrieve can't. So it not a crime, but a similar ability to pass additional causes to final build.

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -653,9 +653,35 @@ private void scheduleBuild(BranchProjectFactory<P, R> factory, final P item, SCM
                     name, getDisplayName());
             return;
         }
```
KostyaSha: also please rename argument `Cause[] cause` to `causes` (events already passing multiple causes here)

stephenc:
> events can contribute own causes, while retrieve can't. So it not a crime, but a similar ability to pass additional causes to final build.

What I am saying is that I should have specified that `SCMSource.fetchActions` should not return `CauseAction`, because if you want to trigger a build for a different cause, you should just trigger the build normally... in which case you get to supply all your own causes yourself.

So from the point of view of my preferred specification of SCM API, I would much rather that it not be abused to inject `CauseAction`... but it was my mistake not to specify that originally... and it is not the biggest crime in the universe... and shouldn't have any major unintended consequences... but you are probably banging your head against stuff because you are holding the hammer wrong

KostyaSha:
> you should just trigger the build normally...

I would be happy to do it, but for branch/pr that didn't exist before it impossible because `childObserver` is hidden and headObserver is created every time new. And orphaned items cleanup may also remove jobs with direct triggering.

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -647,15 +647,41 @@ protected void computeChildren(final ChildObserver<P> observer, final TaskListen
...
+                               TaskListener listener, String name, Cause[] causes, Action... actions) {
         if (!isBuildable()) {
             listener.getLogger().printf("Did not schedule build for branch: %s (%s is disabled)%n",
                     name, getDisplayName());
             return;
         }
-        Action[] _actions = new Action[actions.length + 1];
-        _actions[0] = new CauseAction(cause);
-        System.arraycopy(actions, 0, _actions, 1, actions.length);
+        // JENKINS-48090 see Queue.Item.getCauses() which only operates on the first CauseAction
```
KostyaSha: That sounds at all that it was bad idea to wrap causes into action in core...

stephenc: Well we could argue that this is actually a bug in core... namely that it only unwraps the first `CauseAction` but I do not want to get into a fight on that

stephenc:
> I would be happy to do it, but for branch/pr that didn't exist before it impossible because childObserver is hidden and headObserver is created every time new. And orphaned items cleanup may also remove jobs with direct triggering.

I think we could let the `BranchBuildStrategy` contribute causes. Would that be an acceptable alternative? (BTW I also want to add the ability for `BranchBuildStrategy` to compare the previous revision so that you can have a strategy that doesn't rebuild PRs if the target branch changes)

stephenc:
@KostyaSha something like this in BranchBuildStrategy:

screen shot 2017-12-04 at 22 11 06

WDYT?

stephenc:
And we chalk the merging of `CauseAction` instances to a core bug and close this PR?

KostyaSha:
> And we chalk the merging of CauseAction instances to a core bug and close this PR?

imho PR is fine. 

> I think we could let the BranchBuildStrategy contribute causes.

I think scmsource should still be able to contribute. BranchBuildStrategy will work only for `isAutomatic` that is not sued in my case.

PR #120, opened 2017-12-13
[FIXED JENKINS-48536] OrganizationFolder should ensure MultibranchProject calls SCMSource.afterSave()

stephenc:
See [JENKINS-48536](https://issues.jenkins-ci.org/browse/JENKINS-48536)

### DIFF
--- a/src/main/java/jenkins/branch/OrganizationFolder.java
+++ b/src/main/java/jenkins/branch/OrganizationFolder.java
@@ -1368,7 +1368,12 @@ public final class OrganizationFolder extends ComputedFolder<MultiBranchProject<
                             return;
                         }
                         String folderName = NameEncoder.encode(projectName);
-                        MultiBranchProject<?, ?> existing = observer.shouldUpdate(folderName);
+                        // HACK: observer.shouldUpdate will restore the buildable flag of the child, so pre-inspect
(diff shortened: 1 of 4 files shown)


stephenc:
First the failing test case...

stephenc:
Then the fix in d4f20e4

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -1385,7 +1385,8 @@ public void complete() throws IllegalStateException, IOException, InterruptedExc
                                 } finally {
                                     bc.commit();
                                 }
-                                if (isBuildable()) {
+                                existing.fireSCMSourceAfterSave(existing.getSCMSources());
+                                if (isBuildable() && existing.isBuildable() && existing.updateDigests()) {
```
stephenc: NOTE: we should have been considering the MBP's `isBuildable()` once that method was added.

stephenc: NOTE: we now may suppress excessive indexing of the MBP if the digest is unchanged. Some people may view this as a regression if they were relying on the side-effect of a full scan of an `OrganizationFolder` triggering a full scan of all MBPs... but we also have users who were complaining... and there really is no reason to force a scan of the child MBPs (in fact we have users complaining about unnecessary child scanning)

PR #122, opened 2018-01-04
[JENKINS-47859] First half of migration code

stephenc:
https://issues.jenkins-ci.org/browse/JENKINS-47859


src/main/java/jenkins/branch/PropertyMigration.java
```diff
@@ -0,0 +1,185 @@
...
+import java.util.TreeSet;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import jenkins.model.Jenkins;
+import org.apache.commons.lang.StringUtils;
+import org.jvnet.localizer.Localizable;
+
+/**
+ * An extension point that captures the need for a complex migration of a folder property into some other configuration
+ * of the containing folder.
```
jglick: Why do we need this? A simple `@Initializer` should suffice.

src/main/resources/jenkins/branch/PropertyMigration/MonitorImpl/message.properties
```diff
@@ -0,0 +1,26 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+blurb=Some of your jobs are using deprecated functionality that has been replaced by functionality in a new plugin. <br/>\
+  Please install the following plugin(s):<pre>{0}</pre>
```
jglick: The admin monitor seems like overkill to me.

Suggest a simpler path. Add form validation for all this stuff saying that it is deprecated and to use the new plugin.  That plugin should convert existing folders in an initializer when it starts up, and use `DescriptorVisibilityFilter` to hide the old features. And we can add the new plugin to `platform-plugins.json`, `workflow-aggregator` dep, whatever, so that people will tend to get it without asking.

stephenc:
After some tricks I picked up on other plugins I have enhanced the admin monitor to allow for a guided installation of the required plugins:

screenshot 2018-11-16 at 14 00 56

Because of the widespread usage of this plugin, I feel the guided migration is a better plan than you would get with a plain `@Initializer` which is why I am introducing dedicated machinery to manage the migration.

tempora-mutantur:
Hi.
As the `Suppress automatic SCM triggering` property is depricated right now, could you, please, describe, how we can achieve the same behaviour with the `Named branch` build strategy. We used `Suppress automatic SCM triggering` for jobs that we don't want to run automatically, but want them to contain all pull requests/tags/branches and discover new of them automatically with scan/hooks updates, without launching them.

kevindelord:
Also looking for a way to achieve the same behaviour.
Thanks in advanced !

stephenc:
Something like this:

screenshot 2019-02-19 at 16 25 52

Will:

1. Discover Branches, Pull requests from forks and pull requests from the origin
2. Only build branches automatically, but will build all branches automatically. The pull requests will have jobs created for them but you will need to manually request a build from the Jenkins UI

For a counter example:

screenshot 2019-02-19 at 16 27 41

Which will:

1. Discover branches, pull requests (but only from the origin) and tags
2. Will only build the master branch and pull requests, but will not rebuild pull requests automatically if the only change was the target branch (which would allow the pull request validation to become stale but avoids build storms). The tags will have a job created for them, but the job will not be automatically built, instead you must manually trigger the job from the jenkins UI

HTH

kevindelord:
@stephenc thank you so much for your quick replies. You helped a lot!

It wasn't clear for us that by installing the plugin the automatic SCM triggering was suppressed by default and that we had to add build strategies to enable it for the named branches.

tempora-mutantur:
@stephenc As I understand, there is no way to implement the same behavior that was provided by the _Suppress automatic SCM triggering_ strategy? So, all branches/pull requests are discovered automatically and jobs for them don't launch builds because of SCM changes. I used such jobs a lot, want to know how to migrate to the _Build strategies_ plugin.

stephenc:
@tempora-mutantur not entirely clear what you mean. For any configuration you had previously, there exists a 1:1 mapping between that configuration and the new one using Build strategies. 

Note: there is a subset of configurations where we cannot programmatically infer the 1:1 mapping (specifically around automatically building PRs) but I assure you the corresponding configuration exists.

In addition, build strategies enables additional enhancements that were not previously possible.

—-

Or maybe you need to rephrase your use case so we can show how to set that up

tempora-mutantur:
@stephenc

I think that I've found the way to setup Branch Build Strategies to work for me.
I have a job that is BitBucket multibranch pipeline. We use this to deploy our product so I need it to have all branches/pull requests/tags.
I want them just to be discovered automatically but don't won't any of them to start automatically because of SCM changes.
So, I've added the "Named branches" strategy and excluded all the branches. So, as I see all branches/PRs/tags are discovered automatically and don't start because of scm changes.
Thank you.

PR #124, opened 2018-01-19
JENKINS-49051 - adding event field to BranchEventCause class

yurigo79:
I've added the event as a filed. So it is accessible from the getCause() method

### DIFF
--- a/src/main/java/jenkins/branch/BranchEventCause.java
+++ b/src/main/java/jenkins/branch/BranchEventCause.java
@@ -49,7 +49,7 @@ public final class BranchEventCause extends Cause {
 
     private final String description;
 
-    private final Object event;
+    private final SCMEvent event;
 
     BranchEventCause(@NonNull SCMEvent<?> event, @CheckForNull String description) {
         timestamp = event.getTimestamp();
@@ -79,7 +79,7 @@ public final class BranchEventCause extends Cause {
     }
 
     @NonNull
-    public Object getEvent(){
+    public SCMEvent getEvent(){
         return event;
     }
 
--- a/src/main/java/jenkins/branch/BranchEventCause.java
+++ b/src/main/java/jenkins/branch/BranchEventCause.java
@@ -49,10 +49,13 @@ public final class BranchEventCause extends Cause {
 
     private final String description;
 
+    private final Object event;
+
     BranchEventCause(@NonNull SCMEvent<?> event, @CheckForNull String description) {
         timestamp = event.getTimestamp();
         origin = event.getOrigin();
         this.description = StringUtils.isBlank(description) ? event.description() : description;
+        this.event = event;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/jenkins/branch/BranchEventCause.java
```diff
@@ -49,10 +49,13 @@
 
     private final String description;
 
+    private final Object event;
```
stephenc: would be better as `SCMEvent`... but irrespective, there is an issue with serialization as `SCMEvent` contract does not include serializability, so we may not be able to serialize it to store it

src/main/java/jenkins/branch/BranchEventCause.java
```diff
@@ -75,6 +78,11 @@ public String getOrigin() {
         return origin;
     }
 
+    @NonNull
+    public Object getEvent(){
```
stephenc: should be typed

yurigo79:
@stephenc I didn't get your point about serializability. Do you mean that BranchEventCause  is serializable now and it will not be serializable if I will add the SCMEvent field?

stephenc:
So the causes are serializable but I think I had wanted to allow events to be non-serializable so that they could contain cached contextual I formation that can be used to avoid excessive rate limit usage.

I need to cross-check with scm-api to determine whether this is a good plan at all. An alternative might be to allow events to add a bag of properties to a cause, as that way you don’t have a full event object... or perhaps have a memoization api that allows SCMEvent instances to self-scrub to a serializable implementation

src/main/java/jenkins/branch/BranchEventCause.java
```diff
@@ -75,6 +78,11 @@ public String getOrigin() {
         return origin;
     }
 
+    @NonNull
```
stephenc: Old causes will have a null event, this is going to have to be `@CheckForNull`

src/main/java/jenkins/branch/BranchEventCause.java
```diff
@@ -49,10 +49,13 @@
 
     private final String description;
 
+    private final SCMEvent event;
+
     BranchEventCause(@NonNull SCMEvent<?> event, @CheckForNull String description) {
         timestamp = event.getTimestamp();
         origin = event.getOrigin();
         this.description = StringUtils.isBlank(description) ? event.description() : description;
+        this.event = event;
```
stephenc: Since the event has to be `@CheckForNull`...
```suggestion
        this.event = event instanceof Serializable ? event : null;
```

src/main/java/jenkins/branch/BranchEventCause.java
bitwiseman:
```suggestion
    @CheckForNull
```

bitwiseman:
@yurigo79 
It looks like this change stalled out.   I'm going to close this PR for the moment and we can discuss design in the JIRA and then reopen this as appropriate.

PR #125, opened 2018-03-07
Trigger repository scans more often

smaslennikov:
...since I'm not ready to dive into fixing this properly. Solves
https://issues.jenkins-ci.org/browse/JENKINS-48960 (for us, at least)


stephenc:
Correct approach is to enable the configuration of this from an org folder property rather than hard-code to 1d.

Closing this issue as this global hack will affect everyone

PR #127, opened 2018-08-08
[FIXED JENKINS-50561] Add rateLimitBuilds symbol

abayer:
[JENKINS-50561](https://issues.jenkins-ci.org/browse/JENKINS-50561)

I wanted to avoid "throttle" in the symbol, since that's already a
thing over in Throttle Concurrent Builds, and "throttle" means
something different there than it does here.

### DIFF
--- a/src/main/java/jenkins/branch/RateLimitBranchProperty.java
+++ b/src/main/java/jenkins/branch/RateLimitBranchProperty.java
@@ -45,6 +45,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import jenkins.model.Jenkins;
 import net.sf.json.JSONObject;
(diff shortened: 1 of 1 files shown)


abayer:
Nope - good point. I'll work on that.

abayer:
@jglick Done - https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/284

abayer:
Debating whether `RateLimitBranchProperty.Throttle` should have a symbol too - honestly, I don't think so, if for no other reason than I can't think of a good name, and the `rateLimitBuilds(throttle: [count: 1, durationName: 'day', userBoost: false])` syntax is Good Enough.

PR #128, opened 2018-09-05
[FIXED JENKINS-38706] Add some tweaks to make the fix usable in the real world

stephenc:
Also adds experimental NameMangler alternative workspace locator implementation


rsandell:
Seems like #129 overwrote this.

PR #129, opened 2018-09-21
[JENKINS-2111] Managed workspace indices

jglick:
[JENKINS-2111](https://issues.jenkins-ci.org/browse/JENKINS-2111): the long-discussed overhaul of workspace selection at least for branch projects. Unlikely the previous technique from #50 (JENKINS-34564), the workspace path stays short (like [JENKINS-30148](https://issues.jenkins-ci.org/browse/JENKINS-30148)) as we do not need to add any cryptographic hash to prevent collisions between projects: workspace subdirectories are _allocated_ and tracked, so we can just add a simple uniquifying suffix as needed. That should allow [JENKINS-38706](https://issues.jenkins-ci.org/browse/JENKINS-38706) to be closed. Also handles renames ([JENKINS-22240](https://issues.jenkins-ci.org/browse/JENKINS-22240)).


pom.xml
```diff
@@ -65,7 +65,8 @@
   </scm>
 
   <properties>
-    <jenkins.version>1.642.3</jenkins.version>
+    <jenkins.version>2.73.3</jenkins.version>
```
jglick: Made various things more convenient, such as https://github.com/jenkinsci/jenkins/pull/2914.

pom.xml
```diff
@@ -85,16 +86,6 @@
   </pluginRepositories>
 
   <dependencies>
-    <dependency>
-      <groupId>org.kohsuke</groupId>
-      <artifactId>access-modifier-annotation</artifactId>
```
jglick: These should not have been here.

src/main/java/jenkins/branch/BranchProjectFactory.java
```diff
@@ -204,7 +204,7 @@ public P decorate(P project) {
             List<BranchProperty> properties = new ArrayList<BranchProperty>(branch.getProperties());
             Collections.sort(properties, DescriptorOrder.reverse(BranchProperty.class));
             for (BranchProperty property : properties) {
-                JobDecorator<P, R> decorator = property.jobDecorator(project.getClass());
+                JobDecorator<P, R> decorator = property.jobDecorator((Class) project.getClass());
```
jglick: Incomprehensible compiler errors in `-source 8`.

src/main/java/jenkins/branch/NameMangler.java
```diff
@@ -41,7 +41,7 @@
  */
 public final class NameMangler {
 
-    private static final int MAX_SAFE_LENGTH = 32;
+    static final int MAX_SAFE_LENGTH = 32;
```
jglick: This is now the default being applied to workspace lengths.

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -55,51 +71,153 @@
...
+     * The most characters to allow in a workspace directory name, relative to the root.
+     * Zero to disable altogether.
+     * Not used for new workspaces; only for continuing to use workspaces created prior to {@link #MAX_LENGTH}.
+     */
+    @Deprecated
     static /* not final */ int PATH_MAX = Integer.getInteger(WorkspaceLocatorImpl.class.getName() + ".PATH_MAX", PATH_MAX_DEFAULT);
 
+    static /* not final */ int MAX_LENGTH = Integer.getInteger(WorkspaceLocatorImpl.class.getName() + ".MAX_LENGTH", NameMangler.MAX_SAFE_LENGTH);
+
+    static /* not final */ boolean MULTIBRANCH_ONLY = Boolean.parseBoolean(System.getProperty(WorkspaceLocatorImpl.class.getName() + ".MULTIBRANCH_ONLY", /* TBD */ "true"));
```
jglick: Could be set to `false` to turn on the smarter behavior for all projects, including standalone Pipeline and even freestyle, matrix, etc. On the one hand I do not like having installation of what is otherwise a library plugin affect behavior of core functionality. On the other hand, the name shortening is very useful for projects in deep folder hierarchies; the character sanitization is useful for all sorts of project generation systems other than multibranch; and the automatic cleanup is useful for pretty much everyone, since `WorkspaceCleanupThread` is unreliable.

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -55,51 +71,153 @@
...
+    private static final String INDEX_FILE_NAME = "workspaces.txt";
+
     @Override
     public FilePath locate(TopLevelItem item, Node node) {
-        if (PATH_MAX == 0) {
+        return locate(item, node, true);
+    }
+
+    private static FilePath locate(TopLevelItem item, Node node, boolean create) {
+        if (MULTIBRANCH_ONLY && !(item.getParent() instanceof MultiBranchProject)) {
```
jglick: Maybe could be a three-way flag to disable the feature altogether?

jglick:
Test failure is due to a Windows FS problem which I am tracking down.

jglick:
Seems to be another Windows test failure I cannot reproduce locally and which looks unrelated to my changes. Will see if it goes away after planned test expansion commits.

svanoort:
@joseblas or @dwnusbaum Could one of you give this a peek please?  If you think it's solid, happy to see it released.

jglick:
If you are reviewing, then you should also check an update to the integration test in https://github.com/jenkinsci/workflow-multibranch-plugin/pull/82, plus the two other PRs that depends on.

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -53,53 +70,186 @@
...
-    /** copied from {@link Jenkins} */
-    static String expandVariablesForDirectory(String base, String itemFullName, String itemRootDir) {
-        return Util.replaceMacro(base, ImmutableMap.of(
-                "JENKINS_HOME", Jenkins.getActiveInstance().getRootDir().getPath(),
-                "ITEM_ROOTDIR", itemRootDir,
-                "ITEM_FULLNAME", itemFullName,   // legacy, deprecated
-                "ITEM_FULL_NAME", itemFullName.replace(':','$'))); // safe, see JENKINS-12251
+    private static void save(Map<String, String> index, FilePath workspace) throws IOException, InterruptedException {
+        // FilePath.renameTo does not support REPLACE_EXISTING, and AtomicFileWriter does not work remotely.
+        // Anyway we are synchronizing access to this file so the only potential problem is half-written content.
```
dwnusbaum: Maybe you could add something like the expected number of pairs to the first line of the file to help diagnose a corrupted index in `#load`, although with os/disk caches I'm not sure how likely it would be for that data to be written correctly if there is a problem with the rest of the file.

vivek:
@dwnusbaum good to merge this one?

stephenc:
As I understand, one of the biggest issues for Windows systems is too long a path, for some people even the 32 character mangled names may prove too long.

It would be awesome - now we have tracking - to allow a per-node configurable strategy (default is to use the name mangler, and only default provided here)

That way a Windows User could install a mangler that just picks an available one or two character name (failing up to 3+ characters) where they are willing to trade knowing which workspace is which from the name for shortest possible paths and having to look up in the index file

dwnusbaum:
@vivek It looks fine to me, but I am not very familiar with this API, so it would probably be best for @rsandell or @stephenc to review it.

jglick:
> allow a per-node configurable strategy […so] a Windows User could install a mangler that just picks an available one or two character name (failing up to 3+ characters)

Might be overkill. I think the new default maximum should be small enough for most users (assuming they use a reasonably concise `remoteFS`); if not, they can still run with

    -Djenkins.branch.WorkspaceLocatorImpl.MAX_LENGTH=10

or whatever.

Note that the max length includes any applied uniquification suffix, so it is really a hard limit¹ and you cannot set the max so small that an exponentially large number of workspaces would overflow. You would of course have to have an absurdly large number of Jenkins projects for that to be an issue.

¹This does _not_ apply to suffixes added by `WorkspaceList`, such as `@2`, or the `@tmp` added by various build steps.

jglick:
Of course we could also just set a smaller default max value on Windows. I do not have a strong opinion either way. Can always be tweaked later.

daniel-beck:
The logging here is really annoying with masters that don't have executors (so their master workspace path shouldn't matter). From ci.j.io:

	Oct 04, 2019 12:05:02 PM WARNING jenkins.branch.WorkspaceLocatorImpl getWorkspaceRoot
	JENKINS-2111 path sanitization ineffective when using legacy Workspace Root Directory ‘${ITEM_ROOTDIR}/workspace’; switch to ‘${JENKINS_HOME}/workspace/${ITEM_FULL_NAME}’ as in JENKINS-8446 / JENKINS-21942
	Oct 04, 2019 12:05:02 PM FINE jenkins.branch.WorkspaceLocatorImpl locate
	no available workspace root for hudson.model.Hudson@2ad98027 so skipping org.jenkinsci.plugins.workflow.job.WorkflowJob@615a99f0[Infra/infra-reports/PR-22]
	Oct 04, 2019 12:07:01 PM WARNING jenkins.branch.WorkspaceLocatorImpl getWorkspaceRoot
	JENKINS-2111 path sanitization ineffective when using legacy Workspace Root Directory ‘${ITEM_ROOTDIR}/workspace’; switch to ‘${JENKINS_HOME}/workspace/${ITEM_FULL_NAME}’ as in JENKINS-8446 / JENKINS-21942
	Oct 04, 2019 12:07:01 PM FINE jenkins.branch.WorkspaceLocatorImpl locate
	no available workspace root for hudson.model.Hudson@2ad98027 so skipping org.jenkinsci.plugins.workflow.job.WorkflowJob@723417aa[Infra/charts/master]
	Oct 04, 2019 12:07:01 PM WARNING jenkins.branch.WorkspaceLocatorImpl getWorkspaceRoot
	JENKINS-2111 path sanitization ineffective when using legacy Workspace Root Directory ‘${ITEM_ROOTDIR}/workspace’; switch to ‘${JENKINS_HOME}/workspace/${ITEM_FULL_NAME}’ as in JENKINS-8446 / JENKINS-21942
	Oct 04, 2019 12:07:01 PM FINE jenkins.branch.WorkspaceLocatorImpl locate
	no available workspace root for hudson.model.Hudson@2ad98027 so skipping org.jenkinsci.plugins.workflow.job.WorkflowJob@7b8287f7[Infra/acceptance-tests/install-lts-debian-package]

jglick:
@daniel-beck

> The logging here is really annoying with masters that don't have executors (so their master workspace path shouldn't matter).

No such log message should be printed for a system with no master executor, because this code path is only run when looking for a workspace on the master node. I cannot speculate as to the cause of such messages on ci.jenkins.io; would need to have some sort of debug access, or steps to reproduce from scratch.

PR #130, opened 2018-09-22
[JENKINS-45814] Changed suppress property behavior

lelik9:
Hi. I've little bit updated "suppress automatic SCM triggering" property to add more flexible behavior.
I've added ability to choice what kind of event should be suppressed.
image
For instance, if you have a lot of branch and most of them is triggered after the indexing we lose a lot of resources, and to avoid this would be better to have option to suppress only this type of trigger, but save main functionality and trigger builds by webhooks.


stephenc:
The plan is to remove this as the Branch Build Strategy replaces the need for the hack of "Suppress automatic SCM triggering", closing as I do not want to add more changes to deprecated code

stephenc:
See https://issues.jenkins-ci.org/browse/JENKINS-47859 and https://github.com/jenkinsci/branch-api-plugin/pull/122

PR #133, opened 2018-11-09
More logging for JENKINS-2111

jglick:
Adds some diagnostics to #129.


src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -157,7 +160,9 @@ private static FilePath locate(TopLevelItem item, String fullName, Node node, bo
                 // Already listed:
                 String path = index.get(fullName);
                 if (path != null) {
-                    return workspace.child(path);
+                    FilePath dir = workspace.child(path);
+                    LOGGER.finer("index already lists " + dir + " for " + item + " on " + node);
```
rsandell: You are using parameters on all other logging statements but not here?

PR #138, opened 2018-11-30
Updating some test deps

jglick:
Supersedes #57. Picks up https://github.com/jenkinsci/git-client-plugin/pull/381.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -70,7 +70,7 @@
     <jenkins.version>2.107.3</jenkins.version>
     <java.level>8</java.level>
     <scm-api.version>2.2.7</scm-api.version>
-    <git-plugin.version>3.3.0</git-plugin.version>
+    <git-plugin.version>4.0.0-beta3</git-plugin.version>
   </properties>
 
   <repositories>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -69,8 +69,8 @@
     <changelist>-SNAPSHOT</changelist>
     <jenkins.version>2.107.3</jenkins.version>
     <java.level>8</java.level>
-    <scm-api.version>2.2.7</scm-api.version>
-    <git-plugin.version>4.0.0-beta3</git-plugin.version>
+    <scm-api.version>2.3.2</scm-api.version>
```
stephenc: 2.4.0 probably now

jglick: As you prefer.

pom.xml
```diff
@@ -69,8 +69,8 @@
     <changelist>-SNAPSHOT</changelist>
     <jenkins.version>2.107.3</jenkins.version>
     <java.level>8</java.level>
-    <scm-api.version>2.3.2</scm-api.version>
-    <git-plugin.version>3.3.0</git-plugin.version>
+    <scm-api.version>2.4.0</scm-api.version>
+    <git-plugin.version>4.0.0-rc</git-plugin.version>
```
rsandell: Not sure that I'm very comfortable with depending on an "rc" version of a plugin that is also removed from the update center?

jglick: This is for `test` scope.

PR #139, opened 2018-12-27
Modifying the 'NoTriggerOrganizationFolderProperty' deprecation message

deweya:
The **"(Deprecated) Automatic branch project triggering"** option on plugins such as bitbucket/github-branch-source display a confusing deprecation message. The message says that the old option has been replaced with a "Named branch" build strategy, but it's unclear how to access the new option. 

We found that this option is exposed by installing the "Basic Branch Build Strategies" plugin. This PR modifies the message to inform users of the build strategies plugin.

### DIFF
--- a/src/main/resources/jenkins/branch/Messages.properties
+++ b/src/main/resources/jenkins/branch/Messages.properties
@@ -52,4 +52,4 @@ OrganizationFolder.OrJoinN.First={0}, {1}
 OrganizationFolder.OrJoinN.Middle={0}, {1}
 OrganizationFolder.OrJoinN.Last={0} or {1}
 TaskNounUiTextProvider.TaskNoun=Scan
-NoTriggerOrganizationFolderProperty.PropertyMigration=The "Automatic branch project triggering" option has been replaced with a "Named branch" build strategy
+NoTriggerOrganizationFolderProperty.PropertyMigration=The "Automatic branch project triggering" option has been replaced with a "Named branch" build strategy. Install the "Basic Branch Build Strategies" plugin for this new option.


kevindelord:
Could you please explain what should be done then?

The current (deprecated) feature is perfectly working for us at the moment. Therefore I wonder why this is getting deprecated... I could not find any commit or info about this deprecation.

deweya:
I think the maintainers of this plugin can explain that and the decision to deprecate the feature better than I can.

I wrote this PR just because my team and I couldn't figure out what the new 'non-deprecated' option was based on what was displayed on the UI alone. It wasn't until we dug into the [source code](https://github.com/jenkinsci/branch-api-plugin/blob/0ae692141b3ba878b9fd6c774c063fefdd7ff7c2/src/main/java/jenkins/branch/NoTriggerOrganizationFolderProperty.java#L49) that we found we actually needed to install an entirely new plugin, [Basic Branch Build Strategies](https://github.com/jenkinsci/basic-branch-build-strategies-plugin).

kevindelord:
You are all too correct! Your PR led me into digging in the code as well.
I've asked the maintainers for more details and they quickly answered on this thread:
https://github.com/jenkinsci/branch-api-plugin/pull/122#issuecomment-465206374

stephenc:
The deprecated warning is intended to be present *until* you install the plugin. The install the plugin message is an admin alert (and only the admin can install plugins). Once the plugin is installed, the migration code *should* replace your deprecated configuration with the migrated configuration. The message is intended to advise the option being deprecated.

Additionally, somebody else could provide an alternative migration plugin so we cannot assume that [Basic Branch Build Strategies](https://github.com/jenkinsci/basic-branch-build-strategies-plugin) is  the one that people will actually want to install... for example because Basic Branch Build Strategies cannot tell whether `(master|\QPR-\E\d+)` means:

* build the master branch and all branches named after the JIRA project `PR`; or
* build the master branch and all pull requests

Never mind the fun of parsing a regex that would be involved in even trying to determine what to do... 

Can you imagine somebody with a regex like `^((ma((int(enance)?-\d+)|(ster)))|([pP][rR]-\d+))$` which is equivalent to `^(master|maint-\d+|maintenance-\d+|pr-\d+|pR-\d+|Pr-d+|PR-\d+)$`

It is near impossible to infer the intent of a regex as regards what should be matched... but if somebody else wants to create a plugin with an alternative strategy that tries to infer PRs (say by randomly probling the regex with `PR-1`, `PR-45`, `PR-12456` and 10 random ones in between) and then adding the change requests as part of the migration... I'm not going to stop them, but it will not be part of [Basic Branch Build Strategies](https://github.com/jenkinsci/basic-branch-build-strategies-plugin)

stephenc:
Now if you are saying that the deprecated functionality is not working while the warning is displayed... well that's a bug. The intent was that the deprecated functionality should work until the migration code has been able to run (which requires a migration plugin to be installed)

kevindelord:
Sorry if I have got you confused but the deprecated functionality is working fine.

We simply decided to fix this "deprecation" warning/message and ended up here.
The new plugin is now installed, the concept understood and it's all working now!
Thanks :]

PR #142, opened 2019-03-12
Handle IOException in onOnline when failing to delete workspace, to prevent a node disconnection

giacomolozito:
Hello,
currently a failed workspace deletion attempt within onOnline will result in node disconnection, because the exception is not handled.

Unfortunately workspace deletion issues are getting more common as users run jobs inside docker containers that might run with a different user than jenkins, thus resulting in files left around in the workspace that jenkins can't delete. While this is a user problem and should be dealt with by ensuring jobs clean up their own files, what would normally be just an annoyance (the workspace files left around) can turn into an outage as the node fails to re-connect if it can't delete a workspace.

So this patch mitigates the issue by preventing onOnline from failing and by warning the operator that there are workspaces requiring manual cleanup (or jobs requiring a review).

Please note that https://issues.jenkins-ci.org/browse/JENKINS-55597 might be related to this or it might be another plugin - in my case it was branch-api not handling the exception so I patched it.

Kind regards

### DIFF
--- a/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
+++ b/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
@@ -531,7 +531,12 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
                                 String childName = child.getName();
                                 if (childName.equals(path) || childName.startsWith(path + COMBINATOR)) {
                                     listener.getLogger().println("deleting obsolete workspace " + child);
-                                    child.deleteRecursive();
+                                    try {
+                                        child.deleteRecursive();
+                                    } catch (IOException x) {
+                                        LOGGER.log(Level.WARNING, "could not delete workspace " + child, x);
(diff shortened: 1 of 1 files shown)


bitwiseman:
Verified - this is a broader issue.   Opened https://issues.jenkins-ci.org/browse/JENKINS-57111.  

That said, this looks like a reasonable way to address this - giving meaningful/helpful output when this happens. 

Could you add a test for this to [WorkspaceLocatorImplTest.java](https://github.com/jenkinsci/branch-api-plugin/blob/master/src/test/java/jenkins/branch/WorkspaceLocatorImplTest.java)?  

Then we can merge.
Thanks!

bitwiseman:
@giacomolozito 
Any chance you can add a test for this?

bitwiseman:
I'm going to go ahead and merge this for inclusion in the next release.  It is a safe change.

bitwiseman:
@giacomolozito 
FYI, the underlying fix for this was released in v2.177.

PR #144, opened 2019-03-25
[JENKINS-38552] Set the revision even if the build does not happen

fbelzunc:
Giving some context about this change...

I am trying to implement [JENKINS-38552](https://issues.jenkins-ci.org/browse/JENKINS-38552). To avoid the first build to happen when the first scan workflow multibranch happens is quite important in an enterprise context, ie in a big organization, it might trigger hundred of agents.

The best way to achieve this goal seems to be to create a new `BranchBuildStrategy` in the `basic-branch-build-strategies-plugin` using the ability to read the current/previous revision. The idea would be to trigger the build only when there is a previous revision.

Said all of this, from what I can see in the code, there is not any reason why `factory.setRevisionHash(item, revision)` must be dependent of the build to happen. So I am moving this outside scheduleBuild function.

See also https://github.com/jenkinsci/basic-branch-build-strategies-plugin/pull/4


rsandell:
This is an API change of sorts so pinging @stephenc just in case

stephenc:
I think the changes are not consistent with respect to "happens before" of the setting last revision vs checking if buildable. That needs to be fixed in case the strategy looks outside of its parameters.

No objections to having the previous revision be updated as it is necessary to enable some build strategy implementations

fbelzunc:
@stephenc Sorry, I am not an expert on this API - would you mind to be a bit more clear, please?

I am not sure if you mean that I should not set the revision in case it is not buildable.

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -2054,14 +2093,17 @@ public void observe(@NonNull SCMHead head, @NonNull SCMRevision revision) throws
                                 revision
                         );
                         needSave = true;
+                        try {
+                            factory.setRevisionHash(project, revision);
```
stephenc: is this before or after the call to `isAutomaticBuild`

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -2072,15 +2114,18 @@ public void observe(@NonNull SCMHead head, @NonNull SCMRevision revision) throws
                         }
                     } else if (revision.isDeterministic()) {
                         SCMRevision prevRevision = _factory.getRevision(project);
+                        try {
+                            factory.setRevisionHash(project, revision);
```
stephenc: is this before or after the call to `isAutomaticBuild`

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -2110,11 +2155,14 @@ public void observe(@NonNull SCMHead head, @NonNull SCMRevision revision) throws
                                 needSave = true;
                                 // get the previous revision
                                 SCMRevision prevRevision = _factory.getRevision(project);
+                                try {
+                                    factory.setRevisionHash(project, revision);
```
stephenc: is this before or after the call to `isAutomaticBuild`

stephenc:
On every code path where you set the revision, you need to be consistent as to whether you persist the revision before or after the call to `isAutomaticBuild`.

From one point of view it doesn't matter whether all before or all after as long as they are all the same side.

From an API perspective, the `BranchBuildStrategy` has access to the `SCMSource` and hence the `SCMSourceOwner` so it could be doing *crazy* things that might end up looking at the persisted last revision... so from that point of view I suspect we want the `factory.setRevisionHash(project, revision);
` calls *always after* the  call to `isAutomaticBuild`... but if you can see a compelling reason for the alternative sequence I am open to persuasion. In any case they all need to be consistent

fbelzunc:
@stephenc Now, all of them are before `isAutomaticBuild ` - the reason why I am looking at this is that the right way to avoid the first scan to trigger a build, so in the `BuildBranchStrategy` we can do something like the thing below.

```
    @Override
    public boolean isAutomaticBuild(@NonNull SCMSource scmSource, @NonNull SCMHead scmHead, @NonNull SCMRevision currRevision, SCMRevision prevRevision) {
        if (prevRevision != null) {
            return true;
        }
        return false;
    }
```

fbelzunc:
It seems there is an issue on the OSS agents 

```
java.nio.file.FileSystemException: /home/jenkins/workspace/Plugins/branch-api-plugin: No space left on device
    at sun.nio.fs.UnixException.translateToIOException(UnixException.java:100)
    at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
    at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:116)
    at sun.nio.fs.UnixFileSystemProvider.createDirectory(UnixFileSystemProvider.java:385)
    at java.nio.file.Files.createDirectory(Files.java:682)
    at java.nio.file.Files.createAndCheckIsDirectory(Files.java:789)
    at java.nio.file.Files.createDirectories(Files.java:775)
    at hudson.FilePath.mkdirs(FilePath.java:3228)
    at hudson.FilePath.access$1300(FilePath.java:209)
    at hudson.FilePath$Mkdirs.invoke(FilePath.java:1250)
    at hudson.FilePath$Mkdirs.invoke(FilePath.java:1246)
    at hudson.FilePath$FileCallableWrapper.call(FilePath.java:3041)
    at hudson.remoting.UserRequest.perform(UserRequest.java:212)
    at hudson.remoting.UserRequest.perform(UserRequest.java:54)
    at hudson.remoting.Request$2.run(Request.java:369)
    at hudson.remoting.InterceptingExecutorService$1.call(InterceptingExecutorService.java:72)
    at java.util.concurrent.FutureTask.run(FutureTask.java:264)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1135)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
    at java.lang.Thread.run(Thread.java:844)
```

stephenc:
If we set the last revision *before* calling `isAutomaticBuild` then _in the event that a build strategy goes peeking_ it would see the revision as set and might then decide this wasn't the first build. Can you explain why you think it should be before and not after?

What about the case where the BranchBuildStrategy plugin is broken? if you set the revision before then the revision will be set and then when the user upgrades to a working branch build strategy the build will be seen as already processed?

I think you want after not before

fbelzunc:
@stephenc I see your point know - I don't really care about before/after as long as it does not happen when there is a build - as it was before.

fbelzunc:
@stephenc Would you mind to review again when you are out of being busy, please?

Alex-Vol-SV:
This change broke a very useful behavior that I came to depend on for my CI builds. 

Using a "Skip build if Pull Request Title contains pattern" strategy I am able to have developers dictate that their PR is not yet ready to automatically build because they are using it as a mechanism to review/discuss design implementation details. With this custom plugin strategy we relied on the prior behavior to not set the last revision if a build did not get scheduled. When the developers completed the implementation after the early review discussion they updated their PR title to automatically enable building the PR.

What would it take to make this optional/configurable?

stephenc:
IMHO breaking a user’s custom build strategy is a critical regression. If a way cannot be found to update the user’s build strategy to restore the original behaviour then this change needs to be backed out and a new plan of attack developed

fbelzunc:
Hello, I am not sure if I agree with @stephenc , but I am in any case happy to do whatever you think is the best since you are the maintainer.

The reason why I disagree are:

1. The main reason is that from a conceptual point of view, there is a new revision each time there is a change. It should not matter if there is, or not, a build on Jenkins
2. If I understand correctly, this is a custom plugin strategy, and when you prefer to use a custom approach instead of contributing to the OSS, then these are the kind of problems you face. You will need to cope with this. It is not possible to satisfy all custom things...

Said this, we can revert this change or maybe think about a way to accomplish the @Alex-Vol-SV 's goal. If @Alex-Vol-SV shares with me code, then I might be able to think into an alternative to knew when a new build was performed.

stephenc:
As it stands, how do I implement a strategy that build when a GitHub label is applied? If you can show how to write such a strategy then this change can stay... if not then you’ll need to revert and try again

stephenc:
@fbelzunc I think you'll can solve by building on top rather than reverting:

* introduce storage of a new revision so that we know know both the last seen revision and the last built revision.
* Modify the BranchBuildStrategy to move the SPI method to include this revision in the parameters. May need legacy implementation fallback to use the "last built" revision for strategies using the old SPI method.
* Consult the strategies if either of these revisions are not the same as the current revision.
* Always update the last seen revision after consulting the strategies. Only update the last build revision if the strategies signalled a build should be scheduled.

WDYT? @rsandell

fbelzunc:
@stephenc Yes, I was thinking as well in having the last seen revision and the last built revision.

I am willing to implement this, but give me a bit of time since I am not an expert in this field. For the moment, @Alex-Vol-SV could just rely on the previous versions of the branch-api.

Alex-Vol-SV:
+1 on the distinction between lastseen/lastbuilt

Thanks for considering the alternatives, yes, I can work around this. Probably will do a custom build of branch-api-plugin with this change adjusted to satisfy my immediate needs. 

I am upgrading Jenkins to latest versions of some plugins including the most recent composite basic strategy changes which solved a problem I was working around before using an invasive and not very well thought through change in branch-api. 

By the way, the composite strategy implementation in basic-strategies plugin is not binary compatible with older strategy implementations as the branch-api implementation is attempting to be. The commit skip strategy implementations cannot be used inside All/Any/None compositions. There may be other strategy plugins that are also affected by the new BranchStrategy interface changes in this manner.

I plan to release the custom strategy as part of the [Commit Skip Plugin](https://github.com/jenkinsci/scm-trait-commit-skip-plugin/) which I modified to implement this behavior. This addition to the commit skip behavior is unique in that it relies on the commit last built remaining unchanged unless there is a build.

I only implemented this custom strategy for Bitbucket server, I did not have time to research Github scmsource for a similar feature. It provides generally useful behavior and would like to not perpetually maintain a custom plugin.

Thank you for this latest batch of enhancements, it made my life easier except for this one little thing.

PR #145, opened 2019-03-27
Support for programmatically generating OrganizationFolders

awittha:
In my own life, I am building a Jenkins plugin that adds the ability to generate `OrganizationFolder`s for a user against an SCM server (each of which will then generate `MultiBranchProjects` for each repo, as an `OrganizationFolder` does).

Suppress modify permissions in OrganizationFolder when it is a direct child of a ComputedFolder
=====

The [`MultiBranchProject` detects if it is a child of an `OrganizationFolder`](https://github.com/jenkinsci/branch-api-plugin/blob/master/src/main/java/jenkins/branch/MultiBranchProject.java#L776-L797) and hides from users the ability to directly update and delete it, because it is managed programmatically by the `ComputedFolder` aspect of `OrganizationFolder`.

`OrganizationFolder` itself does _not_ do this, so the automatically-generated organization folders can be modified and deleted by users after their creation. This isn't ideal.

I have taken the (above-linked) code block from `MultiBranchProject` that hides the modification permissions when it should, and added it to `OrganizationFolder`

I have made one change: Instead of checking for a _specific implementation_ parent class, it will hide those permissions any time it is a child of a `ComputedFolder` - so anyone that computes & generates `OrganizationFolder`s will get the correct behavior regardless of their specific implementation class' type.

I would have preferred to keep this entirely in my own plugin code by extending `OrganizationFolder` but it is a `final` class and so I must make the code change here. 

**I believe this is a low-risk contribution with value to all plugin authors, not just myself**.

I don't know of anyone (or any plugin) that currently puts `OrganizationFolder`s in `ComputedFolder`s _and_ expects human users to be able to modify & delete them through the WebUI. If any such users _do_ exist, though, this change would break their workflow.

Do not use default MultiBranchProjectFactory(ies) onCreatedFromScratch if some exist
=====

Currently, when an OrganizationFolder is created the `onCreatedFromScratch()` method will add all enabled-by-default MultiBranchProjectFactory types to the OrganizationFolder.

This updates that behavior to skip adding the factories if and only if there were already some factories set. This allows code that creates an OrganizationFolder to also specify which project factories the folder should use, and ensures that the OrganizationFolder won't go and add additional factories on its own.

**I believe this is a low-risk contribution.**

Only other workflows that were programmatically creating OrganizationFolders _and_ relying on the default MultiBranchProjectFactory classes being added to the folders _in addition to at least one non-default factory that the workflow added_ would experience a change in behavior.

### DIFF
--- a/src/main/java/jenkins/branch/OrganizationFolder.java
+++ b/src/main/java/jenkins/branch/OrganizationFolder.java
@@ -24,43 +24,7 @@
 
 package jenkins.branch;
 
-import static jenkins.scm.api.SCMEvent.Type.CREATED;
-import static jenkins.scm.api.SCMEvent.Type.UPDATED;
-
-import java.io.File;
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Set;
-import java.util.TreeSet;
-import java.util.concurrent.TimeUnit;
-import java.util.logging.Level;
-import java.util.logging.Logger;
-
-import javax.annotation.CheckForNull;
-import javax.annotation.Nonnull;
-import javax.servlet.ServletException;
-
-import org.acegisecurity.AccessDeniedException;
-import org.acegisecurity.Authentication;
-import org.apache.commons.io.Charsets;
-import org.apache.commons.io.FileUtils;
-import org.apache.commons.lang.StringUtils;
-import org.jenkins.ui.icon.Icon;
-import org.jenkins.ui.icon.IconSet;
-import org.jenkins.ui.icon.IconSpec;
-import org.jvnet.localizer.LocaleProvider;
-import org.kohsuke.accmod.Restricted;
-import org.kohsuke.accmod.restrictions.NoExternalUse;
-import org.kohsuke.stapler.StaplerRequest;
-import org.kohsuke.stapler.StaplerResponse;
-
+import antlr.ANTLRException;
 import com.cloudbees.hudson.plugins.folder.AbstractFolderDescriptor;
 import com.cloudbees.hudson.plugins.folder.AbstractFolderProperty;
 import com.cloudbees.hudson.plugins.folder.ChildNameGenerator;
@@ -74,8 +38,6 @@ import com.cloudbees.hudson.plugins.folder.computed.PeriodicFolderTrigger;
 import com.cloudbees.hudson.plugins.folder.views.AbstractFolderViewHolder;
 import com.google.common.collect.ImmutableSet;
 import com.thoughtworks.xstream.XStreamException;
-
-import antlr.ANTLRException;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.BulkChange;
 import hudson.Extension;
@@ -100,6 +62,24 @@ import hudson.security.ACL;
 import hudson.security.Permission;
 import hudson.util.DescribableList;
 import hudson.util.StreamTaskListener;
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+import javax.servlet.ServletException;
 import jenkins.model.Jenkins;
 import jenkins.model.TransientActionFactory;
 import jenkins.scm.api.SCMEvent;
@@ -118,6 +98,22 @@ import jenkins.scm.api.SCMSourceOwner;
 import jenkins.scm.api.metadata.ObjectMetadataAction;
 import jenkins.scm.impl.SingleSCMNavigator;
 import jenkins.scm.impl.UncategorizedSCMSourceCategory;
+import org.acegisecurity.AccessDeniedException;
+import org.acegisecurity.Authentication;
+import org.apache.commons.io.Charsets;
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang.StringUtils;
+import org.jenkins.ui.icon.Icon;
+import org.jenkins.ui.icon.IconSet;
+import org.jenkins.ui.icon.IconSpec;
+import org.jvnet.localizer.LocaleProvider;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+import org.kohsuke.stapler.StaplerRequest;
+import org.kohsuke.stapler.StaplerResponse;
+
+import static jenkins.scm.api.SCMEvent.Type.CREATED;
+import static jenkins.scm.api.SCMEvent.Type.UPDATED;
 
 /**
  * A folder-like collection of {@link MultiBranchProject}s, one per repository.
--- a/src/main/java/jenkins/branch/OrganizationFolder.java
+++ b/src/main/java/jenkins/branch/OrganizationFolder.java
@@ -186,12 +186,16 @@ public final class OrganizationFolder extends ComputedFolder<MultiBranchProject<
     @Override
     public void onCreatedFromScratch() {
         super.onCreatedFromScratch();
-        for (MultiBranchProjectFactoryDescriptor d : ExtensionList.lookup(MultiBranchProjectFactoryDescriptor.class)) {
-            MultiBranchProjectFactory f = d.newInstance();
-            if (f != null) {
-                projectFactories.add(f);
+
+        if( projectFactories.isEmpty() ) {
+            for (MultiBranchProjectFactoryDescriptor d : ExtensionList.lookup(MultiBranchProjectFactoryDescriptor.class)) {
+                MultiBranchProjectFactory f = d.newInstance();
+                if (f != null) {
+                    projectFactories.add(f);
+                }
             }
         }
+
         try {
             addTrigger(new PeriodicFolderTrigger("1d"));
         } catch (ANTLRException x) {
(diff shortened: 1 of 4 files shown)


src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -56,26 +95,10 @@
 import hudson.model.TopLevelItem;
 import hudson.model.View;
 import hudson.model.listeners.SaveableListener;
+import hudson.security.ACL;
+import hudson.security.Permission;
```
awittha: My IDE alphabetized imports.

`hudson.security.ACL` & `hudson.security.Permission` are additions.

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -35,7 +71,10 @@
 import com.cloudbees.hudson.plugins.folder.computed.FolderComputation;
 import com.cloudbees.hudson.plugins.folder.computed.PeriodicFolderTrigger;
 import com.cloudbees.hudson.plugins.folder.views.AbstractFolderViewHolder;
+import com.google.common.collect.ImmutableSet;
```
awittha: My IDE alphabetized imports.

`com.google.common.collect.ImmutableSet` is an addition.

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -24,7 +24,43 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+import javax.servlet.ServletException;
+
+import org.acegisecurity.AccessDeniedException;
+import org.acegisecurity.Authentication;
```
awittha: My IDE alphabetized imports.

`org.acegisecurity.Authentication` is an addition.

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -695,7 +730,7 @@ public String getDescription() {
             );
         }
 
-        //@Override TODO once baseline is 2.x
+        @Override
```
awittha: This appears to have happened, so I did the TODO.

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -676,7 +711,7 @@ public String getCategoryId() {
          *
          * @return A string with the description. {@code TopLevelItemDescriptor#getDescription()}.
          */
-        //@Override TODO once baseline is 2.x
+        @Override
```
awittha: This appears to have happened, so I did the TODO.

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -665,7 +700,7 @@ public TopLevelItem newInstance(ItemGroup parent, String name) {
          *
          * @return A string with the category identifier. {@code TopLevelItemDescriptor#getCategoryId()}
          */
-        //@Override TODO once baseline is 2.x
+        @Override
```
awittha: This appears to have happened, so I did the TODO.

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -177,12 +185,16 @@ public OrganizationFolder(ItemGroup parent, String name) {
     @Override
     public void onCreatedFromScratch() {
         super.onCreatedFromScratch();
-        for (MultiBranchProjectFactoryDescriptor d : ExtensionList.lookup(MultiBranchProjectFactoryDescriptor.class)) {
-            MultiBranchProjectFactory f = d.newInstance();
-            if (f != null) {
-                projectFactories.add(f);
+
+        if( projectFactories.isEmpty() ) {
+            for (MultiBranchProjectFactoryDescriptor d : ExtensionList.lookup(MultiBranchProjectFactoryDescriptor.class)) {
```
awittha: Changed Behavior:

Only add default project factories if there aren't any. This allows code that creates an OrganizationFolder to have confidence that if it sets projectFactories on the folder, the folder will use only and exactly those factories.

Without this change, default factories are always _added_ to whichever factories were set by code that created the folder. This can lead to undesired jobs being created.

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -628,6 +640,33 @@ public String getDisplayName() {
         return super.getDisplayName();
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public ACL getACL() {
+        final ACL acl = super.getACL();
+        if (getParent() instanceof ComputedFolder<?>) {
```
awittha: Changed Behavior:

If the OrganizationFolder is a direct child of a ComputedFolder, do not allow direct configuration or deletion of the Folder, nor creation, modification, or deletion of Views.

bitwiseman:
@awittha 
Looks reasonable to me.  Can you add some tests?

awittha:
Added unit tests & rebased off of latest `master`.

src/test/java/jenkins/branch/OrganizationFolderTest.java
```diff
@@ -140,6 +149,171 @@ public void deletedMarker() throws Exception {
...
+            protected void computeChildren(
+                ChildObserver<OrganizationFolder> observer,
+                TaskListener listener)
+                    throws IOException, InterruptedException
+            {
+                /*
+                 * doesn't look like we can actually compute children during a unit test.
+                 * Normally we'd want to create the OrganizationFolder here
+                 * to match the "real" Jenkins workflow as closely as possible
+                 */
```
bitwiseman: @awittha 
Why can't we compute children during a unit test?

awittha: In my own code, when I want to kick off the computation of children of a ComputedFolder in a "safe" way, I used `ComputedFolder#scheduleBuild(...)`.

This didn't actually result in any computations happening, during the unit tests - the actual `computeChildren(...)` method was never called.

I didn't attach a debugger, but my _guess_  is that when it tries to actually schedule a build ( https://github.com/jenkinsci/cloudbees-folder-plugin/blob/master/src/main/java/com/cloudbees/hudson/plugins/folder/computed/ComputedFolder.java#L580 ), there isn't actually a build queue in the mocked-up Jenkins?

In any case, it isn't terribly important to _actually_ compute children as this code doesn't activate _when an `OrganizationFolder` is created as part of a `ComputedFolder` computation_, but merely when an `OrganizationFolder`'s _parent is a `ComputedFolder`_ - which we can set up without actually having to have a working `ComputedFolder` during the test.

Should I update the comment to be more specific to that effect - that it doesn't work out of the box and we don't really need it anyway - do you think?

bitwiseman: If you want to add a little more to say why it is not a big deal, that'd be great. 

Great work on this.

awittha: Updated the comment; new text:

```
* We don't actually need to compute children during a unit test.
* We just need an OrganizationFolder to have this ComputedFolder
* as its parent.
*
* Since item parents are set on the item when constructing it,
* we can do that when we create the OrganizationFolder.
*
* Also, #scheduleBuild(...) to enqueue a computeChildren
* doesn't work out-of-the-box in the current unit-test setup.
```

bitwiseman:
@reviewbybees Can I get another set of eyes on this it looks good to me, but I'd like a second upvote.

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -24,7 +24,43 @@
 
 package jenkins.branch;
 
-import antlr.ANTLRException;
+import static jenkins.scm.api.SCMEvent.Type.CREATED;
+import static jenkins.scm.api.SCMEvent.Type.UPDATED;
```
stephenc: static after non static

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -24,7 +24,43 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
```
stephenc: all alphabetical a-z no grouping

src/main/java/jenkins/branch/OrganizationFolder.java
stephenc: Sam logic should also be applied to MultiBranchProject itself, i.e. if it is a direct child of a ComputedFolder, rather than hardcoded to OrganizationFolder https://github.com/jenkinsci/branch-api-plugin/blob/ed229d8e1392645c99e5686b8bd652a14ec87f4a/src/main/java/jenkins/branch/MultiBranchProject.java#L772

src/test/java/jenkins/branch/OrganizationFolderTest.java
```diff
@@ -24,21 +24,42 @@
...
-import hudson.model.Result;
-import hudson.model.TaskListener;
-import hudson.model.View;
-import jenkins.branch.harness.MultiBranchImpl;
-import jenkins.scm.api.SCMNavigator;
-import jenkins.scm.impl.SingleSCMNavigator;
-import hudson.scm.NullSCM;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.nullValue;
```
stephenc: Imports order is: static as a group after all non-static. All imports strict a-z

bitwiseman:
@stephenc  
Re 2 above: could that be a separate change?  It seems like that wasn't touched so far in this PR.

stephenc:
@bitwiseman well given that it mirrors the exact change made in this PR, from an API perspective I would rather the two go hand in hand (it's just one line to change anyway)

src/main/java/jenkins/branch/OrganizationFolder.java
stephenc: Avoid Guava like the plague. This can be replaced with JDK classes

src/test/java/jenkins/branch/OrganizationFolderTest.java
```diff
@@ -24,21 +24,26 @@
 
 package jenkins.branch;
 
+import com.cloudbees.hudson.plugins.folder.computed.ChildObserver;
+import com.cloudbees.hudson.plugins.folder.computed.ComputedFolder;
+import com.google.common.collect.ImmutableSet;
```
stephenc: Just say no to Guava

bitwiseman:
@stephenc 
You're three years too late. 
https://github.com/jenkinsci/branch-api-plugin/commit/667d51a05d6f373ce5858cf98f5a7e123cf123b7

But okay, removed.

stephenc:
Better late than never

awittha:
Thanks, y'all!

PR #146, opened 2019-04-05
[JENKINS-56903] Adds a health metric that reports only the primary branch

stephenc:
https://issues.jenkins-ci.org/browse/JENKINS-56903


jglick:
also #147

PR #147, opened 2019-04-05
[JENKINS-56917] Implementation of extension point to enable org folders to decorate their children

stephenc:
https://issues.jenkins-ci.org/browse/JENKINS-56917

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
   <properties>
     <revision>2.3.1</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.107.3</jenkins.version>
+    <jenkins.version>2.130</jenkins.version>
     <java.level>8</java.level>
     <scm-api.version>2.4.1</scm-api.version>
     <git-plugin.version>4.0.0-rc</git-plugin.version>
@@ -97,7 +97,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>cloudbees-folder</artifactId>
-      <version>6.1.0</version>
+      <version>6.8</version>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
   <properties>
     <revision>2.4.0</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.130</jenkins.version>
+    <jenkins.version>2.138</jenkins.version>
     <java.level>8</java.level>
     <scm-api.version>2.4.1</scm-api.version>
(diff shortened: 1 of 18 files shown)


pom.xml
```diff
@@ -67,7 +67,7 @@
   <properties>
     <revision>2.3.1</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.107.3</jenkins.version>
+    <jenkins.version>2.130</jenkins.version>
```
stephenc: Need this for cloudbees-folders

pom.xml
```diff
@@ -97,7 +97,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>cloudbees-folder</artifactId>
-      <version>6.1.0</version>
+      <version>6.8</version>
```
stephenc: Need this for `ComputedFolder.removeTrigger`

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -1430,6 +1437,11 @@ public void complete() throws IllegalStateException, IOException, InterruptedExc
                                 } catch (ANTLRException x) {
```
stephenc: Should we keep `project.addTrigger(new PeriodicFolderTrigger("1d"));` or kill it off?

stephenc:
Screenshot 2019-04-05 at 16 09 44

src/main/java/jenkins/branch/OrganizationFolder.java
rsandell: Eeeeeh. I'm always a bit squemish about removing behaviour.
That said I didn't know this was a thing :)

stephenc: Yep, there are load of people hacking configuration of this because they don't like it as 1 day... with this change we fix that... but that changes the behaviour for people...

another option would be to check for the property and only set if the property is missing

stephenc:
Latest screenshot

Screenshot 2019-04-07 at 21 25 37

stephenc:
@rsandell this should be fully baked now

rsandell:
@stephenc I think you are best suited to author the changelog for this given that users who has messed with the xml in the past needs to be aware of.

pom.xml
```diff
@@ -65,9 +65,9 @@
   </scm>
 
   <properties>
-    <revision>2.3.1</revision>
+    <revision>2.4.0</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.107.3</jenkins.version>
+    <jenkins.version>2.138</jenkins.version>
```
stephenc: bumped from 2.130 to next lts 2.138 per @rsandell 's request

stephenc:
The modified UX for the child orphaned item strategy
drop-down-v2

PR #148, opened 2019-04-10
Java 11 readiness: build both on JDK8 and JDK11

batmat:
Making sure this plugin is constantly checked both building with JDK8 and JDK11.

This goes with the Jenkins Java 11 General Availability announcement made back in March, and we're making a pass to check plugins are looking good on a JDK11.

([to understand the Jenkinsfile content](https://wiki.jenkins.io/display/JENKINS/Java+11+Developer+Guidelines#Java11DeveloperGuidelines-MakesureyourpluginistestedinContinuousIntegrationonJava8andJava11atthesametime))

@jenkinsci/java11-support

### DIFF
--- a/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
+++ b/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
@@ -211,12 +211,12 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
             return null;
         }
     }
-    
+
     private static Map<String, String> load(FilePath workspace) throws IOException, InterruptedException {
         Map<String, String> map = new TreeMap<>();
         FilePath index = workspace.child(INDEX_FILE_NAME);
         if (index.exists()) {
-            try (InputStream is = index.read(); Reader r = new InputStreamReader(is, StandardCharsets.UTF_8); BufferedReader br = new BufferedReader(r)) {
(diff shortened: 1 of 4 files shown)


batmat:
Thanks a lot for the help @bitwiseman !

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -233,6 +233,14 @@ private static FilePath locate(TopLevelItem item, String fullName, Node node, bo
         return map;
     }
 
+    // Workaround spotbugs false-positive redunant null checking for try blocks in java 11
+    // https://github.com/spotbugs/spotbugs/issues/756
+    private static @NonNull InputStream readFilePath(@NonNull FilePath index) throws IOException, InterruptedException {
```
stephenc: We do not use Glick-style annotations. If the method returns `@NonNull` then the `@NonNull` shall be on the line preceding the access modifiers not between the access modifiers and the return type

batmat:
@stephenc if that now suits your requests, could you please merge it? Thanks!

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -233,6 +233,15 @@ private static Map<String, String> load(FilePath workspace) throws IOException,
         return map;
     }
 
+    // Workaround spotbugs false-positive redunant null checking for try blocks in java 11
```
jglick, line 236: #347

PR #149, opened 2019-04-11
[JENKINS-38552] Use a lastSeenRevision + lastBuilt

fbelzunc:
This is a follow-up of the conversation in #144 

To keep the compatibility with custom plugins which might rely on setRevision being set at build time, I am adding a setLastSeenRevision to also being able to track the current revision.

CC @stephenc @rsandell @Alex-Vol-SV

### DIFF
--- a/src/main/java/jenkins/branch/BranchBuildStrategy.java
+++ b/src/main/java/jenkins/branch/BranchBuildStrategy.java
@@ -217,8 +217,6 @@ public abstract class BranchBuildStrategy extends AbstractDescribableImpl<Branch
         return automaticBuild(source, head, currRevision, prevRevision, listener, null);
     }
 
-
-
     /**
      * API: Should the specified {@link SCMRevision} of the {@link SCMHead} for the specified {@link SCMSource} be
      * triggered when the {@link SCMHead} has been detected as created / modified?
--- a/src/main/java/jenkins/branch/BranchBuildStrategy.java
+++ b/src/main/java/jenkins/branch/BranchBuildStrategy.java
@@ -155,7 +155,7 @@ public abstract class BranchBuildStrategy extends AbstractDescribableImpl<Branch
      *                     {@link SCMRevision#isDeterministic()} previous revisions as polling for changes will have
      *                     confirmed that there is a change between this and {@code currRevision} even if the two
      *                     are equal.
-     * @param lastSeenRevision the {@link SCMRevision} that the head is now at
+     * @param lastSeenRevision the {@link SCMRevision} that the head was last seen
      * @param listener     the {@link TaskListener} that can be used for outputting any rational for the decision
      * @return {@code true} if and only if the {@link SCMRevision} should be automatically built when the
      * {@link SCMHead} has been detected as created / modified.
@@ -231,7 +231,7 @@ public abstract class BranchBuildStrategy extends AbstractDescribableImpl<Branch
      *                     {@link SCMRevision#isDeterministic()} previous revisions as polling for changes will have
      *                     confirmed that there is a change between this and {@code currRevision} even if the two
      *                     are equal.
-     * @param lastSeenRevision the {@link SCMRevision} that the head is now at
+     * @param lastSeenRevision the {@link SCMRevision} that the head was last seen
      * @return {@code true} if and only if the {@link SCMRevision} should be automatically built when the
      * {@link SCMHead} has been detected as created / modified.
      * @since 2.4.2
--- a/src/main/java/jenkins/branch/BranchBuildStrategy.java
+++ b/src/main/java/jenkins/branch/BranchBuildStrategy.java
@@ -35,7 +35,6 @@ import java.util.logging.Logger;
 import jenkins.scm.api.SCMHead;
 import jenkins.scm.api.SCMRevision;
 import jenkins.scm.api.SCMSource;
-import jline.internal.Nullable;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.ProtectedExternally;
 
@@ -46,6 +45,9 @@ import org.kohsuke.accmod.restrictions.ProtectedExternally;
  * Methods marked as {@code SPI:} are intended to be implemented by implementers of {@link BranchBuildStrategy}.
  * Methods marked as {@code API:} are intended to be invoked consumers of {@link BranchBuildStrategy}.
  * A consumer invoking a {@code SPI:} method may get a {@link UnsupportedOperationException}.
+ * Methods marked as {@code SPI:} are only to be invoked through the API methods in order to allow safe evolution.
+ * Changing the API may require updating any SPI implementations that are also API consumers, specifically the Any,
+ * All and None implementations in basic-branch-build-strategies
  *
  * @since 2.0.0
  */
@@ -224,7 +226,7 @@ public abstract class BranchBuildStrategy extends AbstractDescribableImpl<Branch
      * @param source       the {@link SCMSource}
      * @param head         the {@link SCMHead}
      * @param currRevision the {@link SCMRevision} that the head is now at
-     * @param lastBuiltRevision the {@link SCMRevision} that the head was last seen at or {@code null} if this is a newly
+     * @param lastBuiltRevision the {@link SCMRevision} that the build head was last seen at or {@code null} if this is a newly
      *                     discovered head. Care should be taken to consider the case of non
      *                     {@link SCMRevision#isDeterministic()} previous revisions as polling for changes will have
      *                     confirmed that there is a change between this and {@code currRevision} even if the two
@@ -232,7 +234,7 @@ public abstract class BranchBuildStrategy extends AbstractDescribableImpl<Branch
      * @param lastSeenRevision the {@link SCMRevision} that the head is now at
      * @return {@code true} if and only if the {@link SCMRevision} should be automatically built when the
      * {@link SCMHead} has been detected as created / modified.
-     * @since 2.1.3
+     * @since 2.4.2
      */
     @SuppressWarnings("deprecation")
     public final boolean automaticBuild(@NonNull SCMSource source,
@@ -243,7 +245,7 @@ public abstract class BranchBuildStrategy extends AbstractDescribableImpl<Branch
                                         @NonNull TaskListener listener) {
         if (Util.isOverridden(BranchBuildStrategy.class, getClass(), "isAutomaticBuild", SCMSource.class,
                 SCMHead.class, SCMRevision.class, SCMRevision.class, SCMRevision.class, TaskListener.class)) {
-            // modern implementation written to the 2.1.3+ spec
+            // modern implementation written to the 2.4.2+ spec
             return isAutomaticBuild(source, head, currRevision, lastBuiltRevision, lastSeenRevision, listener);
         }
         if (Util.isOverridden(BranchBuildStrategy.class, getClass(), "isAutomaticBuild", SCMSource.class,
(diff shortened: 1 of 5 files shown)


Alex-Vol-SV:
You are fast @fbelzunc 
I will check this out pulling the code locally.
I can adjust the scm-commit-skip changes to adapt to this further validating your approach.

Thank you!

fbelzunc:
Well, I will also look at this again tomorrow with fresh eyes, but I would appreciate if @stephenc acks me that I am on the right path :-)

src/main/java/jenkins/branch/BranchBuildStrategy.java
```diff
@@ -130,12 +130,42 @@ public boolean isAutomaticBuild(@NonNull SCMSource source,
...
      */
+    @Deprecated
+    @SuppressWarnings("deprecation")
+    @Restricted(ProtectedExternally.class)
+    public boolean isAutomaticBuild(@NonNull SCMSource source,
+                                             @NonNull SCMHead head,
+                                             @NonNull SCMRevision currRevision,
+                                             @CheckForNull SCMRevision prevRevision,
+                                             @NonNull TaskListener listener) {
+        throw new UnsupportedOperationException("Modern implementation accessed using legacy API method");
```
stephenc: I think you'll need to chain the call through, where `lastSeenRevision` is also the `prevRevision`

src/main/java/jenkins/branch/BranchBuildStrategy.java
```diff
@@ -130,12 +130,42 @@ public boolean isAutomaticBuild(@NonNull SCMSource source,
...
+     * @since 2.4.2
+     */
     @Restricted(ProtectedExternally.class)
     public abstract boolean isAutomaticBuild(@NonNull SCMSource source,
                                              @NonNull SCMHead head,
                                              @NonNull SCMRevision currRevision,
                                              @CheckForNull SCMRevision prevRevision,
-                                             @NonNull TaskListener listener);
+                                             @NonNull TaskListener listener,
+                                             @CheckForNull SCMRevision lastSeenRevision);
```
stephenc: I would go with `(source,head,curRevision,lastBuiltRevision,lastSeenRevision,listener)`

src/main/java/jenkins/branch/BranchBuildStrategy.java
stephenc: nevermind... I may have forgotten that these are supposed to be protected methods and are the SPI not the API

src/main/java/jenkins/branch/BranchBuildStrategy.java
```diff
@@ -184,6 +214,37 @@ public final boolean automaticBuild(@NonNull SCMSource source,
                                         @NonNull SCMRevision currRevision,
                                         @CheckForNull SCMRevision prevRevision,
                                         @NonNull TaskListener listener) {
+        return automaticBuild(source, head, currRevision, prevRevision, listener, null);
```
stephenc: last seen should be defaulted to lastBuilt

src/main/java/jenkins/branch/BranchBuildStrategy.java
```diff
@@ -184,6 +214,37 @@ public final boolean automaticBuild(@NonNull SCMSource source,
                                         @NonNull SCMRevision currRevision,
```
stephenc: we need to mark this version of the API as deprecated and `@Restricted(DoNotUse.class)` (or perhaps `NoExtenalUse`) so that people compiling against the updated dependency are forced to move to the new API

stephenc:
on the right path

src/main/java/jenkins/branch/BranchBuildStrategy.java
fbelzunc: The corresponded classes seem to be marked as `Deprecated` and this one which is the latest one is keeping `@Restricted(ProtectedExternally.class)` - is not this correct?

fbelzunc:
@stephenc ready to re-review.

src/main/java/jenkins/branch/BranchProjectFactory.java
```diff
@@ -181,6 +181,35 @@ public void setRevisionHash(P project, SCMRevision revision) throws IOException
...
 
+    /**
+     * Gets the {@link SCMRevision} that the project was last seen for.
+     *
+     * @param project the project.
+     * @return the {@link SCMRevision} of the last seen.
+     */
+    @CheckForNull
+    public SCMRevision getLasSeenRevision(P project) {
+        XmlFile file = new XmlFile(new File(project.getRootDir(), "scm-last-seen-revision-hash.xml"));
```
rsandell: How often is this being called? Once per scan per folder? once per scan per child per folder?

File operations like this makes me a bit nervous since the IO performance of Jenkins is already a bit crappy and multiplication can produce powerful numbers :)

Could it be stored in the project xml as normal or why was that approach not chosen?

rsandell:
I'll do a bit more thorough review again when stephenc  has given his :+1: 

Should we backport this to a 2.3.1 release as well as a 2.4.1? Or might that mess up people upgrading?

src/main/java/jenkins/branch/BranchProjectFactory.java
stephenc: The reason we store the last built revision in the file was so that the config history plugin doesn't see config changes that are not. This should only be called once per branch per event to the branch or per scan of the repo

fbelzunc:
@stephenc ping pong

src/main/java/jenkins/branch/BranchBuildStrategy.java
```diff
@@ -184,15 +214,47 @@ public final boolean automaticBuild(@NonNull SCMSource source,
...
+     * @param currRevision the {@link SCMRevision} that the head is now at
+     * @param lastBuiltRevision the {@link SCMRevision} that the head was last seen at or {@code null} if this is a newly
+     *                     discovered head. Care should be taken to consider the case of non
+     *                     {@link SCMRevision#isDeterministic()} previous revisions as polling for changes will have
+     *                     confirmed that there is a change between this and {@code currRevision} even if the two
+     *                     are equal.
+     * @param lastSeenRevision the {@link SCMRevision} that the head is now at
+     * @return {@code true} if and only if the {@link SCMRevision} should be automatically built when the
+     * {@link SCMHead} has been detected as created / modified.
+     * @since 2.1.3
```
stephenc:
```suggestion
     * @since 2.4.2
```

src/main/java/jenkins/branch/BranchBuildStrategy.java
```diff
@@ -184,15 +214,47 @@ public final boolean automaticBuild(@NonNull SCMSource source,
...
+    @SuppressWarnings("deprecation")
+    public final boolean automaticBuild(@NonNull SCMSource source,
+                                        @NonNull SCMHead head,
+                                        @NonNull SCMRevision currRevision,
+                                        @CheckForNull SCMRevision lastBuiltRevision,
+                                        @CheckForNull SCMRevision lastSeenRevision,
+                                        @NonNull TaskListener listener) {
+        if (Util.isOverridden(BranchBuildStrategy.class, getClass(), "isAutomaticBuild", SCMSource.class,
+                SCMHead.class, SCMRevision.class, SCMRevision.class, SCMRevision.class, TaskListener.class)) {
+            // modern implementation written to the 2.1.3+ spec
```
stephenc:
```suggestion
            // modern implementation written to the 2.4.2+ spec
```

src/main/java/jenkins/branch/BranchBuildStrategy.java
```diff
@@ -184,15 +214,47 @@ public final boolean automaticBuild(@NonNull SCMSource source,
                                         @NonNull SCMRevision currRevision,
                                         @CheckForNull SCMRevision prevRevision,
                                         @NonNull TaskListener listener) {
+        return automaticBuild(source, head, currRevision, prevRevision, prevRevision, listener);
+    }
+
+    /**
+     * API: Should the specified {@link SCMRevision} of the {@link SCMHead} for the specified {@link SCMSource} be
```
stephenc: Probably should also add a JAVADOC comment to the class warning that the SPI methods are only to be invoked through the API methods in order to allow safe evolution.

Also add a note that changing the API may require updating any SPI implementations that are also API consumers, specifically the Any, All and None implementations in basic-branch-build-strategies

stephenc: Hmmm also perhaps call it out in an update to the `implementation.adoc` https://github.com/jenkinsci/branch-api-plugin/blob/master/docs/implementation.adoc (which on initial inspection appears to be missing even a section on implementing branch build strategies.

@fbelzunc you need to document the subtleties between last built and last seen in that implementation guide. Doesn't need to be long, a sentence or two should do

fbelzunc:
@rsandell hello, ready for your review.

rsandell:
@fbelzunc I've been on holiday but will look over it tomorrow.

Should we backport this to a 2.3.1 release as well as a 2.4.1? Or might that mess up people upgrading?

src/main/java/jenkins/branch/BranchBuildStrategy.java
```diff
@@ -130,11 +132,41 @@ public boolean isAutomaticBuild(@NonNull SCMSource source,
...
+     * @param source       the {@link SCMSource}
+     * @param head         the {@link SCMHead}
+     * @param currRevision the {@link SCMRevision} that the build head is now at
+     * @param lastBuiltRevision the {@link SCMRevision} that the build head was last seen at or {@code null} if this is a newly
+     *                     discovered head. Care should be taken to consider the case of non
+     *                     {@link SCMRevision#isDeterministic()} previous revisions as polling for changes will have
+     *                     confirmed that there is a change between this and {@code currRevision} even if the two
+     *                     are equal.
+     * @param lastSeenRevision the {@link SCMRevision} that the head was last seen
+     * @param listener     the {@link TaskListener} that can be used for outputting any rational for the decision
```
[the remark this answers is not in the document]
rsandell: From the code it looks like `lastBuiltRevision` but I don't think you should have to look at the code to figure that out?

fbelzunc:
@rsandell I just addressed your last feedback.

> Should we backport this to a 2.3.1 release as well as a 2.4.1? Or might that mess up people upgrading?

I am not sure what is the best option. I will let you and Stephen decide it :-)

stephenc:
> I am not sure which is the best option. I will let you and Stephen decide it :-)

Not an API decision. 

backs-away

rsandell:
I think 2.4.0 has been soaking long enough to call it stable? So no backporting needed. As it is yet another API change this should maybe be released as 2.5.0?

stephenc:
@rsandell if questions if a change might merit a minor bump, it needs a minor bump

giannello:
@bitwiseman we've just updated our installed version to 2.5.0 and _all_our builds have been retriggered by the branch indexer. All the jobs have branch indexing log messages like the following:
```
Changes detected: PR-1 (null → d5216816a623a6e1514ecfef93613ac8f73ab132+bf3a02f820238417efd017037bee85c5d79c0632 (68e6fffd6a80cbc8baf65793c77ef1b20649c8fd))
```
Bug? Expected?

fbelzunc:
This is being discussed in #152

bitwiseman:
@giannello  Fixed in v2.5.1 - released now.

giannello:
thanks @bitwiseman - too late for us, we went through the storm. It was actually fine, Jenkins took it like a champ and it can be considered a good load test for our setup.

PR #150, opened 2019-04-17
[JENKINS-54992] Do not override pronoun on projects only on their scan

bitwiseman:
This issue shows that changing the pronoun on a project in
in order to make some actions by the project more clear is
actually problematic.  

As reasonable fix, this change makes
the indexing use the source pronoun(s) while reverting the
projects to using their original pronouns.

This PR came from discussions around
https://github.com/jenkinsci/github-branch-source-plugin/pull/195

After this change the Multibranch Pipeline looks like this: 
Screen Shot 2019-04-16 at 6 16 36 PM

A github organization folder looks like this:
Screen Shot 2019-04-16 at 6 24 25 PM

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -302,10 +302,10 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
     }
 
     /**
-     * {@inheritDoc}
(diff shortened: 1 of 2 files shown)


bitwiseman:
@abayer @stephenc  @rsandell 
So, we can merge this?  (I do not have permissions)

abayer:
@bitwiseman You do now. =)

PR #152, opened 2019-05-21
Attempt to stop build storm on plugin update

bitwiseman:
From discussion of with @MarkEWaite here's one attempt at addressing JENKINS-57588.

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -2079,6 +2079,14 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
                     } else if (revision.isDeterministic()) {
                         SCMRevision scmLastBuiltRevision = _factory.getRevision(project);
                         SCMRevision scmLastSeenRevision = _factory.getLastSeenRevision(project);
+                        if (scmLastSeenRevision == null && scmLastBuiltRevision != null) {
+                            scmLastSeenRevision = scmLastBuiltRevision;
+                            try {
+                                _factory.setLastSeenRevisionHash(project, revision);
+                            } catch (IOException e) {
+                                printStackTrace(e, listener.error("Could not update last seen revision hash"));
+                            }
+                        }
                         if (!revision.equals(scmLastSeenRevision)) {
                             listener.getLogger()
                                     .format("Changes detected: %s (%s → %s)%n", rawName, scmLastSeenRevision, revision);
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -2087,7 +2087,7 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
                                 printStackTrace(e, listener.error("Could not update last seen revision hash"));
                             }
                         }
-                        if (!revision.equals(scmLastSeenRevision)) {
+                        if (!revision.equals(scmLastBuiltRevision)) {
                             listener.getLogger()
                                     .format("Changes detected: %s (%s → %s)%n", rawName, scmLastSeenRevision, revision);
                             needSave = true;


fbelzunc:
As far as I can recall, `scmLastBuiltRevision ` is the new `prevRevision` .

* https://github.com/jenkinsci/branch-api-plugin/pull/149/files

Screenshot 2019-05-22 at 10 06 04

So the code should be

```
                        SCMRevision scmLastBuiltRevision = _factory.getRevision(project);
                        SCMRevision scmLastSeenRevision = _factory.getLastSeenRevision(project);
                        if (!revision.equals(scmLastBuiltRevision)) {
                            listener.getLogger()
                                    .format("Changes detected: %s (%s → %s)%n", rawName, scmLastBuiltRevision, revision);
                            needSave = true;
``` 

* Doesn't this fix the issue?

rsandell:
I knew that a build storm was a risk, but I was very certain that we had tests for it already. But apparently I misremembered and those tests are probably in the ghbs plugin, that's where we had the last build storm.

rsandell:
@fbelzunc That might depend on how you want your "feature" to work.
Should we trigger a build based on what we last saw during a scan or what we last built. Both of them might work just a slight difference in philosophy.

fbelzunc:
`lastBuiltRevision` is the new `prevRevision` 


In the original code, this is what we had:

Screenshot 2019-05-22 at 10 37 24

Then, this changed to this

Screenshot 2019-05-22 at 10 38 10

As I understand from Stephen, we should keep the original behavior and to keep the original behaviour we should just do the change on these lines

```
 } else if (revision.isDeterministic()) {
                        SCMRevision scmLastBuiltRevision = _factory.getRevision(project);
                        SCMRevision scmLastSeenRevision = _factory.getLastSeenRevision(project);
                        if (!revision.equals(scmLastBuiltRevision)) {
                            listener.getLogger()
                                    .format("Changes detected: %s (%s → %s)%n", rawName, scmLastBuiltRevision, revision);
                            needSave = true;
```

* Does it make sense? Maybe I am saying something wrong.

With the current PR, we are certainly solving the Jira attached, but then I think we are not fixing the bug I think I introduced - sorry.

rsandell:
@fbelzunc sounds reasonable.

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -2079,6 +2079,14 @@ public void observe(@NonNull SCMHead head, @NonNull SCMRevision revision) throws
...
                         SCMRevision scmLastSeenRevision = _factory.getLastSeenRevision(project);
+                        if (scmLastSeenRevision == null && scmLastBuiltRevision != null) {
+                            scmLastSeenRevision = scmLastBuiltRevision;
+                            try {
+                                _factory.setLastSeenRevisionHash(project, revision);
+                            } catch (IOException e) {
+                                printStackTrace(e, listener.error("Could not update last seen revision hash"));
+                            }
+                        }
                         if (!revision.equals(scmLastSeenRevision)) {
```
rsandell: The above is good upgrade protection so should probably be there in any case.
Though as @fbelzunc suggests, to keep the previous version's behaviour the if statement should be changed.
```suggestion
                        if (!revision.equals(scmLastBuiltRevision)) {
```

MarkEWaite: Agreed

MarkEWaite:
I'd love to see tests, but I expect there will be an outcry from the community if this is not fixed as soon as possible.  I queued 500 unnecessary builds when I installed branch api 2.5.0 and my cluster spent most of the night completing those builds.  I'm sure there are installations that are much larger than mine.

Maybe the tests could be written after the release of branch api 2.5.1?

rsandell:
@MarkEWaite yea I think we should do that as long as we can somehow get a quick verification that it would work even with the if statement change.

MarkEWaite:
I'm willing to run another verification test with an incremental build if that helps.

I'm not doing deep testing, just checking that no build storm results from the incremental build.

MarkEWaite:
Incremental from the most recent change from Liam (cd9f962e6356d3ed656f1131f4c6d4428e9a26c9) passed my smoke test.

rsandell:
@MarkEWaite You will need to revert your instance to a pre 2.4.0 state since the buildstorm has already passed on your instance the lastSeemRevision would have been populated once already unless you can revert that data first.

bitwiseman:
@rsandell Verified.

MarkEWaite:
Thanks for the reminder @rsandell !  It was enough on past tests if I removed all the '*seen*.xml' files in the jobs folder and its subdirectories.  I'll run my smoke test again with that.

PR #156, opened 2019-06-04
add second and minute granularity

rogerhu:
GitHub consolidated webhooks for releases for created and published, which means we get double webhooks within the same minute.  We found this plugin but really wanted more granularity.  Getting more feedback about whether it's possible to add these two options -- happy to add tests if this is possible. :)

image


rogerhu:
ping on this one =)

bitwiseman:
@rogerhu 
Is there a JIRA for this?  I'm unclear what is being attempted.

rogerhu:
Trying to provide more granularity to the plugin beyond simply hours.  See context above in the description about how we use GitHub webhooks that ends up sending two webhooks within the same minute.

bitwiseman:
> GitHub consolidated webhooks for releases for created and published,

I know a bunch about this plugin and github but not everything.  I'm not making sense of what you're saying.  Walk me through this a little more.  I'm not trying to be pedantic or ask you to write a novel.  Just put at least a couple paragraphs for context.  What is the current behavior, what would you like the behavior to look like.  

I'm pretty sure just adding two lines is not enough, but I really have no way to tell.

rogerhu:
We rely on GitHub to tag/release builds.  By creating a release, GitHub can be configured to send a webhook to Jenkins to trigger a build.  The problem is that creating/releasing builds sends two separate webhooks to Jenkins.  Your plug-in is great for avoiding double-triggering builds.  The problem is the most granular time to trigger builds is 1 hour.  

If we decide to reissue another hotfix for some reason, we'd have to manually trigger a new build because this plugin throttles by an hour.  We'd prefer to throttle by minutes and seconds.

bitwiseman:
Ah! Now I get it.  Thanks!
And I see how this could be a sufficient change. Have you tried it locally?  

I think you'll need to add the new durations to https://github.com/jenkinsci/branch-api-plugin/blob/master/src/main/resources/jenkins/branch/Messages.properties as well. 

Other than that this is looking good to me.

rogerhu:
Thanks.  Will work on updates and testing.

bitwiseman:
@rogerhu  Any update?

rogerhu:
Pushed!  Added a test to also check for the existence of the property in the Message.properties!

I did realize this plugin only throttles build jobs.  Would you open to adding an to discard jobs that exceed the window?

bitwiseman:
All suggestions/ideas are welcome, but for sure that will be a separate change. 
I'd suggest you check JIRA to see if it is something that has already been filed or discussed.  If not, open a new JIRA so the change can be tracked.

PR #158, opened 2019-06-14
[JENKINS-39682] trigger after-save hooks and recalculate

daspilker:
... after `createProjectFromXML` or `updateByXml`.

See [JENKINS-39682](https://issues.jenkins-ci.org/browse/JENKINS-39682?focusedCommentId=369197&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-369197).

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -235,6 +235,12 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
                 }
             }
         }
+        if (Items.currentlyUpdatingByXml()) {
+            fireSCMSourceAfterSave(getSCMSources());
+            if (isBuildable()) {
+                scheduleBuild();
+            }
+        }
     }
 
     /**
--- a/src/test/java/integration/UpdatingFromXmlTest.java
+++ b/src/test/java/integration/UpdatingFromXmlTest.java
@@ -1,0 +1,62 @@
+package integration;
+
+import hudson.model.TopLevelItem;
+import integration.harness.BasicMultiBranchProject;
+import jenkins.scm.impl.mock.MockSCMController;
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.io.input.ReaderInputStream;
+import org.junit.Before;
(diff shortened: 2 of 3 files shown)


bitwiseman:
@daspilker  
This could use some tests.

darxriggs:
@bitwiseman apart from the tests what do you think about the change itself?

daspilker:
@bitwiseman I added some tests. The tests will fail if you remove the fix. It turned out that the `ItemListener` is not necessary.

bitwiseman:
@daspilker Not sure why this failed can you push an updated commit to run a new build?

daspilker:
@bitwiseman there was an infrastructure problem yesterday. Now it's green.

daspilker:
I rebase to current master and CI is green again.

daspilker:
Pinging @bitwiseman and @dwnusbaum. Any news on this issue?

michelzanini:
I have added a comment to https://issues.jenkins-ci.org/browse/JENKINS-39682 and will add it here as well, for visibility.

```
This PR was merged to fix an issue but has caused a worse one in my opinion.

My seed job runs every 4 hours, it scans repositories in Bitbucket and create and/or update hundreds of jobs as a result.

After updating the Branch API plugin to 2.5.5 with this PR included, now every 4 hours, ALL my ~ 250 jobs are triggering a branch scanning at the same time after seed job completes.

This causes several API calls to Bitbucket all at the same time and then they hang, as Bitbucket has a limit on number of API calls.

That means I have to revert the plugin to 2.5.4 as its impossible to continue with this behaviour.


I understand what you are trying to fix here - to run it initially when the job gets created for the first time. But it should not happen when the job gets updated.

Also, using a scan trigger fixes the problem of initial branch scanning and it is recommend to use one anyway (as sometimes Webhooks will be missed). It has never being a problem for us.

Anyway, I think this has made things worse at the point of being a blocker.

Thanks.
```

sirstrahd:
This change also caused all sorts of trouble for us, since we have a job that updates our folder xml configs and this triggers a humongous amount of reindexings. In the end we just forked the plugin and commented this change out, which addressed our issues.

PR #160, opened 2019-06-20
[JENKINS-48837] Add BranchProperty support to OrganizationFolder

nfalco79:
We have more than 50 multibranch jobs configured with branch properties. This should save time every time we change branch properties, the feature that had stop migration to organisation folder.

immagine

Each multibranch project child of the organisation folder have branch property strategy configured (the release link) only for master branch:

immagine

Any suggestion for a test case?


bitwiseman:
@nfalco79 
Could you rebase your changes on the current head of master and re-push?  It would be nice to have a working incremental to use for testing.

nfalco79:
@bitwiseman done

bitwiseman:
@nfalco79  
https://github.com/nfalco79/branch-api-plugin/tree/feature/JENKINS-48837 says "This branch is 1 commit ahead, 6 commits behind jenkinsci:master."

nfalco79:
@bitwiseman You right, I forgot to update my remote with jenkins remote. I see that is too late for 2.5.3 release :(

bitwiseman:
@nfalco79 Don't worry, we're not on a schedule.

nfalco79:
Are there some comments or changes you would I do on this PR?

bitwiseman:
@nfalco79  
Sorry for the delay. To merge, I still need to load this up for manual testing on an instance. It would be good to have some automated tests too if you have time to add some.

nfalco79:
Sure, any suggestion for a test case? I mean do you have any suggestion from which test case I can copy or take an example for these scenario?

bitwiseman:
@nfalco79 I don't know off-hand, sorry.  And I'm swamped until at least Thursday this week.  I'll be able to take a look after that.

nfalco79:
Do not worry, also because today I got an issue. If I modify branch properties (removing or adding), these changes are persisted correctly in the organisation folder config.xml but multibranch children projects are not affected by the changes also after a jenkins restart.

nfalco79:
Ok issue is resolved now. There was two issues:

1. sources variable was set to null in createBranchSources to avoid more call to createBranchSources. But this cause NPE that stop the process of update
2. The NamedExceptionsBranchPropertyStrategy only returns the branch property defined on the first matching named exception. For this reason if I do not show all defined properties for a specific branch.

nfalco79:
@bitwiseman please any news on this? Are there something you would to change?

nfalco79:
I will try

bitwiseman:
@nfalco79 
If there's anything I can do to help, just say so.

nfalco79:
@bitwiseman I'm finally back on this issue.
I add a test case to verify that multibranch project will inherites branch properties defined in the Organisation Folder.

Please let me know if you need somethingelse from my side.

nfalco79:
@bitwiseman any news about this PR? I have add test cases as request. Please reset the "changes request" mark.

nfalco79:
@dwnusbaum @bitwiseman this PR is stuck up please give me some feedback

nfalco79:
rebased

nfalco79:
failures seems not be related to the code, rerun touching a commit

nfalco79:
> [2020-05-01T00:13:50.128Z] [WARNING] Rule 5: org.apache.maven.plugins.enforcer.RequireUpperBoundDeps failed with message:
[2020-05-01T00:13:50.128Z] Failed while enforcing RequireUpperBoundDeps. The error(s) are [
[2020-05-01T00:13:50.128Z] Require upper bound dependencies error for org.slf4j:jcl-over-slf4j:1.7.25 paths to dependency are:
[2020-05-01T00:13:50.128Z] +-org.jenkins-ci.plugins:branch-api:2.5.7-rc852.82aba7c3c8bc
[2020-05-01T00:13:50.128Z]   +-org.slf4j:jcl-over-slf4j:1.7.25
[2020-05-01T00:13:50.128Z] and
[2020-05-01T00:13:50.128Z] +-org.jenkins-ci.plugins:branch-api:2.5.7-rc852.82aba7c3c8bc
[2020-05-01T00:13:50.128Z]   +-org.jenkins-ci.main:jenkins-core:2.222.3
[2020-05-01T00:13:50.128Z]     +-org.slf4j:jcl-over-slf4j:1.7.26

pom.xml has not been modified by this PR. Build error seems to be caused because the project use jenkins version 2.138.4 but build system use recomanded configuration that use 2.222.x This make enforcer fails.

src/main/java/jenkins/branch/NamedExceptionsBranchPropertyStrategy.java
```diff
@@ -101,12 +101,19 @@ public NamedExceptionsBranchPropertyStrategy(@CheckForNull BranchProperty[] defa
     @NonNull
     @Override
     public List<BranchProperty> getPropertiesFor(SCMHead head) {
+        List<BranchProperty> properties = new ArrayList<>();
+
         for (Named named : namedExceptions) {
             if (named.isMatch(head)) {
-                return new ArrayList<>(named.getProps());
+                properties.addAll(named.getProps());
```
dwnusbaum, line 108: This looks unrelated to the main purpose of the PR, and it seems like it could be a breaking change because branch jobs for existing projects could end up with different properties after the update. Why are the changes to this method needed?

pom.xml
```diff
@@ -67,7 +67,6 @@
   <properties>
     <revision>2.5.7</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.138.4</jenkins.version>
```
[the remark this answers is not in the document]
nfalco79: pom.xml changes was a tentative to the build work because 2 days ago the pipeline library has update the jenkins version to 2.222.3 for recomanded version. I would include these changes in the PR if the fix is already in master.

pom.xml
```diff
@@ -109,7 +108,6 @@
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
-      <version>1.8.5</version>
```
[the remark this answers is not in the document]
nfalco79: Maybe also this a tentative. At the time of writing this PR build passed. I will revert any changes to the pom.xml

src/main/java/jenkins/branch/NamedExceptionsBranchPropertyStrategy.java
nfalco79: This changes is required to this functionality, I explain this in [previous comment](https://github.com/jenkinsci/branch-api-plugin/pull/160#issuecomment-520864423). In short if you define a named exception for branch master with property A and other named expection for master and feature/* branches with other property B the property B will be ignored. The strategy return a list of named but for how is written it always returns one.
Just an example of our configuration in production
immagine

At the time of writing these test cases passed, unit test added to october fails. i will try to have a look for failing test cases

pom.xml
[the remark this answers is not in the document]
nfalco79: anyway thank for attention this PR was there for too long time and every time you release new version I have to rebase update and load to our production a custom build. And also advise all other devops to not update this plugin. Time ago I forgot this and I update plugin. This case the configuration of branchproperty do not work and cause our jenkins production down for a couple of hours to understand what happens and to restore all lost configurations. Very very dangerous

src/main/java/jenkins/branch/NamedExceptionsBranchPropertyStrategy.java
dwnusbaum: Ok, but it still seems like that is totally separate from "Add BranchProperty support to OrganizationFolder", this change is "Allow more than one named exception to match a single branch".

"Add BranchProperty support to OrganizationFolder" is something that does not affect existing users, it is a new feature.

"Allow more than one named exception to match each individual branch" is a change in existing behavior for multibranch projects (it sounds like you consider it a bug fix), and in the worst case it could break existing users. Without this change, you can duplicate the property for the `master, support/*` exclusion in the `master` exclusion and remove `master` from the `master, support/*` exclusion to get things to work the way you want.

nfalco79: as workaround you can duplicate the definition to each branch you would the property. But also keep them aligned for each expcetion. I think there is not reason to make an interfaces to return a list of named branch and than returns always one. Also since is not written an explict limitation that only the first named exception is taked than from the point of view of end user this looks a bug. Do you agree?

nfalco79: >  (it sounds like you consider it a bug fix), and in the worst case it could break existing users

since does not works except for the first matching exception there are no users that have named exception for the master (simply because actually does not work). This mean that could be considered a not breaking changes.

dwnusbaum: > from the point of view of end user this looks a bug. Do you agree?

I'm not sure, but my point is mostly that I think it is totally distinct from "Add BranchProperty support to OrganizationFolder", so I would open a separate PR with just that change so it can be discussed separately.

> I think there is not reason to make an interfaces to return a list of named branch and than returns always one

I'm not sure what you mean here. The class has a list of exceptions, but there is no documentation on whether more than one exception can match a single branch, so the fact that the existing code returns when it finds a matching exception rather than checking all of them seems like reasonable behavior to me.

> since does not works except for the first matching exception there are no users that have named exception for the master (simply because actually does not work)

The problem is that someone might have an exception for `master`, and then a later exception that would also match `master` (perhaps unintentionally, or accidentally), and so with this change the properties that `master` would get would change. Maybe that is a very minor risk and so we do not care, I don't know, but I think it should be discussed in a separate PR.

dwnusbaum:
@nfalco79 just a heads up @car-roll is working on fixing the build errors that are present even in master over in #192, so you can probably wait for that to be merged to fix most of the issues with pom.xml and the existing tests.

nfalco79:
i  finally resolve marking slf4j suite to 1.7.26 version, so that work with jekins 2.222.3 version on build

nfalco79:
now test pass on linux jdk8 but not on jdk11 because of mockito
> Mockito cannot mock this class: class jenkins.branch.BranchCategoryFilterTest$MockSCMSource.
[2020-05-01T18:57:14.868Z] Most likely it is a private class that is not visible by Mockito

src/main/java/jenkins/branch/NamedExceptionsBranchPropertyStrategy.java
nfalco79: Separate PR avaiable #193

nfalco79:
I'm not able to replicate locally also I run Windows 10 openJDK 10.0.7 or Oracle JDK 11.0.2
using mvn clean verify -Djenkins.version=2.222.3 -Dtest=BranchCategoryFilterTest

nfalco79:
rebased

nfalco79:
@dwnusbaum the same test failing in my closed PR are failing also here:

- TestTimedOutException
- EventsTest

nfalco79:
@bitwiseman I had rebased (again), how much longer to reach the end since [this comment](https://github.com/jenkinsci/branch-api-plugin/pull/160#issuecomment-537407063)

This PR was just splitted as requested. Files was the same at the time of https://github.com/jenkinsci/branch-api-plugin/pull/160#pullrequestreview-295828387

nfalco79:
ping

bitwiseman:
@nfalco79 Thanks for sticking with this. This looks like a very useful feature, but other tasks keep taking priority. Sorry.  I'll look at this as soon as I have time.

nfalco79:
any news here

nfalco79:
@bitwiseman @dwnusbaum 
Another relase was passed any change to get this missing feature in the code base?

dwnusbaum:
@nfalco79 The changes that went into the last release are all trivial, mostly refactoring or test changes, and the release only went out to try to fix PCT issues when testing the plugin. This is a new feature, which means it takes more time to review (need think about design, other code that might be affected/impacted, compatibility, regressions, etc.), and I have not had time to review it yet.

nfalco79:
I can understand but the PR was open quite since 1 year. The feature is not different than one already exist for multibranch project. This allows users that use branch property to move to organisation project, share common settings and automate project creation/deletion. Every branch-api release I have to patch manually and load on ours jenkins production.
No other way to patches are available. Just a few of months ago our team update the plugin and forgot the patch, jenkins was offline for 4 hours fortuanlly we had a backup of configurations.

bitwiseman:
@nfalco79 
There will be another release today with trivial changes. 
Please be patient.  
I'll look at this next week. 
We will look at this next week.

bitwiseman:
@nfalco79 
I'm cutting another release with test fixes today.  I've a task to my work queue to look at this next week.  This is going to get done. Okay?

nfalco79:
Ok, let me update., and thanks.

bitwiseman:
Bounced for CI.  :sob: 

I've had a chance to look at it and I get what is being done. I've had a chance to manually play with it a little and it looks good so far.

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -334,6 +355,26 @@ public boolean isSingleOrigin() {
...
+     */
+    public BranchPropertyStrategy getStrategy() {
+        return strategy != null ? strategy : new DefaultBranchPropertyStrategy(new BranchProperty[0]);
+    }
+
+    /**
+     * Sets the branch property strategy.
+     *
+     * @param strategy chosen.
+     * @since 2.5.6
```
bitwiseman, line 372:
```suggestion
     * @since 2.5.9
```

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -334,6 +355,26 @@ public boolean isSingleOrigin() {
         return projectFactories;
     }
 
+    /**
+     * Gets the strategy.
+     *
+     * @return the strategy.
+     * @since 2.5.6
```
bitwiseman, line 362:
```suggestion
     * @since 2.5.9
```

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -163,6 +172,13 @@
      */
     private transient String facDigest;
 
+    /**
+     * The {@link #propertyStrategy} digest used to detect if we need to trigger a rescan on save.
+     *
+     * @since 2.5.6
```
bitwiseman, line 178:
```suggestion
     * @since 2.5.9
```

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -142,6 +144,13 @@
      */
     private DescribableList<BranchBuildStrategy, BranchBuildStrategyDescriptor> buildStrategies = new DescribableList<>(this);
 
+    /**
+     * The branches properties.
+     *
+     * @since 2.5.6
```
bitwiseman, line 150:
```suggestion
     * @since 2.5.9
```

bitwiseman:
@nfalco79  Thanks for you endurance and perseverance on this PR.

jglick:
This seems to make `NoTriggerOrganizationFolderProperty` _almost_ redundant—you could use the `NoTriggerBranchProperty` configured at the org level—except for the fact that the `NoTriggerOrganizationFolderProperty.branches` regexp is more powerful than the `NamedExceptionsBranchPropertyStrategy.Named.name` glob. Perhaps there could be a new `BranchPropertyStrategy` based on regexp matches? In that case `NoTriggerOrganizationFolderProperty` could be automatically replaced with the newer more general configuration style in `OrganizationFolder.readResolve`.

PR #164, opened 2019-07-17
[JENKINS-54864] Disable NoTriggerOrganizationFolderProperty migration and undeprecate the property for now

dwnusbaum:
See [JENKINS-54864](https://issues.jenkins-ci.org/browse/JENKINS-54864). 

This PR disable the automatic migration of `NoTriggerOrganizationFolderProperty` to a `Named branch` build strategy and undeprecates the property for now. Affects the migration and admin monitor added in #122. 

CC @jtnord @vivek


src/main/resources/jenkins/branch/Messages.properties
```diff
@@ -54,7 +54,14 @@ OrganizationFolder.OrJoinN.First={0}, {1}
...
 TaskNounUiTextProvider.TaskNoun=Scan
-NoTriggerOrganizationFolderProperty.PropertyMigration=The "Automatic branch project triggering" option has been replaced with a "Named branch" build strategy
+NoTriggerOrganizationFolderProperty.PropertyDeprecation=The "Automatic branch project triggering" option \
+    is deprecated. Use a "Named branch" build strategy from Basic Branch Build Strategies Plugin instead.
+NoTriggerOrganizationFolderProperty.PropertyMigrationWarning=The "Automatic branch project triggering" option \
+    for organization folders is deprecated and will be migrated to a "Named branch" build strategy from \
+    Basic Branch Build Strategies Plugin. If you were using "Automatic branch project triggering" to trigger \
+    builds for change requests (such as GitHub pull requests), then when you install version 1.1.0 or newer of \
+    Basic Branch Build Strategies Plugin, change requests will no longer be built. To preserve the previous behavior, \
+    you will need to manually reconfigure organization folders and add a "Change requests" build strategy.
```
jtnord: would it be possible to add a `href` here and a wiki page such that users can be given a fully worked example of what they need to do?

dwnusbaum: We certainly could, with pictures, etc. For now we are discussing possible followups to this PR. If we end up disabling the migration and/or reverting the deprecation, all of this might be deleted, so I'll hold off until we have some clarity on next steps.

dwnusbaum: 446e87b disables the migration completely and makes the admin monitor no longer show up at all, so now this text will never be seen. I think it makes sense to keep it here for any future work that might end up reenabling the migration.

src/main/java/jenkins/branch/NoTriggerOrganizationFolderProperty.java
```diff
@@ -72,28 +68,17 @@ public String getDisplayName() {
...
-                return FormValidation.warning(Messages.NoTriggerOrganizationFolderProperty_PropertyMigration());
+                return FormValidation.ok();
             } catch (PatternSyntaxException x) {
                 return FormValidation.error(x.getMessage());
             }
         }
 
     }
 
-    static boolean legacyCodeActive() {
```
dwnusbaum: This was being used to hide descriptors once basic-branch-build-strategies was installed, even though the migration didn't happen after the other changes. Thanks to @kshultzCB for catching this!

dwnusbaum:
From some testing earlier with @kshultzCB, users who have already been migrated to the Named Branch build strategy by installing or updating basic-branch-build-strategies should be unaffected by the change (they might still be broken as a result of the original migration, but fix that automatically will probably just make things worse), and users who have not yet migrated will no longer be encouraged to do so (and if they install basic-branch-build-strategies anyway the migration will not happen).

PR #165, opened 2019-08-02
Support markup formatter in description column

Jochen-A-Fuerbacher:
This PR adds the support for formating HTML tags in the description column properly.

Before:
before

After:
after

### DIFF
--- a/src/main/resources/jenkins/branch/DescriptionColumn/column.jelly
+++ b/src/main/resources/jenkins/branch/DescriptionColumn/column.jelly
@@ -4,10 +4,10 @@
     <j:set var="p" value="${col.getPropertyOf(job)}"/>
     <j:choose>
       <j:when test="${p != null and p.objectUrl != null}">
-        <a href="${p.objectUrl}">${col.description(p,job)}</a>
+        <a href="${p.objectUrl}"><j:out value="${col.description(p,job)}"/></a>
       </j:when>
       <j:otherwise>
-        ${col.description(p, job)}
+        <j:out value="${col.description(p, job)}"/>
       </j:otherwise>
     </j:choose>
   </td>
--- a/src/main/resources/jenkins/branch/DescriptionColumn/column.jelly
+++ b/src/main/resources/jenkins/branch/DescriptionColumn/column.jelly
@@ -4,10 +4,10 @@
     <j:set var="p" value="${col.getPropertyOf(job)}"/>
     <j:choose>
       <j:when test="${p != null and p.objectUrl != null}">
-        <a href="${p.objectUrl}"><j:out value="${col.description(p,job)}"/></a>
+        <a href="${p.objectUrl}"><j:out value="${col.formattedDescription(p,job)}"/></a>
       </j:when>
       <j:otherwise>
-        <j:out value="${col.description(p, job)}"/>
+        <j:out value="${col.formattedDescription(p, job)}"/>
       </j:otherwise>
     </j:choose>
   </td>
--- a/src/main/java/jenkins/branch/DescriptionColumn.java
+++ b/src/main/java/jenkins/branch/DescriptionColumn.java
@@ -76,12 +76,12 @@ public class DescriptionColumn extends ListViewColumn {
      *
      * @param p   the metadata action.
      * @param job the job.
-     * @return the description.
+     * @return the description. It is never unfiltered, unescaped HTML.
      * @throws IOException if there was an issue encoding the description.
      */
     @Restricted(NoExternalUse.class)
     @SuppressWarnings("unused") // used via Jelly EL binding
-    public String description(@CheckForNull Object p, @NonNull Object job) throws IOException {
+    public String formattedDescription(@CheckForNull Object p, @NonNull Object job) throws IOException {
         if (p instanceof ObjectMetadataAction) {
             // when the description comes from the metadata, assume plain text and use Util.escape
             String objectDescription = ((ObjectMetadataAction) p).getObjectDescription();
--- a/src/main/java/jenkins/branch/DescriptionColumn.java
+++ b/src/main/java/jenkins/branch/DescriptionColumn.java
@@ -72,7 +72,7 @@ public class DescriptionColumn extends ListViewColumn {
     }
 
     /**
-     * Gets the description of a job.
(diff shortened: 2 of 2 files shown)


Jochen-A-Fuerbacher:
This sould already be caught by OWASP Markup Formatter Plugin.

This is what happens, when you choose "Plain text" as Markup formatter:
**Screenshot of Job:**
plain_text_01
**Screenshot of View:**
plain_text_02

And this is what happens, when you install OWASP Markup Formatter Plugin and choose "Safe HTML" as markup formatter:
**Screenshot of Job:**
safe_html_01
**Screenshot of View:**
safe_html_02

In both cases no XSS get performed.

dwnusbaum:
@Jochen-A-Fuerbacher So your original `after` screenshot is with this change _and_ OWASP Markup Formatter Plugin installed and configured? And if you do not install+configure OWASP Markup Formatter Plugin then even with this change you still see the raw `This is a <a href=...>Link</a`?

I'm not familiar enough with how `MarkupFormatter` works to know the right way to integrate with it offhand, and couldn't find any directly analogous examples in a quick search (if you have any relevant links that would be helpful), so I'd want to do some testing to make sure there are no issues.

src/main/resources/jenkins/branch/DescriptionColumn/column.jelly
```diff
@@ -4,10 +4,10 @@
     <j:set var="p" value="${col.getPropertyOf(job)}"/>
     <j:choose>
       <j:when test="${p != null and p.objectUrl != null}">
-        <a href="${p.objectUrl}">${col.description(p,job)}</a>
+        <a href="${p.objectUrl}"><j:out value="${col.description(p,job)}"/></a>
       </j:when>
       <j:otherwise>
-        ${col.description(p, job)}
+        <j:out value="${col.description(p, job)}"/>
```
dwnusbaum: Ok, I asked around, and normally you would need an explicit call to `MarkupFormatter.translate` like [this](https://github.com/jenkinsci/jenkins/blob/22aa2e6e766074d11249893e3f35e0b99e20d3d0/core/src/main/resources/hudson/widgets/HistoryWidget/entry.jelly#L72) or [this](https://github.com/jenkinsci/credentials-plugin/blob/6f710a7e30156e94aff8a3c6c6799cef04ad35df/src/main/resources/com/cloudbees/plugins/credentials/CredentialsStoreAction/DomainWrapper/index.jelly#L33). However, in this case, `DescriptionColumn.description(Object, Object)` does that call internally: https://github.com/jenkinsci/branch-api-plugin/blob/db3c6487e3fa9f0214b6e50e5808eba84b531aff/src/main/java/jenkins/branch/DescriptionColumn.java#L95

So I think that makes this change safe as-is. Probably worth a comment here explaining why, since normally it would not be. Maybe clearer to make the Java code return the raw content and do the escaping in Jelly, although you'd need to make sure other uses of `DescriptionColumn.description(Object, Object)` didn't have any issues and that escaping the result of the first branch from the Java code doesn't break anything.

Jochen-A-Fuerbacher:
@dwnusbaum Correct. The original _before_ screenshot were done without any changes and the original _after_ screenshot were done with the changes, installed OWASP Markup Formatter and selected "Safe HTML" in the global security settings.

With my changes and no change to the markup formatter (_plain text_), I still see `This is a <a href=...>Link</a>`.

src/main/resources/jenkins/branch/DescriptionColumn/column.jelly
Jochen-A-Fuerbacher: With this return value, the job description gets translated into HTML according to the selected markup formatter. In OWASP Markup Formatter that's done in [RawHtmlMarkupFormatter.translate()](https://github.com/jenkinsci/antisamy-markup-formatter-plugin/blob/c21daf9debe89aa45314ede7cf25f6f734ab3002/src/main/java/hudson/markup/RawHtmlMarkupFormatter.java#L33-L52), using the [MyspacePolicy](https://github.com/jenkinsci/antisamy-markup-formatter-plugin/blob/c21daf9debe89aa45314ede7cf25f6f734ab3002/src/main/java/hudson/markup/MyspacePolicy.java).

The default markup formatter in the Jenkins core however is using the [plain text markup formatter](https://github.com/jenkinsci/jenkins/blob/a5db82c0253dc88d72aa13ebeab225f0db2aa46c/core/src/main/java/hudson/markup/EscapedMarkupFormatter.java#L49-L51), that escapes the syntax by the rules defined in the [Util class](https://github.com/jenkinsci/jenkins/blob/a5db82c0253dc88d72aa13ebeab225f0db2aa46c/core/src/main/java/hudson/Util.java#L908-L946).

Jochen-A-Fuerbacher:
@daniel-beck Could you look at this and give a statement if this PR might lead to a XSS vulnerability?

daniel-beck:
Looks reasonable, Devin discovered already why.

Would recommend making this self-documenting by renaming https://github.com/jenkinsci/branch-api-plugin/blob/db3c6487e3fa9f0214b6e50e5808eba84b531aff/src/main/java/jenkins/branch/DescriptionColumn.java#L84 (it's NoExternalUse anyway) e.g. to `formattedDescription`, as well as fixing up the Javadoc to explain it is never to return unfiltered, unescaped HTML.

bitwiseman:
@Jochen-A-Fuerbacher It looks like there's a request to change the name and some javadoc text.  Do you have time to do that?

Jochen-A-Fuerbacher:
@stefanbrausch @RobinRSchulz Could you deal with this? I gave both of you push rights to my fork repo [Jochen-A-Fuerbacher:html-in-column](https://github.com/Jochen-A-Fuerbacher/branch-api-plugin/tree/html-in-column).

bitwiseman:
@Jochen-A-Fuerbacher @stefanbrausch @RobinRSchulz  Circling back.  We'll be doing a release soon and would like to get this added.

Jochen-A-Fuerbacher:
@bitwiseman Done.

bitwiseman:
@dwnusbaum  Does this look good now to you?

daniel-beck:
The PR title freaks me out every time I see it, hope it's OK I renamed it.

PR #166, opened 2019-08-19
Why not to expose commit hash from SCM as an environment variable?

ababushk:
I've started to use GitHub Branch Source plugin for my project. While developing my pipeline I was trying to set PR commit status for which I commit hash was needed. I was wondering if there any way to get it without `checkout` step. It didn't look too impossible - in Multibranch Pipelines and Organization Folders Jenkins "knows" from which commit has Jenkinsfile was taken. I've written the this https://gist.github.com/ababushk//930980d79ab09c74da34be459563f760 and added it to my Global Library. But it looks so "hacky" that I've started to search for a place where other environment variables like `CHANGE_ID` and `BRANCH_NAME` are created. I've found `BranchNameContributor` class to which I've added variable that I desire so much.
I'm completely new in Jenkins plugin development and Java in general, but I'm wondering why such simple thing as commit hash is not available by default.

### DIFF
--- a/src/main/java/jenkins/branch/BranchNameContributor.java
+++ b/src/main/java/jenkins/branch/BranchNameContributor.java
@@ -34,6 +34,7 @@ import java.io.IOException;
 import java.util.Date;
 import jenkins.scm.api.SCMHeadOrigin;
 import jenkins.scm.api.SCMHead;
+import jenkins.scm.api.SCMRevision;
 import jenkins.scm.api.metadata.ContributorMetadataAction;
 import jenkins.scm.api.metadata.ObjectMetadataAction;
 import jenkins.scm.api.mixin.ChangeRequestSCMHead;
@@ -57,10 +58,14 @@ public class BranchNameContributor extends EnvironmentContributor {
             BranchProjectFactory projectFactory = ((MultiBranchProject) parent).getProjectFactory();
             if (projectFactory.isProject(j)) {
                 Branch branch = projectFactory.getBranch(j);
+                SCMRevision rev = projectFactory.getRevision(j);
                 SCMHead head = branch.getHead();
                 // Note: not using Branch.name, since in the future that could be something different
                 // than SCMHead.name, which is what we really want here.
                 envs.put("BRANCH_NAME", head.getName());
+                if (rev != null) {
+                    envs.put("CHANGE_REVISION", rev.toString());
+                }
                 if (head instanceof ChangeRequestSCMHead) {
                     envs.putIfNotNull("CHANGE_ID", ((ChangeRequestSCMHead) head).getId());
                     SCMHead target = ((ChangeRequestSCMHead) head).getTarget();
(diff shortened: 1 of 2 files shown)


src/main/java/jenkins/branch/BranchNameContributor.java
```diff
@@ -57,10 +58,14 @@ public void buildEnvironmentFor(Job j, EnvVars envs, TaskListener listener) thro
             BranchProjectFactory projectFactory = ((MultiBranchProject) parent).getProjectFactory();
             if (projectFactory.isProject(j)) {
                 Branch branch = projectFactory.getBranch(j);
+                SCMRevision rev = projectFactory.getRevision(j);
                 SCMHead head = branch.getHead();
                 // Note: not using Branch.name, since in the future that could be something different
                 // than SCMHead.name, which is what we really want here.
                 envs.put("BRANCH_NAME", head.getName());
+                if (rev != null) {
+                    envs.put("CHANGE_REVISION", rev.toString());
```
bitwiseman: I'm not sure that `rev.toString()` is the right thing here, as it is intended to be 
["human-readable"](https://github.com/jenkinsci/scm-api-plugin/blob/master/src/main/java/jenkins/scm/api/SCMRevision.java#L68).  It may work in some of implementations but it, I'm not sure.

It looks like `SCMRevision` doesn't have a method for what you're trying to do.  There probably needs to be a  `getId()` method added to that in order for this to work reliably. Note, we'll also need to handle [ChangeRequestSCMRevision](https://github.com/jenkinsci/scm-api-plugin/blob/master/src/main/java/jenkins/scm/api/mixin/ChangeRequestSCMRevision.java) differently since it has several possible revisions that users would be interested in.  

It might be better if each SCM source implemented their own environment contributor.

bitwiseman (changes requested):
Thanks for submitting this PR!  

This is a common request and it is great to get some traction on it.  
Unfortunately, the situation is a little more complex that it might seem at first. 

Perhaps we could discuss a bit and I can help you figure what something that will work overall. 

Thanks!

ababushk:
Hi @bitwiseman , thanks for taking a look at this!
My great desire is to retrieve information about revision from which Jenkinsfile was taken without calling 'checkout' step (which requires node context).
This PR wasn't submitted as a complete solution, it is just quick'n'dirty way to get revision string. I'm completely new at Java and Jenkins plugins development, the only source of knowledge for me is doc files in this repository :) I agree that SCM source implementations should be able to contribute to the environment, I'll be able to add something like 'revision timestamp' then, and the way to retrieve it differs for GitHub and GitLab for example. Is there a way for an SCM Source implementation to contribute to the environment?

jglick:
> I was wondering if there any way to get it without `checkout` step. It didn't look too impossible

The return value of this step should be a map containing the information you need.

jglick:
(Though beware GitHub PR merge builds.)

src/main/java/jenkins/branch/BranchNameContributor.java
```diff
@@ -57,10 +58,14 @@ public void buildEnvironmentFor(Job j, EnvVars envs, TaskListener listener) thro
             BranchProjectFactory projectFactory = ((MultiBranchProject) parent).getProjectFactory();
             if (projectFactory.isProject(j)) {
                 Branch branch = projectFactory.getBranch(j);
+                SCMRevision rev = projectFactory.getRevision(j);
```
jglick, line 61: The revision changes during the lifetime of the branch project, whereas this is being called on the `Job`, so this cannot possibly be right.

jglick (changes requested):

jglick:
see https://github.com/jenkinsci/branch-api-plugin/issues/728

PR #72, opened 2017-01-18
[JENKINS-41124] Move name mangling support responsibility to AbstractFolder

stephenc:
See [JENKINS-41124](https://issues.jenkins-ci.org/browse/JENKINS-41124)
Downstream of https://github.com/jenkinsci/cloudbees-folder-plugin/pull/82

Additional test cases:
- [x] Need a test that verifies after migration and then a restart that all is still as expected.
- [x] Need a test that verifies after a migration and then a reload that all is still as expected.
- [x] Need a test that verifies reloading the folder and not all of Jenkins leaves all still as expected.
- [x] Need a test that verifies after a migration and thene a restart and then a reload that all is still as expected.
- [x] Need a set of all those migration tests using the migration from a 2.0.0 pristine set of items
- [x] Need a set of all those migration tests using the migration from a 1.x set of items that was migrated to 2.0.0 and then lands on this set
- [x] Need a set of migration tests that verifies a pristine instance is in the same state as a migrated instance

@reviewbybees esp @jglick @michaelneale

PR #168, opened 2019-10-11
[JENKINS-59756] Do not duplicate input in NameEncoder.decode

dwnusbaum:
See [JENKINS-59756](https://issues.jenkins-ci.org/browse/JENKINS-59756).

Somehow, encoding logic got mixed up in `NameEncoder.decode`, causing all inputs to that method that contained a `%` to have their contents duplicated in addition to not totally decoding special characters. For example, `NameEncoder.decode("test%2Ffoo")` returned `tteesstt%2/Fffoooo`. It looks like this method has been broken since it was added in https://github.com/jenkinsci/branch-api-plugin/pull/72.

I'm not totally sure why we haven't caught this until now. My understanding is that it causes the `name-utf8.txt` file for the computed folders for the branch jobs to double in size every time the job for the branch is loaded. We noticed this because a user got an OOME when the file got so large that the characters in it could no longer fit into a Java array.

Probably needs some thought around what to do for existing jobs whose `name-utf8.txt` files are already broken, and some more tests (because apparently this code path is not covered by any existing tests in the plugin). We might also want to handle upper and lowercase versions of the characters with special meaning here, not totally sure about that.

### DIFF
--- a/src/test/java/jenkins/branch/NameEncoderTest.java
+++ b/src/test/java/jenkins/branch/NameEncoderTest.java
@@ -1,0 +1,46 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2019 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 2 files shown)


kshultzCB:
Accompanying bug report is [JENKINS-59756](https://issues.jenkins-ci.org/browse/JENKINS-59756).

dwnusbaum:
`RateLimitBranchPropertyTestFailed.rateLimitsBlockBuilds_medRate` failed on the Windows branch, Linux branches were ok:

```
Expected: (a value equal to or greater than <6800L> and a value less than or equal to <3500000L>)
     but: a value equal to or greater than <6800L> <6537L> was less than <6800L>
```
There were some IO errors in the logs, unclear if they are related:
```
SEVERE	hudson.model.Run#execute: Failed to save build record
java.nio.channels.ClosedByInterruptException
```
Closing and reopening for a fresh build.

dwnusbaum:
I think the reason that this wasn't caught earlier is that it only affects folders whose layout on disk was migrated when #72 landed (and maybe only the contents of `name-utf8.txt` for those folders, not sure if the actual name is also affected). Newly created folders do not use this code. The only other place it is used is as a fallback when a folder cannot be found using a non-decoded name, in which case the bug is harmless and just makes the fallback pointless.

stephenc:
Closing and re-opening to trigger a rebuild

src/test/java/jenkins/branch/NameEncoderTest.java
```diff
@@ -0,0 +1,46 @@
...
+package jenkins.branch;
+
+import org.junit.Test;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.junit.Assert.assertThat;
+
+public class NameEncoderTest {
+
+    @Test public void smokes() throws Exception {
```
stephenc:
```suggestion
    @Test 
    public void smokes() throws Exception {
```

bitwiseman:
@stephenc 
I respect that you felt this was ready to go, but it was still in draft state and we didn't have enough reviewers for this.

PR #169, opened 2019-10-15
[FIXED JENKINS-59786] Report all the health reports of the primary branch, not just the worst one

stephenc:
See https://issues.jenkins-ci.org/browse/JENKINS-59786

### DIFF
--- a/src/main/java/jenkins/branch/PrimaryBranchHealthMetric.java
+++ b/src/main/java/jenkins/branch/PrimaryBranchHealthMetric.java
@@ -25,6 +25,7 @@
 package jenkins.branch;
 
 import com.cloudbees.hudson.plugins.folder.AbstractFolder;
+import com.cloudbees.hudson.plugins.folder.Folder;
 import com.cloudbees.hudson.plugins.folder.health.FolderHealthMetric;
 import com.cloudbees.hudson.plugins.folder.health.FolderHealthMetricDescriptor;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
@@ -33,7 +34,11 @@ import hudson.Util;
 import hudson.model.Actionable;
 import hudson.model.HealthReport;
 import hudson.model.Item;
+import hudson.model.Job;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import jenkins.scm.api.metadata.PrimaryInstanceMetadataAction;
 import org.kohsuke.stapler.DataBoundConstructor;
@@ -111,7 +116,23 @@ public class PrimaryBranchHealthMetric extends FolderHealthMetric {
                     && ((Actionable) item).getAction(PrimaryInstanceMetadataAction.class) != null) {
                 if (reports == null) {
                     reports = new ArrayList<>();
-                    reports.add(getHealthReport(item));
+                    reports.addAll(Util.fixNull(getHealthReports(item)));
+                }
+            }
+        }
+
+        public static List<HealthReport> getHealthReports(Item item) {
+            if (item instanceof Job) {
+                return ((Job<?,?>) item).getBuildHealthReports();
+            } else if (item instanceof Folder) {
+                return ((Folder) item).getBuildHealthReports();
+            } else {
+                try {
+                    Method getBuildHealth = item.getClass().getMethod("getBuildHealthReports");
+                    return (List<HealthReport>) getBuildHealth.invoke(item);
+                } catch (InvocationTargetException | IllegalAccessException | NoSuchMethodException var2) {
+                    HealthReport report = getHealthReport(item);
+                    return report == null ? null : Collections.singletonList(report);
                 }
             }
         }
--- a/src/test/java/jenkins/branch/PrimaryBranchHealthMetricTest.java
+++ b/src/test/java/jenkins/branch/PrimaryBranchHealthMetricTest.java
@@ -3,10 +3,13 @@ package jenkins.branch;
 import hudson.model.FreeStyleProject;
 import hudson.model.TopLevelItem;
 import integration.harness.BasicMultiBranchProject;
+import integration.harness.HealthReportingMultiBranchProject;
 import java.util.Collections;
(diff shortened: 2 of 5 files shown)


kshultzCB:
@stephenc , thanks for asking. I've got a handful of questions. Background:

* I've [merged 168 into this PR](https://github.com/kshultzCB/branch-api-plugin/tree/169-with-168), built it, and installed it.
* I'm building a multibranch project based on [this repo](https://github.com/kshultzCB/creates-and-deletes). As you can see in the [master branch's commit history](https://github.com/kshultzCB/creates-and-deletes/commits/master), I've created some long running builds, some intentionally failing builds, and some successful ones. Build history for master looks like this now:

  image

* The repo I'm building also has a branch in it called [intentionally-broken-branch](https://github.com/kshultzCB/creates-and-deletes/tree/intentionally-broken-branch), which as its name implies, [fails its build](https://github.com/kshultzCB/creates-and-deletes/blob/intentionally-broken-branch/Jenkinsfile#L5).

OK, cool...but what I'm seeing here when I back out a level in Jenkins and hover over the MBP's weather icon, is this:
image

Which is telling me only about the failed builds in intentionally-broken-branch.

Is this the expected behavior?

src/test/java/jenkins/branch/PrimaryBranchHealthMetricTest.java
```diff
@@ -98,8 +103,12 @@ public void given__multibranch_with_primary__when__reporting_health__then__non_e
...
             FreeStyleProject stable = prj.getItem("stable");
             assertThat("We now have the stable branch", stable, notNullValue());
-            assertThat(prj.getBuildHealthReports(), not(containsInAnyOrder(master.getBuildHealth())));
-            assertThat(prj.getBuildHealthReports(), containsInAnyOrder(stable.getBuildHealth()));
+            assertThat(prj.getBuildHealthReports(), not(containsInAnyOrder(
+                    master.getBuildHealthReports().stream().map(Matchers::is).collect(Collectors.toList())
+            )));
+            assertThat(prj.getBuildHealthReports(), containsInAnyOrder(
+                    stable.getBuildHealthReports().stream().map(Matchers::is).collect(Collectors.toList())
+            ));
```
kshultzCB: A `reason` in these two `assertThat`s would really help explain the condition being tested for.

src/test/java/jenkins/branch/PrimaryBranchHealthMetricTest.java
```diff
@@ -77,7 +80,9 @@ public void given__multibranch_with_primary__when__reporting_health__then__prima
             prj.invalidateBuildHealthReports();
             FreeStyleProject master = prj.getItem("master");
             assertThat("We now have the master branch", master, notNullValue());
-            assertThat(prj.getBuildHealthReports(), containsInAnyOrder(master.getBuildHealth()));
+            assertThat(prj.getBuildHealthReports(), containsInAnyOrder(
+                    master.getBuildHealthReports().stream().map(Matchers::is).collect(Collectors.toList())
+            ));
```
kshultzCB: A `reason` in this `assertThat` would help explain the condition being tested for.

src/test/java/jenkins/branch/PrimaryBranchHealthMetricTest.java
stephenc: The only reason for the ‘stream’,’map’,’collect’ is that the `containsInAnyOrder(List)` expects a list of matchers not a list of instances

stephenc: We want to verify the list has all the primary branches reports and none of the other’s

bitwiseman:
@stephenc 
Did you answer Karl's question and I just didn't see it?

src/main/java/jenkins/branch/PrimaryBranchHealthMetric.java
```diff
@@ -111,7 +116,23 @@ public void observe(Item item) {
...
                     reports = new ArrayList<>();
-                    reports.add(getHealthReport(item));
+                    reports.addAll(Util.fixNull(getHealthReports(item)));
+                }
+            }
+        }
+
+        public static List<HealthReport> getHealthReports(Item item) {
+            if (item instanceof Job) {
+                return ((Job<?,?>) item).getBuildHealthReports();
```
bitwiseman: Are there performance effects or changing this from querying a single item to querying a set of items?

src/test/java/jenkins/branch/PrimaryBranchHealthMetricTest.java
stephenc: we want to check that the folder's health reports are the health reports of the master branch. We do not care about the order of the health reports (though it should be the same as long as they all have different `%` values). We need to wrap each health report in a matcher to stop the generic types being wrong

src/main/java/jenkins/branch/PrimaryBranchHealthMetric.java
stephenc: Should be faster as the single report needs top get the list first

stephenc:
@bitwiseman the original feature was added in https://github.com/jenkinsci/branch-api-plugin/pull/146

src/test/java/jenkins/branch/PrimaryBranchHealthMetricTest.java
bitwiseman: @stephenc 
The request was that you update the `assertThat` call to include a `reason` string.

src/test/java/jenkins/branch/PrimaryBranchHealthMetricTest.java
bitwiseman: @stephenc  
Please add `reason` strings to the `assertThat` method calls in the code.

stephenc: Personally I think the reason is redundant once you see past the noise of generic type erasure, but I have added reasons anyway

stephenc:
@bitwiseman what's stopping the merge?

bitwiseman:
Build failing?

PR #172, opened 2019-12-20
Add new method for performance improvements

res0nance:
Signed-off-by: Raihaan Shouhell <raihaan.shouhell@autodesk.com>

Currently just to check if there are no visible items it has to do a permissions check on every item.
This method just adds a quick exit which should improve performance especially considering this method is synchronized.

This method might make more sense in ComputedFolder considering https://github.com/jenkinsci/branch-api-plugin/blob/ad83debf8f48d6ad2661ec2549116b3df8cb0f5b/src/main/java/jenkins/branch/OrganizationFolder.java#L603

But I'd like to start the discussion on this


res0nance:
With organizationfolder looking like it needs the same api this is probably better handled in the folder plugin

PR #174, opened 2020-01-07
Reduce scope of locks in MultibranchProjectViewHolder and OrganizationFolderViewHolder

res0nance:
This PR:
- Reduces scope of locking for better performance
- Refactors code to reduce `hasPermission` calls by half

This was inspired by a stacktrace we often see:
```
"Handling POST /job/git-org/job/git-repo/ajaxExecutors from 10.148.73.74 : qtp2113273959-23348031 View/ajaxExecutors.jelly ExecutorStepExecution/PlaceholderTask/PlaceholderExecutable/executorCell.jelly" prio=5 BLOCKED
	jenkins.branch.MultiBranchProjectViewHolder.getPrimaryView(MultiBranchProjectViewHolder.java:126)
	com.cloudbees.hudson.plugins.folder.AbstractFolder$1.primaryView(AbstractFolder.java:266)
	hudson.model.ViewGroupMixIn.getPrimaryView(ViewGroupMixIn.java:171)
	com.cloudbees.hudson.plugins.folder.AbstractFolder.getPrimaryView(AbstractFolder.java:743)
	hudson.model.View.isDefault(View.java:421)
	hudson.model.AbstractItem.getUrl(AbstractItem.java:542)
	hudson.model.Run.getUrl(Run.java:994)
	org.jenkinsci.plugins.workflow.support.steps.ExecutorStepExecution$PlaceholderTask.getUrl(ExecutorStepExecution.java:515)
	sun.reflect.GeneratedMethodAccessor488.invoke(Unknown Source)
	sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	java.lang.reflect.Method.invoke(Method.java:498)
```

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProjectViewHolder.java
+++ b/src/main/java/jenkins/branch/MultiBranchProjectViewHolder.java
@@ -75,8 +75,8 @@ public class MultiBranchProjectViewHolder extends AbstractFolderViewHolder {
     /**
      * The primary view name.
      */
-    @SuppressFBWarnings(value = "IS2_INCONSISTENT_SYNC", justification = "Only need to synchronize initialization")
-    private transient String primaryView = null;
+    @GuardedBy("this")
+    private transient volatile String primaryView = null;
 
     /**
      * Constructor.
--- a/src/main/java/jenkins/branch/MultiBranchProjectViewHolder.java
+++ b/src/main/java/jenkins/branch/MultiBranchProjectViewHolder.java
@@ -74,7 +74,6 @@ public class MultiBranchProjectViewHolder extends AbstractFolderViewHolder {
     /**
      * The primary view name.
      */
-    @GuardedBy("this")
     private transient String primaryView = null;
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/jenkins/branch/MultiBranchProjectViewHolder.java
```diff
@@ -70,11 +71,11 @@
...
      */
     @GuardedBy("this")
-    private transient List<View> views = null;
+    private transient volatile List<View> views = null;
     /**
      * The primary view name.
      */
-    @GuardedBy("this")
+    @SuppressFBWarnings(value = "IS2_INCONSISTENT_SYNC", justification = "Only need to synchronize initialization")
     private transient String primaryView = null;
```
res0nance: Not 100% on if this should be volatile or not

src/main/java/jenkins/branch/OrganizationFolderViewHolder.java
```diff
@@ -68,11 +69,11 @@
...
      */
     @GuardedBy("this")
-    private transient List<View> views = null;
+    private transient volatile List<View> views = null;
     /**
      * The primary view name.
      */
-    @GuardedBy("this")
+    @SuppressFBWarnings(value = "IS2_INCONSISTENT_SYNC", justification = "Only need to synchronize initialization")
     private transient String primaryView = null;
```
res0nance: Ditto here

src/main/java/jenkins/branch/MultiBranchProjectViewHolder.java
dwnusbaum: Yeah, I'm not really sure if it's necessary in this case. I would probably make `primaryView` `volatile` for now to be safe, since there should still be a performance improvement because of the reduced locking, and if anyone wants to investigate the semantics in detail in the future in case `primaryView` doesn't need to be `volatile`, they can do so.

I guess the easier way to reason about it would be to put `views` and `primaryView` in a composite object that holds them via `final` fields instead of having them be directly in `MultiBranchProjectViewHolder` (at that point the holder might not even need `volatile` because it would be immutable?), which would make it a more typical case of double-checked locking instead of having two fields interact with the lock.

src/main/java/jenkins/branch/OrganizationFolderViewHolder.java
dwnusbaum: Same thoughts as above.

res0nance:
Seems to be a test flake on windows

PR #178, opened 2020-01-31
Fix some override TODOs

res0nance:
Signed-off-by: Raihaan Shouhell <raihaan.shouhell@autodesk.com>

These are well old TODOs so clean them up and inherit their documentation


res0nance:
@bitwiseman any chance of a release? master is fine as is the rest of the PRs are more nice to have but does not improve any features

PR #180, opened 2020-01-31
Follow up to JENKINS-39300

res0nance:
Simplify code, less is more 🤣


bitwiseman:
What am I looking at here?  Could you as some links or a bit more description?  Why is it okay to remove this?

src/main/java/jenkins/branch/LocalizedViewsTabBar.java
```diff
@@ -1,65 +0,0 @@
...
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.List;
-import javax.annotation.Nonnull;
-import org.kohsuke.accmod.Restricted;
-import org.kohsuke.accmod.restrictions.NoExternalUse;
-
-/**
- * JENKINS-39300 work-around. Replace with {@link DefaultViewsTabBar} once JENKINS-39300 is in baseline core.
```
res0nance: This apparently was a workaround because `DefaultViewsTabBar` was not in core

res0nance:
@bitwiseman Sorry about that. Didn't realize the TODO was not immediately obvious like the others.

PR #183, opened 2020-02-17
Test cleanups

res0nance:
Cleanup some of the deprecated code in tests


res0nance:
For some weird reason i get an NPE in core, so i avoided the permissions checks by using the map directly. It should improve the performance of the tests marginally but more importantly it seems to avoid this weird NPE edge case.

res0nance:
I'm going to rework this before opening another PR

PR #186, opened 2020-03-20
prevent building branches/PRs existing before the first branch indexing

agabrys:
Hello,
we provide CI for many teams in a big company. Unfortunately, we have a huge problem with server loads after the spin up process, after restarts and jobs modification. Multibranch jobs execute branch indexing on every single change and directly after the job creation. It is great, because people see what is available. On the other side, when people don't delete old branches, we waste a lot of resources on building old stuff. In our case it completely blocks servers for days. We tried to use build strategies, but none of them solved it. The closest solution was [SkipInitialBuildOnFirstBranchIndexing](https://github.com/jenkinsci/basic-branch-build-strategies-plugin/blob/master/src/main/java/jenkins/branch/buildstrategies/basic/SkipInitialBuildOnFirstBranchIndexing.java), but unfortunately it:
* blocks all new branches and PRs builds executed by webhooks (it always skip the first build no matter how it is triggered)
* won’t solve the problem when the branch indexing is executed for a second time (again, everything is build)

For now we use the patch visible in this PR (I'm aware that it won't be merged - it is just a workaround, I skipped tests). I wanted to ask - do you have any ideas on how to solve this properly? We are open to implement the proposed solution 🙂 

For now, I have two ideas to make it configurable:

1. add an option to disable automatic builds on branch indexing
1. extend [BranchBuildStrategy](https://github.com/jenkinsci/branch-api-plugin/blob/master/src/main/java/jenkins/branch/BranchBuildStrategy.java) class and pass information about the trigger cause

Both solutions don't cover the workaround solution, because they will only allow the skipping of all builds caused by branch indexing (we will lose the ability to execute builds of branches/PRs for which webhooks have not been received - for example, lost due to network problems).

### Disable Automatic build on branch indexing

The first solution requires to add a new property to the multibranch project and simply skip the build if `causeFactory instanceof IndexingCauseFactory`. I don't know how to add this configurable property, but I can get this knowledge 😉 

This approach does not change the API, just adds a new parameter.

### Extend BranchBuildStrategy API

This approach is backward compatible, but introduces new methods to API (in a class which is extended by many plugins).

How could we implement it:
* add a new `BranchBuildStrategy#isAutomaticBuild` method which takes one more parameter (the trigger cause - indexing vs. webhook):
```
public boolean isAutomaticBuild(@NonNull TriggerCause cause,
                                @NonNull SCMSource source,
                                @NonNull SCMHead head,
                                @NonNull SCMRevision currRevision,
                                @CheckForNull SCMRevision lastBuiltRevision,
                                @CheckForNull SCMRevision lastSeenRevision,
                                @NonNull TaskListener listener) {
    // by default delegate to the version without the cause
    return isAutomaticBuild(source, head, currRevision, lastBuiltRevision, lastSeenRevision, listener);
}
```
* add a new `BranchBuildStrategy#automaticBuild` which will be executed by the `MultiBranchProject`:
```
public final boolean automaticBuild(@NonNull TriggerCause cause,
                                    @NonNull SCMSource source,
                                    @NonNull SCMHead head,
                                    @NonNull SCMRevision currRevision,
                                    @CheckForNull SCMRevision lastBuiltRevision,
                                    @CheckForNull SCMRevision lastSeenRevision,
                                    @NonNull TaskListener listener) {
        if (Util.isOverridden(BranchBuildStrategy.class, getClass(), "isAutomaticBuild", TriggerCause.class,
                SCMSource.class, SCMHead.class, SCMRevision.class, SCMRevision.class, SCMRevision.class,
                TaskListener.class)) {
            return isAutomaticBuild(cause, source, head, currRevision, lastBuiltRevision, lastSeenRevision, listener);
        }
        return automaticBuild(source, head, currRevision, lastBuiltRevision, lastSeenRevision, listener);
}
```
It falls back to the original ``automaticBuild`` method (backward compatible).
* create a new build strategy (in new or existing plugin) `SkipBuildOnBranchIndexing`

---

Let me know what do you think.

Kind regards
Adam Gabryś

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -157,6 +157,8 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
 
     private transient String srcDigest, facDigest;
 
+    private boolean firstBranchIndexingExecuted;
+
     /**
      * Constructor, mandated by {@link TopLevelItem}.
      *
@@ -652,6 +654,7 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
             long end = System.currentTimeMillis();
             listener.getLogger().format("[%tc] Finished branch indexing. Indexing took %s%n", end,
                     Util.getTimeSpanString(end - start));
+            firstBranchIndexingExecuted = true;
         }
     }
 
@@ -2234,6 +2237,17 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
                                          @NonNull SCMRevision currRevision,
                                          @CheckForNull SCMRevision lastBuiltRevision,
                                          @CheckForNull SCMRevision lastSeenRevision) {
+            if (causeFactory instanceof IndexingCauseFactory) {
+                if (!firstBranchIndexingExecuted) {
+                    return false;
+                }
+                // the build will be scheduled only for branches and PRs created/modified after the
+                // first branch indexing (disable building everything after the job creation, but
+                // allows automatic build of branches/PRs which were not executed by webhook)
+                if (lastBuiltRevision == null && currRevision.equals(lastSeenRevision)) {
+                    return false;
+                }
+            }
             BranchSource branchSource = null;
             for (BranchSource s: MultiBranchProject.this.sources) {
                 if (s.getSource().getId().equals(source.getId())) {


agabrys:
@jglick do you know who we can ask? I'm sure that many other people have the same problem (blocking the server due to building a lot of branches after the job creation).

jglick:
Try the developers’ list.

agabrys:
I started a discussion:
https://groups.google.com/d/msg/jenkinsci-dev/NJoc_YQMy44/W9CzZYW3CAAJ

bitwiseman:
Responded on the email thread, but I'll copy here: 

Adam, 
I think the feature/behavior you're suggesting is definitely worth implementing.  It would let people safely create new projects without dealing with build storms.

So to reiterate - the behavior we're looking for is:

* Project is created
* Branches are indexed, but not built
* Polling (or hooks) now enabled
* Scheduled indexing now auto-builds new branches if selected. Hooks will trigger builds

You mentioned [SkipInitialBuildOnFirstBranchIndexing ](https://github.com/jenkinsci/basic-branch-build-strategies-plugin/blob/master/src/main/java/jenkins/branch/buildstrategies/basic/SkipInitialBuildOnFirstBranchIndexing.java).  I think something like that is way to go here.  Only instead of disabling the "first indexing of each branch", it would be "Skip Build on First Job Indexing".  I know there are several possible ways ways to detect the state "this is the first time a project has been indexed", but I'm not sufficiently knowledgable to be able to point you to the exact right spot.  Sorry. But I firmly believe this should be doable without changing the SCM or Branch API.  We just need to do some digging to find it.  

Sound good?

agabrys:
@jglick I think we found a solution which could be merged. The [NoTriggerBranchProperty](https://github.com/jenkinsci/branch-api-plugin/blob/branch-api-2.5.6/src/main/java/jenkins/branch/NoTriggerBranchProperty.java) skips builds if they are caused by SCM change (indexing or event). It means only builds scheduled manually are executed. The following line

https://github.com/jenkinsci/branch-api-plugin/blob/b99766caece8edb67adc0ce2dde8e3f0d2df67c5/src/main/java/jenkins/branch/NoTriggerBranchProperty.java#L84

test if the build should be executed (now always `false`). We can make the branch property parametrized - add `skipped actions`:
* `all` (default) - skips all
* `branch-indexing` skips builds scheduled by branch indexing
* `event` skips builds

Btw, this does not work:

https://github.com/jenkinsci/branch-api-plugin/blob/b99766caece8edb67adc0ce2dde8e3f0d2df67c5/src/main/java/jenkins/branch/NoTriggerBranchProperty.java#L78

because I see no option to add `OverrideIndexTriggersJobProperty` to the multibranch job.

gmshake:
Hi @agabrys, can you kindly create a jira issue to track this please?

gmshake:
Draft PRs to resolve this issue: 

1. https://github.com/jenkinsci/branch-api-plugin/pull/198 support detecting automatic build with causes 
2. https://github.com/jenkinsci/basic-branch-build-strategies-plugin/pull/10 new custom BranchBuildStrategy 

The `CustomSkipInitialBuildOnFirstBranchIndexing` 's goal:

- Skip first build only on branch indexing event(cause). This skip build of regular branches, pull requests and tags.
- Later, build only when changed detected. Compute changes by comparing current revision and last seen revision.

agabrys:
Hi @gmshake, sorry for very delayed answer, I missed your replies 😞 
I created a Jenkins ticket today: [JENKINS-63673 Allow configuring in NoTriggerBranchProperty which builds should be suppressed](https://issues.jenkins-ci.org/browse/JENKINS-63673)

I'll try to check your changes - I should have time in October.

jonesbusy:
Hi all! We basically have the same use case in our company. We realized that we broke our webhooks by using SkipInitialBuildOnFirstBranchIndexing...

Thanks

jzr1991:
Is this dead?

jmendiara:
@agabrys Thanks for your time and research on this issue. I'm about to deal with build storms in several big monorepos moving to Jenkins in the short time, and this issue will be a headache for us. 
Any movement from maintainers in adopting a solution will be hightly appreciated

agabrys:
I analyzed the topic a little bit deeper and I think I found a better solution (#244). There is a branch property responsible for suppressing SCM triggers (indexing and events). I made it configurable, so people may disable branch indexing and build only when webhooks are delivered. I think I should be able to make the PR ready for verification till end of this week.

For now I'm closing this PR.

fmarot:
I'd love to see it merged. Build storms are a problem when creating new job referencing old repositories.

PR #189, opened 2020-04-09
WorkspaceLocatorImpl should not use Node instances as monitors

jglick:
Looking at a thread dump from CloudBees Core (not that it likely matters) like

```
Handling GET /…/ from … View/index.jelly View/sidepanel.jelly
java.lang.Thread.State: BLOCKED (on object monitor)
	at com.cloudbees.opscenter.client.cloud.StandardCloudSlave.getNodeUsage(StandardCloudSlave.java:312)
	- waiting to lock <0x…> (a com.cloudbees.opscenter.client.cloud.StandardCloudSlave)
	at com.cloudbees.opscenter.client.cloud.StandardCloudSlave.getNumExecutors(StandardCloudSlave.java:381)
	at com.cloudbees.opscenter.client.cloud.OperationsCenterCloudComputer.isAcceptingTasks(OperationsCenterCloudComputer.java:71)
	at hudson.model.Computer.getIconClassName(Computer.java:753)
"Executor #0 for … : executing PlaceholderExecutable:ExecutorStepExecution.PlaceholderTask{…}" …
   java.lang.Thread.State: TIMED_WAITING (on object monitor)
	at java.lang.Object.wait(Native Method)
	at hudson.remoting.FastPipedInputStream.read(FastPipedInputStream.java:175)
	- locked <0x…> (a [B)
	at sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:284)
	at sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:326)
	at sun.nio.cs.StreamDecoder.read(StreamDecoder.java:178)
	- locked <0x…> (a java.io.InputStreamReader)
	at java.io.InputStreamReader.read(InputStreamReader.java:184)
	at java.io.BufferedReader.fill(BufferedReader.java:161)
	at java.io.BufferedReader.readLine(BufferedReader.java:324)
	- locked <0x…> (a java.io.InputStreamReader)
	at java.io.BufferedReader.readLine(BufferedReader.java:389)
	at jenkins.branch.WorkspaceLocatorImpl.load(WorkspaceLocatorImpl.java:222)
	at jenkins.branch.WorkspaceLocatorImpl.locate(WorkspaceLocatorImpl.java:159)
	- locked <0x…> (a com.cloudbees.opscenter.client.cloud.StandardCloudSlave)
	at jenkins.branch.WorkspaceLocatorImpl.locate(WorkspaceLocatorImpl.java:129)
	at jenkins.branch.WorkspaceLocatorImpl.locate(WorkspaceLocatorImpl.java:125)
	at hudson.model.Slave.getWorkspaceFor(Slave.java:340)
	at org.jenkinsci.plugins.workflow.support.steps.ExecutorStepExecution$PlaceholderTask$PlaceholderExecutable.run(ExecutorStepExecution.java:827)
	at hudson.model.ResourceController.execute(ResourceController.java:97)
	at hudson.model.Executor.run(Executor.java:427)
```

Setting aside why the agent was unresponsive to small Remoting requests, we should not be blocking like this.

Note that we are acquiring the lock _prior_ to potentially benefitting from the caching introduced in #188.

### DIFF
--- a/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
+++ b/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
@@ -58,6 +58,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
 import java.util.WeakHashMap;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -305,14 +306,13 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
         }
     }
 
-    private static final Map<Node, Object> nodeLocks = new WeakHashMap<>();
+    private final Map<String, Object> nodeLocks = new ConcurrentHashMap<>();
     private static Object lockFor(Node node) {
-        synchronized (nodeLocks) {
-            // Avoiding new Object() to prepare for http://cr.openjdk.java.net/~briangoetz/valhalla/sov/02-object-model.html
-            // Avoiding new String(…) because static analyzers complain
-            // Could use anything but hoping that a future JVM enhanced thread dumps to display monitors of type String
-            return nodeLocks.computeIfAbsent(node, _node -> new StringBuilder("WorkspaceLocatorImpl lock for ").append(node.getNodeName()).toString());
-        }
+        // Avoiding WeakHashMap<Node, T> since Slave overrides hashCode/equals
+        // Avoiding new Object() to prepare for http://cr.openjdk.java.net/~briangoetz/valhalla/sov/02-object-model.html
+        // Avoiding new String(…) because static analyzers complain
+        // Could use anything but hoping that a future JVM enhanced thread dumps to display monitors of type String
+        return ExtensionList.lookupSingleton(WorkspaceLocatorImpl.class).nodeLocks.computeIfAbsent(node.getNodeName(), name -> new StringBuilder("WorkspaceLocatorImpl lock for ").append(name).toString());
     }
 
     private static final Pattern GOOD_RAW_WORKSPACE_DIR = Pattern.compile("(.+)[/\\\\][$][{]ITEM_FULL_?NAME[}][/\\\\]?");
--- a/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
+++ b/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
@@ -172,7 +172,7 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
             throw new IllegalArgumentException("Dangerous job name `" + fullName + "`"); // better not to mess around
         }
         try {
-            synchronized (node) {
+            synchronized (lockFor(node)) {
                 Map<String, String> index = load(workspace);
                 // Already listed:
                 String path = index.get(fullName);
@@ -302,6 +302,16 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
         public Void invoke(File f, VirtualChannel channel) throws IOException, InterruptedException {
             new TextFile(f).write(text);
             return null;
+        }
+    }
+
+    private static final Map<Node, Object> nodeLocks = new WeakHashMap<>();
+    private static Object lockFor(Node node) {
+        synchronized (nodeLocks) {
+            // Avoiding new Object() to prepare for http://cr.openjdk.java.net/~briangoetz/valhalla/sov/02-object-model.html
+            // Avoiding new String(…) because static analyzers complain
+            // Could use anything but hoping that a future JVM enhanced thread dumps to display monitors of type String
+            return nodeLocks.computeIfAbsent(node, _node -> new StringBuilder("WorkspaceLocatorImpl lock for ").append(node.getNodeName()).toString());
         }
     }
 
@@ -459,7 +469,7 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
                         }
                         FilePath workspace = getWorkspaceRoot(node);
                         if (workspace != null) {
-                            synchronized (node) {
+                            synchronized (lockFor(node)) {
                                 Map<String, String> index = load(workspace);
                                 index.remove(tli.getFullName());
                                 save(index, workspace);
@@ -532,7 +542,7 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
                         }
                         FilePath workspace = getWorkspaceRoot(node);
                         if (workspace != null) {
-                            synchronized (node) {
+                            synchronized (lockFor(node)) {
                                 Map<String, String> index = load(workspace);
                                 index.remove(oldFullName);
                                 assert index.containsKey(newFullName); // locate(…, true) should have added it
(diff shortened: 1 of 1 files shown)


jglick:
Also blocked:

```
"AtmostOneTaskExecutor[Periodic Jenkins queue maintenance] [#…]" …
   java.lang.Thread.State: BLOCKED (on object monitor)
	at com.cloudbees.opscenter.client.cloud.StandardCloudSlave.getNodeUsage(StandardCloudSlave.java:312)
	- waiting to lock <0x…> (a com.cloudbees.opscenter.client.cloud.StandardCloudSlave)
	at com.cloudbees.opscenter.client.cloud.StandardCloudSlave.getNumExecutors(StandardCloudSlave.java:381)
	at com.cloudbees.opscenter.client.cloud.OperationsCenterCloudComputer.isAcceptingTasks(OperationsCenterCloudComputer.java:71)
	at hudson.model.Queue$JobOffer.getCauseOfBlockage(Queue.java:298)
	at hudson.model.Queue.maintain(Queue.java:1627)
	at hudson.model.Queue$1.call(Queue.java:333)
	at hudson.model.Queue$1.call(Queue.java:330)
	at …
```

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -305,6 +305,16 @@ public Void invoke(File f, VirtualChannel channel) throws IOException, Interrupt
         }
     }
 
+    private static final Map<Node, Object> nodeLocks = new WeakHashMap<>();
```
jtnord, line 308: I have seen some things elsewhere that explicitly do not use HashMaps for Nodes because the node hash() is not stable.  (e,.g. changing a name or something changes the actual hash).
in other words needs to really be a `WeakIdentityHashMap` which doesn't exist in stock JDK

jtnord: NB the [`Slave`](https://github.com/jenkinsci/jenkins/blob/3f15ad20985066661158ee1bb0b3255e000b3081/core/src/main/java/hudson/model/Slave.java#L552-L565) subclass is specifically safe for this - and the `Node` class does not implement hashCode so delegates to `Object` which is safe. however not all implementations may be

jglick: `Node` itself does not override `hashCode` or `equals`. `Slave` does, based on `.name` which is technically mutable but I think only ever actually set in the constructor or during deserialization: renaming a `Slave` would create a new instance I believe, via `@DataBoundConstructor`.

I would rather base this on object identity if something like `java.beans.WeakIdentityMap` were `public`, but I think it is OK since if you in fact reconfigure the agent (so that the current patch uses the same lock as before) there is no real harm done because the index cache is keyed off of `Channel` identity, which will have changed.

jglick: > not all implementations may be

Theoretically. I could select a lock based on the node name I guess, just in case.

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -305,6 +306,15 @@ public Void invoke(File f, VirtualChannel channel) throws IOException, Interrupt
         }
     }
 
+    private final Map<String, Object> nodeLocks = new ConcurrentHashMap<>();
+    private static Object lockFor(Node node) {
+        // Avoiding WeakHashMap<Node, T> since Slave overrides hashCode/equals
+        // Avoiding new Object() to prepare for http://cr.openjdk.java.net/~briangoetz/valhalla/sov/02-object-model.html
+        // Avoiding new String(…) because static analyzers complain
+        // Could use anything but hoping that a future JVM enhanced thread dumps to display monitors of type String
+        return ExtensionList.lookupSingleton(WorkspaceLocatorImpl.class).nodeLocks.computeIfAbsent(node.getNodeName(), name -> new StringBuilder("WorkspaceLocatorImpl lock for ").append(name).toString());
```
jtnord, line 315: this is safe but will leak for cloud agents that use unique node names.

jglick: True, but only leaks a shortish `String`.

We could also use your original suggestion of `node.getNodeName().intern()`…which would also introduce a similar small leak. No strong opinion.

jtnord: I'm trying to recall where we solved this last (without guava of course!)

jtnord: can't find it.

https://github.com/apache/avro/blob/d5d1cd70e920f45468bf696d3668b496a9adbab3/lang/java/avro/src/main/java/org/apache/avro/util/WeakIdentityHashMap.java is should be license compatible which we may be able to could copy and use here (and then pull into Jenkins core in the future).  WDYT?

String interning was moved from permgen in 1.7 so should be better, but there is the question of thrashing the pool.

1) use guava from core 
2) introduce a small leak
3) use String interning
4) use third party code (either an extra dep or using some appropriately licensed code that is copied into the codebase)
5) write our own WeakIdentityHashMap

I think arguments could be made either way, I guess we just need a consensus on this.

I generally don't like introducing leaks, and am against introducing guava.
Making heavy use of the intern pool would have some unknowns (not sure how much else makes use of the pool to know what implications that would have), so is a bit of an unknown.
So 4 or 5 seem slightly better, and not sure we should write our own if there are appropriatley licenced ones we could use.

WDYT?

jglick: > use guava from core

`CacheBuilder` with `weakKeys`? We already use this in core and Pipeline (https://github.com/jenkinsci/jenkins/blob/3f15ad20985066661158ee1bb0b3255e000b3081/core/src/main/java/hudson/model/Queue.java#L221 + https://github.com/jenkinsci/workflow-support-plugin/blob/173378ee6f1178b73f4ad1c18d0250401455db6a/src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java#L74) so I do not see why not. This would be my preference.

jtnord: > `CacheBuilder` with `weakKeys`

yes.

if you are happy with that then lets go with it.  not part of any serialised form so it at least safe from that PoV.

dwnusbaum:
> EC2 (aws) - Ubuntu 18.04 LTS (i-00bb64b4cd9cd523d) was marked offline

Rebuilding due to infra issues.

jglick:
https://issues.jenkins-ci.org/browse/INFRA-2548 FYI (CC @olblak)

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -305,6 +309,24 @@ public Void invoke(File f, VirtualChannel channel) throws IOException, Interrupt
         }
     }
 
+    // Avoiding WeakHashMap<Node, T> since Slave overrides hashCode/equals
+    private final LoadingCache<Node, Object> nodeLocks = CacheBuilder.newBuilder().weakKeys().build(new CacheLoader<Node, Object>() {
+        @Override
+        public Object load(Node node) throws Exception {
+            // Avoiding new Object() to prepare for http://cr.openjdk.java.net/~briangoetz/valhalla/sov/02-object-model.html
+            // Avoiding new String(…) because static analyzers complain
+            // Could use anything but hoping that a future JVM enhanced thread dumps to display monitors of type String
```
dwnusbaum, line 318: typo?

```suggestion
            // Could use anything but hoping that a future JVM enhances thread dumps to display monitors of type String
```

PR #191, opened 2020-04-15
Add 2.5.6 release notes and use ISO-8601 format for release dates

dwnusbaum:


CHANGELOG.md
```diff
@@ -94,15 +120,17 @@ Apr 8, 2019
 -   Jenkins core version bump to 2.138 LTS line
```
dwnusbaum, line 120: I think that line 110 breaks the markdown preview in the diff view, but it looks ok in the real view. I will try fixing it anyway.

kshultzCB: Yeah that's weird, it looks okay [here](https://github.com/jenkinsci/branch-api-plugin/blob/c0beb319cbf61d5f916cdced8f5d66d7013554ed/CHANGELOG.md).

kshultzCB: I think I found it. Can't put this as a suggestion since you didn't change it, but I think this edit to L110 fixes it up:
```markdown
> **NOTE** if you have been using some custom
```

For clarity, this is changing it from `**NOTE **if` to `**NOTE** if`

dwnusbaum: Yeah that was the problem, 07ec12d fixed it.

PR #194, opened 2020-05-02
Fix mockito issues with JDK11 and jenkins 2.222.3

nfalco79:
This PR will fix the issue of test cases in master when run on JDK11
The solution was remove mockito from this class so that now master can use the mockito version defined in the parent pom.

@dwnusbaum the issue is regardless the version of mockito. If run individually the test pass. If you run all together something play with cache of mockito/bytebuddy that causes the constructor not be visible.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>3.43</version>
+    <version>3.56</version>
     <relativePath />
   </parent>
 
@@ -67,10 +67,9 @@
   <properties>
     <revision>2.5.7</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.138.4</jenkins.version>
     <java.level>8</java.level>
-    <scm-api.version>2.4.1</scm-api.version>
-    <git-plugin.version>3.10.0</git-plugin.version>
+    <jenkins.version>2.176.4</jenkins.version>
+    <slf4jVersion>1.7.26</slf4jVersion>
   </properties>
 
   <repositories>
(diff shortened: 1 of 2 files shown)


nfalco79:
@dwnusbaum build passed

pom.xml
```diff
@@ -68,6 +68,7 @@
     <revision>2.5.7</revision>
     <changelist>-SNAPSHOT</changelist>
     <jenkins.version>2.138.4</jenkins.version>
+    <slf4jVersion>1.7.26</slf4jVersion>
```
dwnusbaum, line 71: This silences the enforcer, but it isn't the ideal way to fix the issue in Jenkins. `slf4j` and associated packages are dependencies of Jenkins core itself, so the version that the plugin specifies here is not actually used in practice (see [this page](https://www.jenkins.io/doc/developer/plugin-development/dependencies-and-class-loading/) for more information). The best way to fix this is to update the parent POM to a 4.x version, which uses a BOM to address conflicts with dependencies of Jenkins core itself, and then update `jenkins.version` here to a version that supports the BOM, e.g. `2.176.4`. After that, you will be able to remove this line, and `1.7.25` will automatically be used when running against the version of Jenkins and `1.7.26` will be used when running against the new version of Jenkins, and the enforcer will be happy.

nfalco79: It was my first tentative in other PR and I removed the jenkins.version to leave the organisation pom guide the jenkins version.
Do you prefer set a specific version of jenkins instead use that one defined on organisation pom?

nfalco79: Any specific reason why prefer 2.176.4 instead of that one defined in organisation pom (2.204) ?

nfalco79: With 2.176.4 test fails

nfalco79: With 2.222.3 test fails.

Override BOM plugin version with those defined in master test fails again.
Single test in eclipse not happens, debug in mvn clean test -Dmave.surefure.debug step by step not happens only run when run all together not in debug.
Seems a timing issue. The test look if a FreestyleProject master exists for top/stuff but it has not yet been by organisation folder scan.

bitwiseman: 2.176.4 is a conservative choice. 
Not sure what do about test failures.

nfalco79: The problem is that does not wok with nothing.
The organisation pom version is the cause <= 3.56 test passes. With version 3.57 fails
Using jenkins.version 2.176.4 test passes (locally) but will fail on CI system because runs with 2.222.3 and will fail on enforcer.

nfalco79:
now fails on different unit tests that always passes, re-trigger the build to verify if are flaky faillures

dwnusbaum:
@nfalco79 I think the root cause of all the remaining issues was static state from earlier tests interfering with later tests, and configuring surefire to use separate processes for each test suite (`<reuseForks>false</reuseForks>`) seems to fix the test failures and allows us to update to version 4.x of the parent POM, see #192.

nfalco79:
I see that in master the build continue to fails this means test are not rock solid

PR #196, opened 2020-05-11
Use System ACL to get size

res0nance:
Because the size is within the display name this can lead to poor performing situations especially with repositories with a large number of branches / pull requests. This performs even more poorly in those repositories because the tabs have to be sorted

```
Handling GET /job/org/job/repo from 10.33.250.197 : qtp1738561265-3033572 View/index.jelly DefaultViewsTabBar/viewTabs.jelly
java.lang.Character.toLowerCase(Character.java:6374)
java.lang.Character.toLowerCase(Character.java:6345)
java.lang.String$CaseInsensitiveComparator.compare(String.java:1203)
java.lang.String$CaseInsensitiveComparator.compare(String.java:1186)
java.lang.String.compareToIgnoreCase(String.java:1239)
hudson.util.CaseInsensitiveComparator.compare(CaseInsensitiveComparator.java:40)
hudson.util.CaseInsensitiveComparator.compare(CaseInsensitiveComparator.java:34)
jenkins.model.IdStrategy$CaseInsensitive.compare(IdStrategy.java:225)
jenkins.model.IdStrategy.equals(IdStrategy.java:120)
nectar.plugins.rbac.groups.Group.isMatch(Group.java:492)
nectar.plugins.rbac.groups.Group.isMatch(Group.java:445)
nectar.plugins.rbac.groups.Group.isMatch(Group.java:382)
nectar.plugins.rbac.groups.GroupContainerACL.hasPermission(GroupContainerACL.java:196)
nectar.plugins.rbac.groups.GroupContainerACL._hasPermission(GroupContainerACL.java:123)
nectar.plugins.rbac.groups.GroupContainerACL.hasPermission(GroupContainerACL.java:76)
org.jenkinsci.plugins.workflow.multibranch.BranchJobProperty$1.hasPermission(BranchJobProperty.java:77)
hudson.security.ACL.hasPermission(ACL.java:87)
hudson.security.AccessControlled.hasPermission(AccessControlled.java:54)
com.cloudbees.hudson.plugins.folder.AbstractFolder.getItems(AbstractFolder.java:1014)
com.cloudbees.hudson.plugins.folder.AbstractFolder.getItems(AbstractFolder.java:1004)
hudson.model.ListView.getItems(ListView.java:214)
hudson.model.ListView.getItems(ListView.java:184)
jenkins.branch.MultiBranchProjectViewHolder$ViewImpl.getDisplayName(MultiBranchProjectViewHolder.java:236)
hudson.views.ViewsTabBar$1.compare(ViewsTabBar.java:91)
hudson.views.ViewsTabBar$1.compare(ViewsTabBar.java:88)
java.util.TimSort.countRunAndMakeAscending(TimSort.java:355)
java.util.TimSort.sort(TimSort.java:220)
java.util.Arrays.sort(Arrays.java:1512)
java.util.ArrayList.sort(ArrayList.java:1462)
hudson.views.ViewsTabBar.sort(ViewsTabBar.java:88)
```

Since the number is just a count it might make sense to skip these costly permissions checks

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProjectViewHolder.java
+++ b/src/main/java/jenkins/branch/MultiBranchProjectViewHolder.java
@@ -34,6 +34,7 @@ import hudson.model.View;
 import hudson.model.ViewDescriptor;
 import hudson.model.ViewGroup;
 import hudson.security.ACL;
+import hudson.security.ACLContext;
 import hudson.security.Permission;
 import hudson.util.DescribableList;
 import hudson.views.DefaultViewsTabBar;
@@ -233,7 +234,10 @@ public class MultiBranchProjectViewHolder extends AbstractFolderViewHolder {
          */
         @Override
         public String getDisplayName() {
-            return category.getDisplayName().toString() + " (" + getItems().size() + ")";
+            //Only used to get a count and to prevent poor performance in cases with large number of items
+            try (ACLContext ctx = ACL.as(ACL.SYSTEM)) {
+                return category.getDisplayName().toString() + " (" + getItems().size() + ")";
+            }
         }
 
         /**
--- a/src/main/java/jenkins/branch/OrganizationFolderViewHolder.java
+++ b/src/main/java/jenkins/branch/OrganizationFolderViewHolder.java
@@ -33,6 +33,7 @@ import hudson.model.View;
 import hudson.model.ViewDescriptor;
 import hudson.model.ViewGroup;
 import hudson.security.ACL;
+import hudson.security.ACLContext;
 import hudson.security.Permission;
 import hudson.views.DefaultViewsTabBar;
 import hudson.views.StatusColumn;
@@ -226,7 +227,10 @@ public class OrganizationFolderViewHolder extends AbstractFolderViewHolder {
          */
         @Override
         public String getDisplayName() {
-            return category.getDisplayName().toString() + " (" + getItems().size() + ")";
+            //Only used to get a count and to prevent poor performance in cases with large number of items
+            try (ACLContext ctx = ACL.as(ACL.SYSTEM)) {
+                return category.getDisplayName().toString() + " (" + getItems().size() + ")";
+            }
         }
 
         /**


src/main/java/jenkins/branch/MultiBranchProjectViewHolder.java
```diff
@@ -233,7 +234,10 @@ public ViewImpl(ViewGroup owner, @NonNull SCMHeadCategory category) {
          */
         @Override
         public String getDisplayName() {
-            return category.getDisplayName().toString() + " (" + getItems().size() + ")";
+            //Only used to get a count and to prevent poor performance in cases with large number of items
+            try (ACLContext ctx = ACL.as(ACL.SYSTEM)) {
+                return category.getDisplayName().toString() + " (" + getItems().size() + ")";
```
dwnusbaum, line 239: I guess the drawback is that if someone doesn't have permission to view all of the items, the number here will be wrong? I wonder how common it is for users to configure different permissions for different jobs in a multibranch project or different multibranch projects in an organization folder. My guess is that it's not super common, so I'm not too worried about it, although I do wonder if some users would consider that a problem in terms of information disclosure if some of the jobs are supposed to be hidden from non-admins or similar.

res0nance: Yes the number would be wrong, I'm not sure what scenarios would read permissions be different especially for a github repo(private or public) since I don't think there is a way to give per branch permissions in github.

Currently I'm not sure of any way around this except maybe editing core where the number is handled in jelly. But this doesn't help the scenario when the views are sorted.

https://github.com/jenkinsci/jenkins/blob/34f5ea55cb77b83da9dc0529f597f2948d425ae4/core/src/main/resources/hudson/views/DefaultViewsTabBar/viewTabs.jelly#L29-L31

src/main/java/jenkins/branch/OrganizationFolderViewHolder.java
```diff
@@ -226,7 +227,10 @@ public ViewImpl(ViewGroup owner, @NonNull SCMSourceCategory category) {
          */
         @Override
         public String getDisplayName() {
-            return category.getDisplayName().toString() + " (" + getItems().size() + ")";
+            //Only used to get a count and to prevent poor performance in cases with large number of items
```
bitwiseman, line 230: How poor is the performance?

res0nance: In the most pathological case(1k Branches with Pull Requests), in can take 15 seconds or so.

This is only for users with an authorization strategy

src/main/java/jenkins/branch/MultiBranchProjectViewHolder.java
bitwiseman: @res0nance 
It's not just that the number would be wrong as Devin said - it gives the user information they shouldn't have.

bitwiseman:
@res0nance 
I'd be okay if with this being added as a setting that can be enabled by the folder owner if they choose it. That setting could default to "Do not show count of children".   😄  

It might also make sense to add a `getItemCount()`, so we don't load and return the list of all items just to get the size of the list. 



I know it outside of scope of this change, but counts like this would be better lazy loaded. So instead of just "DisplayName", tabs would have a "DisplayName" and "ChildCount".

res0nance:
> @res0nance
> I'd be okay if with this being added as a setting that can be enabled by the folder owner if they choose it. That setting could default to "Do not show count of children". 😄
> 
> It might also make sense to add a `getItemCount()`, so we don't load and return the list of all items just to get the size of the list.
> 
> I know it outside of scope of this change, but counts like this would be better lazy loaded. So instead of just "DisplayName", tabs would have a "DisplayName" and "ChildCount", where "ChildCount" would be populated after several seconds.

Honestly lazy loading would definitely be the best way, but I have no idea how to implement that as my jelly-fu is not so strong. 😆

PR #197, opened 2020-05-11
Move to hamcrest assertThat

res0nance:
In Junit 4.13 assertThat has been deprecated in favour of the hamcrest flavour swap the tests to use the hamcrest version


dwnusbaum:
Trying a rebuild, I think the test failures are infra-related.

PR #199, opened 2020-06-30
Fix PCT failures

chriskilding:
Fix the PCT failures unearthed in https://github.com/jenkinsci/bom/pull/179

To do:

- [x] Update test code
- [ ] Test the fixes with PCT locally

PR #200, opened 2020-07-08
[JENKINS-61956] Adapt test to stricter enforcement of createFreeStyleProject arg

jglick:
Compatibility with @calvinpark’s https://github.com/jenkinsci/jenkins/pull/4684. Noticed in #199.

### DIFF
--- a/src/test/java/jenkins/branch/WorkspaceLocatorImplTest.java
+++ b/src/test/java/jenkins/branch/WorkspaceLocatorImplTest.java
@@ -30,6 +30,7 @@ import hudson.scm.NullSCM;
 import hudson.slaves.DumbSlave;
 import hudson.slaves.WorkspaceList;
 import java.io.File;
+import java.io.IOException;
 import java.lang.reflect.Field;
 import java.util.Collections;
 import java.util.logging.Level;
@@ -226,7 +227,11 @@ public class WorkspaceLocatorImplTest {
         s = (DumbSlave) r.jenkins.getNode("remote");
         s.toComputer().connect(true).get();
         assertEquals(Collections.emptyList(), s.getWorkspaceRoot().listDirectories());
-        s.toComputer().getLogText().writeLogTo(0, System.out);
+        try {
+            s.toComputer().getLogText().writeLogTo(0, System.out);
+        } catch (IOException x) { // observed in Windows CI: FileNotFoundException: …\logs\slaves\remote\slave.log
+            x.printStackTrace();
+        }
     }
 
     @Issue({"JENKINS-2111", "JENKINS-58177"})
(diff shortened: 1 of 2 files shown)


jglick:
BTW I checked `workflow-multibranch` and saw some failures, but not obviously related. In particular I was worried about https://github.com/jenkinsci/workflow-multibranch-plugin/blob/388dedec6917d2c4ea4d7c583f3ddb465e92db06/src/test/java/org/jenkinsci/plugins/workflow/multibranch/WorkflowBranchProjectFactoryTest.java#L56-L70 but this passed against 2.237. That is because `ComputedFolder`s go through a different mechanism to add children, so they are allowed to have children with for example `%` in the name even though `checkGoodName` would not tolerate that.

calvinpark:
In my testing of the `workflow-multibranch` for https://github.com/jenkinsci/jenkins/pull/4684, the "illegal characters" in `checkGoogName` were properly encoded therefore were not affected by the change https://github.com/jenkinsci/jenkins/pull/4684#issuecomment-619281176

For your remark on `ComputedFolder`, is that a false positive that should also run `checkGoodName`, or is it passing as it should?

jglick:
> the "illegal characters" in `checkGoogName` were properly encoded therefore were not affected by the change

I think you are misreading the situation. https://github.com/jenkinsci/jenkins/blob/0eca516e80de6ebd6183b3cca670d8f56c6f5500/core/src/main/java/jenkins/model/Jenkins.java#L4030 indeed rejects `%`, which for example would be in the `Item.name` arising from a branch containing a `/`, but it does not matter because `ComputedFolder` is using another mechanism to add generated children which bypasses your patch.

src/test/java/jenkins/branch/WorkspaceLocatorImplTest.java
```diff
@@ -233,8 +233,8 @@ public void deleteOffline() throws Exception {
     @Test
     public void uniquification() throws Exception {
         WorkspaceLocatorImpl.MODE = WorkspaceLocatorImpl.Mode.ENABLED;
-        assertEquals("a_b", r.buildAndAssertSuccess(r.createFreeStyleProject("a$b")).getWorkspace().getName());
-        assertEquals("a_b_2", r.buildAndAssertSuccess(r.createFreeStyleProject("a!b")).getWorkspace().getName());
+        assertEquals("a_b", r.buildAndAssertSuccess(r.createFreeStyleProject("a'b")).getWorkspace().getName());
+        assertEquals("a_b_2", r.buildAndAssertSuccess(r.createFreeStyleProject("a\"b")).getWorkspace().getName());
```
jglick, line 237: Hmm, Windows does not like this…

PR #45, opened 2016-06-08
[JENKINS-34246] Fix MultiBranchProjectFactory API to allow deletion or update of existing children

jglick:
[JENKINS-34246](https://issues.jenkins-ci.org/browse/JENKINS-34246)

@reviewbybees

PR #203, opened 2020-07-09
OrganizationFolderTest.deletedMarker failed in PCT with megawar incl. workflow-multibranch

jglick:
Test from #45 is fragile.

```
java.lang.AssertionError: expected:<0> but was:<1>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.junit.Assert.assertEquals(Assert.java:633)
	at jenkins.branch.OrganizationFolderTest.deletedMarker(OrganizationFolderTest.java:142)
```

due to `WorkflowMultiBranchProjectFactory` (i.e., **from Jenkinsfile**) spuriously recognizing the test “repository”.

### DIFF
--- a/src/test/java/jenkins/branch/OrganizationFolderTest.java
+++ b/src/test/java/jenkins/branch/OrganizationFolderTest.java
@@ -34,6 +34,7 @@ import hudson.model.TaskListener;
 import hudson.model.View;
 import hudson.scm.NullSCM;
 import hudson.security.Permission;
+import integration.harness.BasicMultiBranchProjectFactory;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


src/test/java/jenkins/branch/OrganizationFolderTest.java
```diff
@@ -127,6 +134,9 @@ public void emptyViewEquality() throws Exception {
     @Issue("JENKINS-34246")
     @Test
     public void deletedMarker() throws Exception {
+        assumeThat("TODO fails if jth.jenkins-war.path includes WorkflowMultiBranchProjectFactory since SingleSCMSource ignores SCMSourceCriteria",
```
bitwiseman, line 137: TODO?  What will you be doing here?

jglick: Potentially rewriting the test to do something less brittle, like using `MockSCM` as in newer tests in this class.

PR #202, opened 2020-07-09
Flake in RateLimitBranchPropertyTest.rateLimitsBlockBuilds_maxRate

jglick:
Test from #58, flaking in https://ci.jenkins.io/job/Tools/job/bom/job/PR-179/29/testReport/junit/jenkins.branch/RateLimitBranchPropertyTest/pct_branch_api_2_176_x___rateLimitsBlockBuilds_maxRate/ :

```
java.lang.AssertionError: 
At least the rate implied delay but no more than 500ms longer
Expected: (a value equal to or greater than <3400L> and a value less than or equal to <4100L>)
     but: a value less than or equal to <4100L> <6282L> was greater than <4100L>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at jenkins.branch.RateLimitBranchPropertyTest.rateLimitsBlockBuilds(RateLimitBranchPropertyTest.java:201)
	at jenkins.branch.RateLimitBranchPropertyTest.rateLimitsBlockBuilds_maxRate(RateLimitBranchPropertyTest.java:132)
```

Seems inherently sensitive to timing on the server.

PR #211, opened 2020-09-18
RateLimitBranchPropertyTest.rateLimitsConcurrentBuilds is flaky too

jglick:
Follows up #202. Flake noticed in https://github.com/jenkinsci/bom/pull/301:

```
java.lang.AssertionError: 
At least the rate implied delay but no more than 500ms longer
Expected: (a value equal to or greater than <3400L> and a value less than or equal to <4100L>)
     but: a value equal to or greater than <3400L> <3342L> was less than <3400L>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at jenkins.branch.RateLimitBranchPropertyTest.rateLimitsConcurrentBuilds(RateLimitBranchPropertyTest.java:292)
```

### DIFF
--- a/src/test/java/jenkins/branch/RateLimitBranchPropertyTest.java
+++ b/src/test/java/jenkins/branch/RateLimitBranchPropertyTest.java
@@ -289,7 +289,7 @@ public class RateLimitBranchPropertyTest {
             // it can take more than the requested delay... that's ok, but it should not be
             // more than 500ms longer (i.e. 5 of our Queue.maintain loops above)
             final long delay = (long)(60.f * 60.f / rate * 1000);
-            assertThat("At least the rate implied delay but no more than 500ms longer",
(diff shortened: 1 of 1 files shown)


bitwiseman (approved):
We need to figure out how to reenable these rather than skipping when they fail.

jglick:
> how to reenable these

I guess. Looking at the test code, it is clearly inherently too flaky to run in CI.

bitwiseman:
@jglick 
> Looking at the test code, it is clearly inherently too flaky to run in CI.

But without these tests the feature could break completely and we would not know.

jglick:
> But without these tests the feature could break completely and we would not know.

Indeed. We can only hope that in the unlikely event someone actually touches this code again, that they will run these tests locally to validate it. In the meantime, the test is causing problems for automation and sucking up developer time that could be better spent elsewhere.

PR #214, opened 2020-10-02
onDeleted - CleanupTask causing Out Of Memory exception

kryptag:
The changes made are trying to solve an issue which occurs when there is a large amount of nodes that need their workspace deleted. The idea behind the solution is to have a Thread that works as a provisioner that only executes `ThreadPoolForRemoting.submit()`  if there is sufficient memory. Other users have experienced similar [JENKINS-27514](https://issues.jenkins-ci.org/browse/JENKINS-27514?jql=text%20~%20%22Threadpoolforremoting%22) & [JENKINS-47258](https://issues.jenkins-ci.org/browse/JENKINS-47258?jql=text%20~%20%22Threadpoolforremoting%22).
I do know that it is inherently a problem in Jenkins-Core, we have a Customer that are experiencing an issue when they use Branch-Api to clean up after a job.

So my plan was to alleviate the problem by targeting the specific issue they are experiencing, until the issues in Jenkins-Core can be resolved.

### DIFF
--- a/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
+++ b/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
@@ -441,12 +441,14 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
         }
         
         private static class CleanupTaskProvisioner implements Runnable{
-              
+            
+            @NonNull
             private final TopLevelItem tli;
             
+            @NonNull
             private final Queue<Node> nodes;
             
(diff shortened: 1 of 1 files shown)


kryptag:
Hey @bitwiseman I was unsure who's the maintainer, but you've merged recent pull requests, would you care to review this one?

kryptag:
After multiple tests these changes did not solve the current issue, as the issue is bound to the amount of thread and not memory specific. Issue is that the limit for maximum threads is reached.

bitwiseman:
@kryptag  Sorry I was on leave for a bit.

PR #220, opened 2020-10-14
Do not printStackTrace on an exception you are going to rethrow

jglick:
Noticed in https://github.com/jenkinsci/mercurial-plugin/pull/149 that the `AbortException` gets a stack trace printed (which is wrong—it is a user error) even though https://github.com/jenkinsci/cloudbees-folder-plugin/blob/3afb5acfeeca5a89cf97edc88e8d497b327667c3/src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderComputation.java#L171-L174 was going to print it politely at the end of the branch indexing log.

You should only print a stack trace if you are _not_ rethrowing an exception.

### DIFF
--- a/src/main/java/jenkins/branch/OrganizationFolder.java
+++ b/src/main/java/jenkins/branch/OrganizationFolder.java
@@ -514,17 +514,17 @@ public final class OrganizationFolder extends ComputedFolder<MultiBranchProject<
                     try {
                         bc.commit();
                     } catch (IOException | RuntimeException e) {
-                        printStackTrace(e, listener.error("[%tc] Could not persist folder level actions",
-                                System.currentTimeMillis()));
+                        listener.error("[%tc] Could not persist folder level actions",
+                                System.currentTimeMillis());
                         throw e;
                     }
                     if (saveProject) {
                         try {
                             save();
                         } catch (IOException | RuntimeException e) {
-                            printStackTrace(e, listener.error(
+                            listener.error(
                                     "[%tc] Could not persist folder level configuration changes",
-                                    System.currentTimeMillis()));
+                                    System.currentTimeMillis());
                             throw e;
                         }
                     }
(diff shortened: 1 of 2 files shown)


src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -1509,7 +1509,7 @@ public void complete() throws IllegalStateException, IOException, InterruptedExc
                     } catch (InterruptedException | IOException x) {
                         throw x;
                     } catch (Exception x) {
-                        x.printStackTrace(listener.error("Failed to create or update a subproject " + projectName));
+                        printStackTrace(x, listener.error("Failed to create or update a subproject " + projectName));
```
jglick, line 1512: Separate fix to use the Jenkins cause chaining.

bitwiseman (approved):

jglick:
> [@bitwiseman](/bitwiseman) [bitwiseman](/bitwiseman) reopened this [6 minutes ago](#event-4117239772)

BTW you can just click **Re-run** on the **Jenkins** check.

jtnord:
noting user based regressions when indexing is interrupted.  now there is no sign of what went wrong or where - the details of the `InteruptedException` are completely lost due to the special handling of the interuptedException in cloudbees-folders plugin

https://github.com/jenkinsci/cloudbees-folder-plugin/blob/07dd67fe36db681e7fcc3b73b117af3dc18f66b5/src/main/java/com/cloudbees/hudson/plugins/folder/computed/FolderComputation.java#L168-L171

jglick:
What kind of interesting stack trace were you expecting from an `InterruptedException`? Is this something that “went wrong”?

jtnord:
it does not even need a stack trace, just "it was interrupted" will do (because at the moment all you see is ABORTED and you have no idea why).  the stack is useful to know where if you do want more info, and that is available if you turn on FINEST logging, but first you would need to know - and given aborting is usually by a user, or step both of which are generally acompanied by a log to say why, this leaves the case of some for of InteruptedEx where you no longer have that log anymore and you are like 🤷  

Given [BitBucket branch source](https://github.com/jenkinsci/bitbucket-branch-source-plugin/blob/3817449486a0180a21eaa9c6b5a1342a6e7d18c0/src/test/java/integration/ScanningFailuresTest.java#L80) has an explicit test for this I can only assume that some part of the API Can throw this exception under some conditions (yes the test needs to be changed, but all you are left with in the log is aborted with no clue as to why)

jglick:
> just "it was interrupted" will do (because at the moment all you see is ABORTED and you have no idea why)

Well, `ABORTED` means something was interrupted. For `Run`s we have a `CauseOfInterruption` system which is not yet in `ComputedFolder` but could be added.

PR #225, opened 2020-10-28
remove unnecessary build trigger

geekchow:
Remove unnecessary build trigger. 
This a is a revert of PR https://github.com/jenkinsci/branch-api-plugin/pull/158
Actually, the PR-158 introduce a very serious issue. And lots of ppl already suffer from [it](https://github.com/jenkinsci/branch-api-plugin/pull/158#issuecomment-564523966). 

And logically, the code is not suitable. Because clause `Items.currentlyUpdatingByXml()` doesn't judge if the current item is being updated or not. It judges if any config changed by xml. 

Image this situation, I have thousands of multiplebranch jobs to seed with job-dsl. I change one job's  definition, then all the jobs get branch-indexing job scheduled simultaneously. That would overwhelm the github.

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -235,12 +235,6 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
                 }
             }
         }
-        if (Items.currentlyUpdatingByXml()) {
-            fireSCMSourceAfterSave(getSCMSources());
-            if (isBuildable()) {
-                scheduleBuild();
-            }
-        }
     }
 
     /**


MasonM (approved):

MasonM:
Thank for you entering this! We've also run into this issue.

FYI: There are test failures due to the `src/test/java/integration/UpdatingFromXmlTest.java` file added in https://github.com/jenkinsci/branch-api-plugin/pull/158. I think that will have to be removed. Probably should also remove `src/test/resources/integration/UpdatingFromXmlTest/config.xml`, since it won't be used anymore.

bitwiseman:
@geekchow @MasonM 

Simply reverting doesn't seem like good way to go here. 

#158 was merged over a year ago. It was quality PR that includes tests to verify that it is working. 

I understand that it causes problem for you, however simply reverting the change breaks others. 

If you want to add a feature flag to disable this code, that seems reasonable.  Even better though would be to provide a fix that preserves the improvement in #158 while fixing the way that it breaks your scenario.

geekchow:
> @geekchow @MasonM
> 
> Simply reverting doesn't seem like good way to go here.
> 
> #158 was merged over a year ago. It was quality PR that includes tests to verify that it is working.
> 
> I understand that it causes problem for you, however simply reverting the change breaks others.
> 
> If you want to add a feature flag to disable this code, that seems reasonable. Even better though would be to provide a fix that preserves the improvement in #158 while fixing the way that it breaks your scenario.

Hi @bitwiseman  thanks for you suggestion, let me try to add a feature flag for it.

michelzanini:
Where are we with this? It's years later and I still have to keep my branch-api plugin locked because this issue kills my Jenkins.
I understand someone introduced that before @bitwiseman , but, it was not a good solution, it causes more issues than it's worth.
Possibly the best would be to remove this code by default and only added it via feature flag.

But anyway, if the feature flag exists either way I would be very happy to get this released so I can finally use Jenkins again.

michelzanini:
I implemented a feature flag in #249.
If #249 gets accepted, we can close this one.
@bitwiseman

bitwiseman:
@michelzanini  
I just scanned through the history of comments on the related Jira tickets.  I am sorry for the problems this has caused you and others and I think it is important to find a solution to this. There is a bug here and it should be fixed. 

I think this PR is not a viable fix for the issue and should be closed (regardless of #249 being accepted or not).  From what I can see, the fix merged in #158 is correct behavior that exposes bad behavior in another part of the system. I left this PR open to allow you to reuse this PR to implement a feature flag or other fix. Since you've opened a new PR, this one should be closed.

michelzanini:
@geekchow can you close this PR in favour of #249?
Thanks.

bitwiseman:
@michelzanini 
Hm, I just reread this part of the original description: 

> And logically, the code is not suitable. Because clause Items.currentlyUpdatingByXml() doesn't judge if the current item is being updated or not. It judges if any config changed by xml.

Maybe #158 isn't correct?  More reading needed.

bitwiseman:
We took #249 as a workaround. Closing this one.  There may be a further fix here, but it is not this.

PR #228, opened 2020-11-07
JENKINS-64069 Fix tables to div compat (minimum core 2.264 bump)

timja:
See https://issues.jenkins-ci.org/browse/JENKINS-64069?focusedCommentId=400785&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-400785

For some reason jelly variables aren't being resolved correctly and I can't figure out why.

It work in this plugin when you run with `hpi:run` but if you do `mvn package -P quick-build` and then install it to a different Jenkins it doesn't work (name element isn't passed through)

I tried a few different options, a special tag just for this along with a local taglib and neither worked.

The local taglib didn't even pick up the `divBasedFormLayout`.

This bumps the minimum core to 2.264 (probably needs resubmitting by someone with write access to get a green build or changing the Jenkinsfile in a separate PR first.

If someone can figure out why the values aren't being passed through then we could avoid this PR but given this is the largest tables to divs breakage I think it would be good to just get this sorted.

@dwnusbaum @fqueiruga @oleg-nenashev

PR #229, opened 2020-11-07
Test core versions based on pom

timja:
Preparation for https://github.com/jenkinsci/branch-api-plugin/pull/228

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,5 @@
 buildPlugin(configurations: [
         [ platform: "linux", jdk: "8", jenkins: null ],
-        [ platform: "linux", jdk: "11", jenkins: "2.249" ],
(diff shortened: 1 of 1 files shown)


dwnusbaum (approved):
Thanks! I filed #232 to get a CI build using the updated Jenkinsfile.

fqueiruga (commented):
I don't know what this does :D. Does this mean that the baseline is increased?

timja:
> I don't know what this does :D. Does this mean that the baseline is increased?

the plugin was configured to run CI against newer baselines, the problem is I'm updating the baseline to above these versions which makes CI break.

but because i'm not a 'trusted' contributor to this repository (write access), my changes are ignored so I can't make the build green in my other PR without this change

fqueiruga (approved):

PR #226, opened 2020-11-01
JENKINS-64069 Fix compat with core 2.264

timja:
cc @oleg-nenashev @fqueiruga @dwnusbaum 

for it to work fully it needs https://github.com/jenkinsci/cloudbees-folder-plugin/pull/165

Note: there's still other plugins (at least github-branch-source and workflow-multibranch) that need fixing that are causing JS errors but I couldn't see any actual problems.

PR #231, opened 2020-11-09
Fix appearance of branch source configuration on Jenkins 2.263 and older

dwnusbaum:
#226 accidentally dropped a `colspan` attribute, breaking the appearance of the `BranchSource` config page. We should fix that and release a fIx before we merge #228 otherwise users running old versions of core will have to update core for the issue to be fixed.

Before this PR:

Screen Shot 2020-11-09 at 16 41 04

After this PR:

Screen Shot 2020-11-09 at 16 49 28

### DIFF
--- a/src/main/resources/jenkins/branch/BranchSource/config.jelly
+++ b/src/main/resources/jenkins/branch/BranchSource/config.jelly
@@ -24,7 +24,7 @@
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:branch="/lib/branch-api">
   <branch:rowWrapper name="source">
-    <branch:entryWrapper>
+    <branch:entryWrapper colspan="3">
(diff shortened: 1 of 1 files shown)


timja (approved):

src/main/resources/jenkins/branch/BranchSource/config.jelly
```diff
@@ -24,7 +24,7 @@
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:branch="/lib/branch-api">
   <branch:rowWrapper name="source">
-    <branch:entryWrapper>
+    <branch:entryWrapper colspan="3">
```
jglick, line 27: Should this not be a part of `entryWrapper.jelly`?

dwnusbaum: The `entryWrapper` tag defines `colspan` as an optional attribute because the tag is used in a few places, and not all of them want `colspan="3"`.

car-roll (approved):

fqueiruga (approved):

PR #235, opened 2020-11-15
Use predicate form of getItems() provided by folders plugin

res0nance:
Use the predicate form to avoid checking ACLs when possible


bitwiseman (approved):
This looks very sensible.

PR #236, opened 2020-11-15
Add is primary environment to branch name contributor

cwholmes:
https://issues.jenkins.io/browse/JENKINS-64686
'
During a pipeline execution, it is often valuable to know whether or not the branch currently being executed is the primary branch for the scm source.

This adds an environment contribution in the branch name contributor of `BRANCH_IS_PRIMARY`.

When the `PrimaryInstanceMetadataAction` action exists for the branch this environment variable is set to "true" otherwise it is set to "false".


AnEmortalKid:
@bitwiseman mind taking a peek at this

src/main/resources/jenkins/branch/BranchNameContributor/buildEnv.properties
```diff
@@ -21,6 +21,7 @@
 # THE SOFTWARE.
 
 blurb.BRANCH_NAME=For a multibranch project, this will be set to the name of the branch being built, for example in case you wish to deploy to production from <code>master</code> but not from feature branches; if corresponding to some kind of change request, the name is generally arbitrary (refer to <code>CHANGE_ID</code> and <code>CHANGE_TARGET</code>).
+blurb.BRANCH_IS_PRIMARY=For a multibranch project, this will be set to <code>true</code> if the branch being built is the primary branch of the SCM source; otherwise, it is set to <code>false</code>.
```
bitwiseman, line 24:
```suggestion
blurb.BRANCH_IS_PRIMARY=For a multibranch project, if the SCM source reports that the branch being built is a primary branch, this will be set to <code>true</code>, otherwise this will be set to <code>false</code>. Some SCM sources may report more than one branch as a primary branch while others may not supply this information. 
```

bitwiseman (approved):
This basically looks good to me. 
Is there a Jenkins JIRA ticket for this?  If not, please create one.

cwholmes:
> This basically looks good to me.
> Is there a Jenkins JIRA ticket for this? If not, please create one.

I don't believe a JIRA does exist, and I am not sure that I can create one as I don't have an account and don't seem to be able to create one at the moment.

AnEmortalKid:
I created https://issues.jenkins.io/browse/JENKINS-64686 for this change

cwholmes:
@bitwiseman I have made the suggested changes, but the build is now failing on linux for the same reason as master currently. Could you take another look and possibly get this change merged?

timja:
> @bitwiseman I have made the suggested changes, but the build is now failing on linux for the same reason as master currently. Could you take another look and possibly get this change merged?

will be fixed by: https://github.com/jenkinsci/branch-api-plugin/pull/242

bitwiseman:
Thanks, @timja !

src/main/java/jenkins/branch/BranchNameContributor.java
```diff
@@ -61,6 +62,11 @@ public void buildEnvironmentFor(Job j, EnvVars envs, TaskListener listener) thro
                 // Note: not using Branch.name, since in the future that could be something different
                 // than SCMHead.name, which is what we really want here.
                 envs.put("BRANCH_NAME", head.getName());
+                if (branch.getAction(PrimaryInstanceMetadataAction.class) != null) {
+                    envs.put("BRANCH_IS_PRIMARY", "true");
+                } else {
+                    envs.put("BRANCH_IS_PRIMARY", "false");
```
usmonster, line 68: Late review/question—instead of setting it to `"false"` (which is a truthy string), would it be possible to set it to the empty string?

```suggestion
                    envs.put("BRANCH_IS_PRIMARY", "");
```

This would make it easier for scripts to check this value without having to remember to convert it to a Boolean or comparing string literals. Please let me know if this makes sense, and I can open a PR.

(Hopefully it's not too late/disruptive to make this change since it looks like this hasn't been released yet. 🤞 )

bitwiseman: @usmonster 
If we haven't released since this was merged, I'd still be open to changing the behavior but I'm not convinced we should.  I generally think these variables being used by any number of systems, so designing specifically for shell script behavior isn't my goal.  `false` is much clearer in my java-centric mind.  Perhaps you could file a PR with the proposed change  some examples and tag @cwholmes and we can discuss in a thread that isn't off the tail end of a merged PR.  😄

usmonster: Thanks for your quick response, @bitwiseman!

To be clear, this isn't a shell script-specific proposal, but I was thinking more about the most common context in which these variables will be used, i.e. in the Groovy scripts that define Jenkins pipelines. Maybe the variable shouldn't even be set if it's not a primary branch? In any case, I'll file a new PR to formally propose and discuss this. Thanks again!

timja: my 2 cents is you should only be comparing for the string == "true".

imo current behaviour is fine. but 🤷

usmonster: 👉  #246

PR #237, opened 2020-11-24
Add the ability to limit the number of cleanup threads

kryptag:
[First PR](https://github.com/jenkinsci/branch-api-plugin/pull/214) The first PR i made was intended to resolve the issue which was caused by 'onDeleted'. After having the changes run that the issue wasn't memory related but rather the OOM exception was raised after the agents disconnected. 

So this is the second draft trying to resolve this issue, so far this has been running for more than 2 weeks without any issues.
The idea is to let the users set a limit for how many threads can run in Jenkins, and if that cap is reached the 'onDeleted' will wait with spawning new threads until the limit isn't reached. 

This might be a crude way to implement this check, it might be better to try at set the limit on only the plugin instead of Jenkins instance.

I'm hoping this might be useful as is and in the very least, spark a discussion about how we might resolve this.

### DIFF
--- a/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
+++ b/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
@@ -383,6 +383,8 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
     @Extension
     public static class Deleter extends ItemListener {
 
+        private static final int CLEANUP_THREAD_LIMIT = parseToInt(System.getProperty(Deleter.class.getName() + ".CLEANUP_THREAD_LIMIT", "0"), 0);
+
         @Override
         public void onDeleted(Item item) {
             if (!(item instanceof TopLevelItem)) {
@@ -392,7 +394,7 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
             Jenkins jenkins = Jenkins.get();
             Computer.threadPoolForRemoting.submit(new CleanupTask(tli, jenkins));
             // Starts provisioner Thread which is tasked with starting cleanup Threads
-            new CleanupTaskProvisioner(tli, jenkins.getNodes()).run();
+            new CleanupTaskProvisioner(tli, jenkins.getNodes(), CLEANUP_THREAD_LIMIT).run();
         }
 
         @Override
@@ -409,6 +411,14 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
 
         /** Number of {@link CleanupTask} which have been scheduled but not yet completed. */
         private static int runningTasks;
+
+        private static int parseToInt(String input, int defaultValue) {
+            try {
+                return Integer.parseInt(input);
+            } catch(NumberFormatException ex) {
+                return defaultValue;
+            }
+        }
 
         // Visible for testing
         static synchronized void waitForTasksToFinish() throws InterruptedException {
@@ -434,13 +444,14 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
             @NonNull
             private final Queue<Node> nodes;
 
-            private final int threadLimit = getThreadLimit();
+            private final int threadLimit;
 
-            public CleanupTaskProvisioner(TopLevelItem tli, List<Node> nodes) {
+            public CleanupTaskProvisioner(TopLevelItem tli, List<Node> nodes, int threadLimit) {
                 this.tli = tli;
                 this.nodes = new LinkedList<>(nodes);
+                this.threadLimit = threadLimit;
                 if (threadLimit > 0) {
-                    LOGGER.log(Level.INFO, "ThreadLimit Loaded from Environment : {0}", threadLimit);
+                    LOGGER.log(Level.INFO, "Cleanup thread limit set: {0}", threadLimit);
                 }
             }
 
@@ -467,20 +478,6 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
                 int currentThreadCount = ManagementFactory.getThreadMXBean().getThreadCount();
 
                 return currentThreadCount < threadLimit;
-            }
-
-            private int getThreadLimit() {
-                String branchApiThreadLimit = System.getenv("BRANCH_API_THREAD_LIMIT");
-                if (branchApiThreadLimit == null) {
-                    return 0;
-                }
-
-                try {
-                    return Integer.parseInt(branchApiThreadLimit);
-                } catch (NumberFormatException e) {
-                    LOGGER.log(Level.SEVERE, "Failed to parse BRANCH_API_THREAD_LIMIT: {0}", branchApiThreadLimit);
-                    return 0;
-                }
             }
         }
 
--- a/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
+++ b/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
@@ -387,7 +387,11 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
 
         private static /* almost final */ int CLEANUP_THREAD_LIMIT = SystemProperties.getInteger(Deleter.class.getName() + ".CLEANUP_THREAD_LIMIT", Integer.valueOf(0)).intValue();
 
+        /** Semaphore for limiting number of scheduled {@link CleanupTask} */
         private static Semaphore cleanupPool = new Semaphore(CLEANUP_THREAD_LIMIT, true);
+
+        /** Number of {@link CleanupTask} which have been scheduled but not yet completed. */
+        private static int runningTasks;
 
         @Override
         public void onDeleted(Item item) {
@@ -396,9 +400,9 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
             }
             TopLevelItem tli = (TopLevelItem) item;
             Jenkins jenkins = Jenkins.get();
-            Computer.threadPoolForRemoting.submit(new CleanupTask(tli, jenkins, cleanupPool));
+            Computer.threadPoolForRemoting.submit(new CleanupTask(tli, jenkins));
             // Starts provisioner Thread which is tasked with starting cleanup Threads
-            new CleanupTaskProvisioner(tli, jenkins.getNodes(), cleanupPool).run();
+            new CleanupTaskProvisioner(tli, jenkins.getNodes()).run();
         }
 
         @Override
@@ -413,8 +417,19 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
             }
         }
 
-        /** Number of {@link CleanupTask} which have been scheduled but not yet completed. */
-        private static int runningTasks;
+        static void acquireThread() throws InterruptedException {
+            if (CLEANUP_THREAD_LIMIT <= 0) {
+                return;
+            }
+            cleanupPool.acquire();
+        }
+
+        static void releaseThread() {
+            if (CLEANUP_THREAD_LIMIT <= 0) {
+                return;
+            }
+            cleanupPool.release();
+        }
 
         // Visible for testing
         static synchronized void waitForTasksToFinish() throws InterruptedException {
@@ -440,21 +455,17 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
             @NonNull
             private final Queue<Node> nodes;
 
-            @NonNull
-            private final Semaphore cleanupPool;
-
-            public CleanupTaskProvisioner(TopLevelItem tli, List<Node> nodes, Semaphore cleanupPool) {
+            public CleanupTaskProvisioner(TopLevelItem tli, List<Node> nodes) {
                 this.tli = tli;
                 this.nodes = new LinkedList<>(nodes);
-                this.cleanupPool = cleanupPool;
             }
 
             @Override
             public void run() {
                 try {
                     while (!nodes.isEmpty()){
-                        cleanupPool.acquire();
-                        Computer.threadPoolForRemoting.submit(new CleanupTask(tli, nodes.remove(), cleanupPool));
+                        Deleter.acquireThread();
+                        Computer.threadPoolForRemoting.submit(new CleanupTask(tli, nodes.remove()));
                     }
                 } catch (Exception e) {
                     LOGGER.log(Level.WARNING, e.getMessage());
@@ -470,13 +481,9 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
             @NonNull
             private final Node node;
 
-            @NonNull
-            private final Semaphore cleanupPool;
-
-            CleanupTask(TopLevelItem tli, Node node, Semaphore cleanupPool) {
+            CleanupTask(TopLevelItem tli, Node node) {
                 this.tli = tli;
                 this.node = node;
-                this.cleanupPool = cleanupPool;
                 taskStarted();
             }
 
@@ -519,7 +526,7 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
                     }
                 } finally {
                     t.setName(oldName);
-                    cleanupPool.release();
+                    Deleter.releaseThread();
                     taskFinished();
                 }
             }
--- a/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
+++ b/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
@@ -434,7 +434,7 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
             @NonNull
             private final Queue<Node> nodes;
 
-            private final int threadLimit = Math.max(0, Integer.parseInt(System.getenv("BRANCH_API_THREAD_LIMIT")));
+            private final int threadLimit = getThreadLimit();
 
             public CleanupTaskProvisioner(TopLevelItem tli, List<Node> nodes) {
                 this.tli = tli;
@@ -469,6 +469,19 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
                 return currentThreadCount < threadLimit;
             }
 
+            private int getThreadLimit() {
+                String branchApiThreadLimit = System.getenv("BRANCH_API_THREAD_LIMIT");
+                if (branchApiThreadLimit == null) {
+                    return 0;
+                }
+
+                try {
+                    return Integer.parseInt(branchApiThreadLimit);
+                } catch (NumberFormatException e) {
+                    LOGGER.log(Level.SEVERE, "Failed to parse BRANCH_API_THREAD_LIMIT: {0}", branchApiThreadLimit);
+                    return 0;
+                }
+            }
         }
 
         private static class CleanupTask implements Runnable {
--- a/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
+++ b/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
@@ -50,12 +50,15 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
+import java.lang.management.ManagementFactory;
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


kryptag:
@dwnusbaum saw you merged the last couple of commits and i was wondering how to get a maintainer to review this Pull Request. Hopefully me pinging you isn't too much of a hassle.

bitwiseman:
This looks reasonable, it is safe in that it won't effect those who do not use it. 
However, is there a way to make it so the administrator doesn't need to set this limit? Some progressive back off?  Hiding functionality behind feature flags  and environment settings is fine but is not very discoverable.

kryptag:
@bitwiseman 
Both progressive back off and finding a way to calculate threads based on the CPU, sounds like a good idea. I was just wondering if that could be implemented in further improvements. As this implementation will already be useful for users that are facing similar issues. By giving more users access to this feature, it could give us more feedback on how it runs and what we could improve.

bitwiseman (approved):

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -438,6 +446,46 @@ private static synchronized void taskFinished() {
...
+            
+            @NonNull
+            private final TopLevelItem tli;
+            
+            @NonNull
+            private final Queue<Node> nodes;
+            
+            public CleanupTaskProvisioner(TopLevelItem tli, List<Node> nodes) {
+                this.tli = tli;
+                this.nodes = new LinkedList<>(nodes);
```
bitwiseman, line 460:
```suggestion
                this.nodes = new LinkedList<>(nodes);
                if (threadLimit > 0) {
                    LOGGER.log(Level.INFO, "ThreadLimit Loaded from Environment : {0}", threadLimit);
                }
```

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -438,6 +446,46 @@ private static synchronized void taskFinished() {
...
         }
+        
+        private static class CleanupTaskProvisioner implements Runnable{
+            
+            @NonNull
+            private final TopLevelItem tli;
+            
+            @NonNull
+            private final Queue<Node> nodes;
+
```
bitwiseman, line 457:
```suggestion
                        @NonNull
                        private final int threadLimit = Math.max(0, Integer.parseInt(System.getenv("BRANCH_API_THREAD_LIMIT", 0)));
```

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -438,6 +446,46 @@ private static synchronized void taskFinished() {
...
+                    boolean isRunning = true;
+                    while(isRunning){                   
+                        // If thread limit is reached break While Loop - Check if Queue is empty if not - retry
+                        while(hasFreeThreadVolume() && !nodes.isEmpty()){
+                            Computer.threadPoolForRemoting.submit(new CleanupTask(tli, nodes.poll()));
+                            LOGGER.log(Level.INFO, "LIVE THREAD COUNT : {0}", ManagementFactory.getThreadMXBean().getThreadCount());
+                        }
+                    // If the Queue is empty it will set isRunning to False which will terminated the Loop
+                    isRunning = !nodes.isEmpty();
+                    }
```
bitwiseman, line 475:
```suggestion
                    while (!nodes.isEmpty()){                   
                        if (hasFreeThreadVolume()) {
                            Computer.threadPoolForRemoting.submit(new CleanupTask(tli, nodes.remove()));
                        } else {
                            Thread.sleep(1000);
                        }
                        LOGGER.log(Level.INFO, "LIVE THREAD COUNT : {0}", ManagementFactory.getThreadMXBean().getThreadCount());
                    }
```

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -438,6 +446,46 @@ private static synchronized void taskFinished() {
...
+                    }   
+                } catch (Exception e) {
+                    LOGGER.log(Level.WARNING, e.getMessage());
+                }
+            }
+        }
+        
+        private static boolean hasFreeThreadVolume(){
+            int threadLimit = Integer.parseInt(System.getenv("BRANCH_API_THREAD_LIMIT"));
+            LOGGER.log(Level.INFO, "ThreadLimit Loaded from Environment : {0}", threadLimit);
```
bitwiseman, line 484:
```suggestion
        private boolean hasFreeThreadVolume(){
            if (threadLimit <= 0) {
                return true;
            }
```

bitwiseman (changes requested):
Suggestions to clean up code. 

Otherwise, seems reasonable to me.

kryptag:
@bitwiseman I might have messed something up, I haven't used this view before. Is there a way to untangle the mess I have made?

bitwiseman:
@kryptag 
Fetch the changes to your local machine and edit there?  I'm not sure I understand you question.

kryptag:
@bitwiseman Will do, it was just because I used the Github Ui to do the merging and it seemed to cause some issues.

kryptag:
@bitwiseman Think I have gotten your changes in, feel free to give it a look when possible.

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -403,8 +406,8 @@ public void onDeleted(Item item) {
             TopLevelItem tli = (TopLevelItem) item;
             Jenkins jenkins = Jenkins.get();
             Computer.threadPoolForRemoting.submit(new CleanupTask(tli, jenkins));
-            for (Node node : jenkins.getNodes()) {
-                Computer.threadPoolForRemoting.submit(new CleanupTask(tli, node));
+            // Starts provisioner Thread which is tasked with starting cleanup Threads
+            new CleanupTaskProvisioner(tli, jenkins.getNodes()).run();
             }
```
bitwiseman, line 411:
```suggestion
            new CleanupTaskProvisioner(tli, jenkins.getNodes()).run();
```

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -438,6 +441,61 @@ private static synchronized void taskFinished() {
...
+                }
+                if (threadLimit > 0) {
+                    LOGGER.log(Level.INFO, "ThreadLimit Loaded from Environment : {0}", threadLimit);
+                }
+                if (threadLimit > 0) {
+                    LOGGER.log(Level.INFO, "ThreadLimit Loaded from Environment : {0}", threadLimit);
+                }
+                if (threadLimit > 0) {
+                    LOGGER.log(Level.INFO, "ThreadLimit Loaded from Environment : {0}", threadLimit);
+                }
```
bitwiseman, line 471:
```suggestion
```

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -438,6 +441,61 @@ private static synchronized void taskFinished() {
...
+        private static class CleanupTaskProvisioner implements Runnable{
+            
+            @NonNull
+            private final TopLevelItem tli;
+            
+            @NonNull
+            private final Queue<Node> nodes;
+                        @NonNull
+                        private final int threadLimit = Math.max(0, Integer.parseInt(System.getenv("BRANCH_API_THREAD_LIMIT", 0)));
+            public CleanupTaskProvisioner(TopLevelItem tli, List<Node> nodes) {
```
bitwiseman, line 454:
```suggestion
            private final Queue<Node> nodes;
            
            @NonNull
            private final int threadLimit = Math.max(0, Integer.parseInt(System.getenv("BRANCH_API_THREAD_LIMIT", 0)));

            public CleanupTaskProvisioner(TopLevelItem tli, List<Node> nodes) {
```

bitwiseman (dismissed):
Please fix the failures and I'll review again.

bitwiseman:
@kryptag 
Have you built this locally?

kryptag:
@bitwiseman i'm getting errors regarding the 'waitForTasksToFinish()' method. Maven can't find the method which fails the build. This haven't been a problem earlier. Do you have any ideas on how i can resolve this?

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
dwnusbaum: @kryptag The indentation in the file is all messed up now because you removed the opening brace of the loop but kept the closing one. That's why you had to remove `@Override` from some methods and probably why the build is failing. You need to apply @bitwiseman's suggestion and remove the extra closing brace.

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -403,12 +406,11 @@ public void onDeleted(Item item) {
...
             Jenkins jenkins = Jenkins.get();
             Computer.threadPoolForRemoting.submit(new CleanupTask(tli, jenkins));
-            for (Node node : jenkins.getNodes()) {
-                Computer.threadPoolForRemoting.submit(new CleanupTask(tli, node));
+            // Starts provisioner Thread which is tasked with starting cleanup Threads
+            new CleanupTaskProvisioner(tli, jenkins.getNodes()).run();
             }
         }
 
-        @Override
```
dwnusbaum, line 411: Once you fix the indentation issues, this should be added back.

dwnusbaum (dismissed):

kryptag:
@dwnusbaum Thanks for the help, I think I messed up a lot when I used the Github gui too resolve the code changes. I have fixed the issues and the test now runs as it should. I just started running the full test suite and will push it's successful. 

Sorry for any inconvenience I might have caused.

kryptag:
I find it weird that I can't recreate those fails when I run the tests locally.

bitwiseman:
I was able to reproduce the test failure locally.

praqma-thi:
I've rebased and tidied up Kryptag's changes locally and managed to reproduce the failing test.
Before I dig in to fix it, I wanted to check if there's still interest in merging this in?
If so, I'm happy to help get this across the finish line.

volvo-ritte:
This PR has solved a major problem we had in our production-CI (that we use at Volvo GTT). We have a lot of build agents and the current implementation creates too many processes. 

It would be great if this PR can be merged so we don't need to maintain a modified version of the plugin.

sigossi:
> This PR has solved a major problem we had in our production-CI (that we use at Volvo Trucks). We have a lot of build agents and the current implementation creates too many processes.
> 
> It would be great if this PR can be reviewed and merged so we don't need to maintain a modified version of the plugin. @bitwiseman @dwnusbaum

Agree to this. Hoping for this patch to be merged into the official release

WalrusArtist:
Also from Volvo Trucks here. Looking forward to this merge soon.

chasip:
looks useful! what's the status?

sigossi:
We have been running with a custom plugin version based on this PR for a few months now and it has solved the server crashes for us.

During the summer we were happy to discover that this PR solved our repeated server crashes with tens of thousands of thread spawned by the Branch API cleanup handling, we configured a thread cap at 15000 threads to make the server survive, and it has been running successfully for months now:
``` 
2022-07-13 19:44:13.347+0000 [id=10631] INFO    j.b.WorkspaceLocatorImpl$Deleter$CleanupTaskProvisioner#run: LIVE THREADDD COUNT : 14,996
2022-07-13 19:44:13.350+0000 [id=9374]  INFO    j.b.WorkspaceLocatorImpl$Deleter$CleanupTaskProvisioner#run: LIVE THREADDD COUNT : 14,997
2022-07-13 19:44:13.352+0000 [id=9374]  INFO    j.b.WorkspaceLocatorImpl$Deleter$CleanupTaskProvisioner#run: LIVE THREADDD COUNT : 14,998
2022-07-13 19:44:13.354+0000 [id=9374]  INFO    j.b.WorkspaceLocatorImpl$Deleter$CleanupTaskProvisioner#run: LIVE THREADDD COUNT : 14,999
2022-07-13 19:44:13.356+0000 [id=25737] INFO    j.b.WorkspaceLocatorImpl$Deleter$CleanupTaskProvisioner#run: LIVE THREADDD COUNT : 15,000
2022-07-13 19:44:13.356+0000 [id=25737] WARNING j.b.WorkspaceLocatorImpl$Deleter$CleanupTaskProvisioner#run: Failed to deploy new CleanupTask due to reaching threadLimit : 15,000
``` 

Recently though we accidentally pulled in the official release of the Branch API plugin due to a dependency from another plugin and a few minutes later the cleanup task started spawning tens of thousands of threads causing the server to crash again :/
We are heavily dependent on this branch, and are hoping for it to be approved and merged in a near future. Is this something that is likely to happen? @car-roll @timja @jglick

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -444,6 +446,64 @@ private static synchronized void taskFinished() {
...
+                if (threadLimit <= 0) {
+                    return true;
+                }
+                int currentThreadCount = ManagementFactory.getThreadMXBean().getThreadCount();
+
+                return currentThreadCount < threadLimit;
+            }
+
+            private int getThreadLimit() {
+                String branchApiThreadLimit = System.getenv("BRANCH_API_THREAD_LIMIT");
```
car-roll, line 493: Can we put this at the top of the class with the other system properties and do something like:
```suggestion
                static final int = Integer.getInteger(WorkspaceLocatorImpl.class.getName() + "BRANCH_API_THREAD_LIMIT", 0);
```

praqma-thi: Hey @car-roll, happy to make more adjustments where necessary.
Tidying up the thread limit property is a good idea, but it's used in `CleanupTaskProvisioner`, an internal class of `Deleter`, which is a static class. I'm not sure how to combine that with the suggestion.

Do I move the property to `Deleter` and pass it into `CleanupTaskProvisioner`'s constructor (suggestion below), or is there some other way of achieving what you're after?

```java
    @Extension
    public static class Deleter extends ItemListener {

        private static final String CLEANUP_THREAD_LIMIT = System.getProperty(Deleter.class.getName() + "CLEANUP_THREAD_LIMIT", 0);
        
        @Override
        public void onDeleted(Item item) {
            // <omitted code>

            // Starts provisioner Thread which is tasked with starting cleanup Threads
            new CleanupTaskProvisioner(tli, jenkins.getNodes(), CLEANUP_THREAD_LIMIT).run();
        }
```

car-roll: That looks good 👍

praqma-thi: Hey @car-roll, I just pushed the changes to replace the environment variable with a system property.

I jammed a small function into `Deleter` to help parse it to an integer. There's a similar function in `org.apache.commons.lang3.NumberUtils`, but I didn't quite want to fiddle with the project's dependencies. Any opinions there?

sigossi:
Great progress here! Thanks for the effort guys! 
I am very much hoping to get this PR merged soon. Are we only one review away from getting this into master? @car-roll @bitwiseman @timja @dwnusbaum

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -438,6 +450,51 @@ private static synchronized void taskFinished() {
...
+                } catch (Exception e) {
+                    LOGGER.log(Level.WARNING, e.getMessage());
+                }
+            }
+
+            private boolean hasFreeThreadVolume(int threadLimit){
+                if (threadLimit <= 0) {
+                    return true;
+                }
+                int currentThreadCount = ManagementFactory.getThreadMXBean().getThreadCount();
```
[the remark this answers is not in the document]
sigossi: These are good improvements. But can it be pushed into an improvement work package? Just to get out of the current situation first?

I don't think most users will set this System property. And the ones who are experiencing the repeated server crashes like us, just want it somewhere within the max nproc limit of the user running Jenkins. It is not important for us that cleanup is done through the branch api plugin specifically since we manage that ourselves anyway (and we have many other jobs which aren't using the branch-api cleanup feature to keep disk space usage low)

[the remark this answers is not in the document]
praqma-thi: Hey James, thanks a lot for the feedback. It definitely seems reasonable. I tidied up the property parsing, and gave using the Semaphore a shot.

Originally, I ended up passing the Semaphore down to the `CleanupTaskProvider`/`CleanupTask`. I ended up having to check if there even was a limit before trying to `acquire`/`release` in multiple places, which made it all a bit messy. I believe it would work if we had a (large?) limit by default, then no checks would've been required.

Ultimately, I opted for adding two static methods to `Deleter`. Not sure if I like it, but it does make it easier to use.
Thoughts?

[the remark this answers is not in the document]
sigossi: Thanks @praqma-thi 
@jtnord what are your thoughts on the current state of this PR?

jtnord (approved):
As the Deleter is an Extension you should always be able to get it via `ExtensionList.lookupSingleton()` to avoid the requirement for a `static` of the code - but seems reasonable to me.

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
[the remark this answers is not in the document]
jtnord: Hi @praqma-thi 

> Ultimately, I opted for adding two static methods to Deleter. Not sure if I like it, but it does make it easier to use.
Thoughts?

the Listener is a singleton `Extension` so you can obtain it at runtime with `ExtensionList.lookupSingleton()` to avoid having to have something static or pass the code around.

if you want to try that I will hold merging for a short while. otherwise let me know to merge release

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -438,6 +460,32 @@ private static synchronized void taskFinished() {
...
+            }
+
+            @Override
+            public void run() {
+                try {
+                    while (!nodes.isEmpty()){
+                        Deleter.acquireThread();
+                        Computer.threadPoolForRemoting.submit(new CleanupTask(tli, nodes.remove()));
+                    }
+                } catch (Exception e) {
```
jtnord, line 483: NIT:  if the exception is either a [`RejectedExecutionException`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/RejectedExecutionException.html) or an `InteruptedException` should we continue to process any of the nodes?

a rejectedExecution would seem to imply the execturor service was shut down so all future attempts to submit something would fail.
Likewise an `InteruptedException` somone wanted to interupt us whilst we where waiting to aquire a lock - normally that would be "please stop what you are doing gracefully"

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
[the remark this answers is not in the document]
praqma-thi: Hey James, happy to give that a shot when I'm back from my holiday, plus tidying up the exception handling mentioned below. It'll be next year for me, but I'll gladly dig into it regardless of whether it gets merged now or not.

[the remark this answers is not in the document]
praqma-thi: Hi James, I finally got out of hibernation. I rebased onto master and replaced the static methods with getting a reference using `lookupSingleton()`. I'm quite happy with the changes, if there's more improvements, feel free to let me know.

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
praqma-thi: Hey @jtnord, I took a quick peek at this. As it is, wouldn't any exception terminate the loop? Or do you wish to only terminate the loop on either the `RejectedExecutionException`/`InterruptedException` and continue when encountering others?

jtnord: Hi @praqma-thi

Sorry about that when first reviewing it looked like I thought the `catch` was inside the while.

jtnord (approved):

PR #244, opened 2021-03-19
[JENKINS-63673] Allow configuring suppressed events in NoTrigger properties

agabrys:
The branch property was by default suppressing all triggers: indexing and events (e.g. webhooks). The indexing mechanism is problematic for cloud environments when Jenkins configuration is not kept on a persistent volume. Every Jenkins creation schedules builds of all branches and pull requests which may block the instance for many hours or even days. This change introduces a configuration option which allows administrators/developers to decide if both or only one trigger's kind should be suppressed. It is useful for the mentioned cloud environments because the indexing mechanism could be disabled without ignoring webhooks.

How it looks:
view

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!

Link to the backlog item: [JENKINS-63673](https://issues.jenkins.io/browse/JENKINS-63673).


agabrys:
The feature is ready for review. I added unit tests and I also installed it on Jenkins and checked how it works. I created a repository with 2 branches and next used Job DSL to create the job with this configuration in the `multibranchPipelineJob` closure:
```groovy
branchSources {
    branchSource {
        source {
            github {
               // ...
            }
        }
        strategy {
           allBranchesSame {
                props {
                    suppressAutomaticTriggering  {
                        strategyId(2)
                    }
                }
            }
        }
    }
}
```
Jenkins didn't schedule automatic builds for the both branches. Then I created a new branch and webhook was delivered. The build for the new branch has been started. Next, I disabled the webhook configuration in GitHub and created again a new branch. After clicking the `Scan Repository Now` button the branch appeared, but the build has not been started. All looks fine for me 🙂

Some comments from my side:
1. The unit tests are quite long (~190 seconds; previously ~50 seconds; longer: ~2 minutes 20 seconds). I checked how the other tests are implemented and all of them are using JenkinsRule etc, so I did the same. We may safe some time by testing the `NoTriggerBranchProperty.Dispatcher#shouldSchedule` method by using Mockito to fake the input parameters (no need for the starting Jenkins instance)
2. The order of the operations in the `NoTriggerBranchProperty.Dispatcher#shouldSchedule` method could be improved. For example there is no sense to iterate on every action if the `p` parameter doesn't keep a job which parent is a multi branch project. I didn't change it because I wanted to keep the diff as small as possible, but maybe there is a sense to change it 🙂
3. I didn't add tests for the `NoTriggerBranchProperty.DescriptorImpl.doFillStrategyIdItems` method, because I didn't see similar tests in other classes. Let me know if I should add tests for it. Generally, it is a simple method which returns available options

In case of any changes or improvement ideas - let me know 🙂

agabrys:
I added `@since 2.6.4` in JavaDocs.

jonesbusy:
Excellent work @agabrys ! Hope it will get released soon

agabrys:
I removed changes from the `CHANGELOG.md` file. This file should not be modified since 2.5.4.

agabrys:
@bitwiseman I apologize for the direct ping, but I don't know where to ask. Is there any chance to check this PR in the near future? Maybe there are a regular cycles when the community PRs are checked?

agabrys:
I rebased the PR branch with the `master` branch.

agabrys:
And again rebase with `master`.

agabrys:
My friend told me that labels are a little bit confusing, so I changed options of the `Suppression strategy`:
* `All` → `Suppress all`
* `Triggered by Indexing` → `Suppress only triggered by indexing`
* `Triggered by Events` → `Suppress only triggered by events`

src/main/resources/jenkins/branch/Messages.properties
```diff
@@ -28,6 +28,9 @@ DescriptionColumn.displayName=Project description
 ItemColumn.DisplayName=Name
 NamedExceptionsBranchPropertyStrategy.DisplayName=Named branches get different properties
 NoTriggerBranchProperty.suppress_automatic_scm_triggering=Suppress automatic SCM triggering
+NoTriggerBranchProperty.strategy.all=Suppress all
+NoTriggerBranchProperty.strategy.indexing=Suppress only triggered by indexing
+NoTriggerBranchProperty.strategy.events=Suppress only triggered by events
```
bitwiseman, line 33:
```suggestion
NoTriggerBranchProperty.strategy.all=Suppress all, runs will not be trigger by indexing or events
NoTriggerBranchProperty.strategy.indexing=Suppress runs triggered by indexing
NoTriggerBranchProperty.strategy.events=Suppress runs triggered by webhook events
```

bitwiseman (changes requested):
Great PR.  Consider using strings instead of numbers, what do you think?

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
```diff
@@ -45,9 +48,45 @@
 @Restricted(NoExternalUse.class)
 public class NoTriggerBranchProperty extends BranchProperty {
 
+    /**
+     * All builds triggered by SCM are suppressed.
+     *
+     * @since 2.6.4
+     */
+    public static final int ALL = 1;
```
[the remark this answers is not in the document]
agabrys: No problem. I changed integers to strings:
* `1` → `all`
* `2` → `indexing`
* `3` → `events`

src/main/resources/jenkins/branch/Messages.properties
agabrys: I added the `webhook` word to the `NoTriggerBranchProperty.strategy.all` property value too, to keep docs consistent.

```
NoTriggerBranchProperty.strategy.all=Suppress all, runs will not be trigger by indexing or webhook events
NoTriggerBranchProperty.strategy.indexing=Suppress runs triggered by indexing
NoTriggerBranchProperty.strategy.events=Suppress runs triggered by webhook events
```

agabrys:
I applied the suggestions. Let me know what do you think 🙂 

When all is fine, I'll rebase the branch with the `master` branch and squash commits (keep history clean).

agabrys:
I see the 2.6.4 version has been released today, so I updated the `@since` tags too.

ineffyble:
Would love to see this merged @bitwiseman

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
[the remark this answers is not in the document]
agabrys: I was thinking, maybe these names are better:
```
// 1
suppressAutomaticTriggering {
    strategyId('all')
}
```
```
// 2
suppressAutomaticTriggering {
    strategyId('suppressIndexing')
}
```
```
// 3
suppressAutomaticTriggering {
    strategyId('suppressEvents')
}
```

GMouron:
This is great work, would be super useful for us, thanks @agabrys

marcel-dias:
Also looking into this. Can I do something to help this PR get merged?

lechuk47:
@bitwiseman Any chances to approve the PR? We are looking into this too.

wernla:
Great work. This change is super helpful for us. Is there anything we can do, to get this merged?

agabrys:
I rebased branch with `master` and squashed all commits.

wernla:
@bitwiseman : can you have a look into this?

src/main/resources/jenkins/branch/NoTriggerBranchProperty/config.jelly
```diff
@@ -24,4 +24,8 @@ THE SOFTWARE.
 -->
 
 <?jelly escape-by-default='true'?>
-<j:jelly xmlns:j="jelly:core"/>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+  <f:entry title="${%Suppression strategy}" field="strategyId">
+    <f:select default="1"/>
```
agabrys, line 29: I think it should be changed to
```suggestion
    <f:select default="all"/>
```

I'll verify it tomorrow. If I am right, I'll update the PR.

agabrys: Yes, it was a mistake and I fixed it 🙂

src/test/java/jenkins/branch/NoTriggerBranchPropertyTest.java
```diff
@@ -30,21 +30,33 @@
...
+import jenkins.scm.impl.mock.MockSCMSource;
 import org.junit.Rule;
+import org.junit.Test;
 import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.JenkinsRule;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
```
DanSt, line 58: These belong at the beginning of the file I believe. Not just by Java conventions but also when looking at other similar files, this convention seems to hold true.

src/main/resources/jenkins/branch/Messages.properties
```diff
@@ -28,6 +28,9 @@ DescriptionColumn.displayName=Project description
 ItemColumn.DisplayName=Name
 NamedExceptionsBranchPropertyStrategy.DisplayName=Named branches get different properties
 NoTriggerBranchProperty.suppress_automatic_scm_triggering=Suppress automatic SCM triggering
+NoTriggerBranchProperty.strategy.all=Suppress all, runs will not be trigger by indexing or webhook events
```
DanSt, line 31:
```suggestion
NoTriggerBranchProperty.strategy.all=Suppress all, runs will not be triggered by indexing or webhook events
```

agabrys: I applied the suggestion and squashed the commits 🙂

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
```diff
@@ -45,9 +48,45 @@
...
+     */
+    public static final String INDEXING = "indexing";
+    /**
+     * Only builds triggered by {@link BranchEventCause} are suppressed.
+     *
+     * @since 2.6.6
+     */
+    public static final String EVENTS = "events";
+
+    private String strategyId = ALL;
```
jglick, line 70: Better to use an `enum`.

agabrys: I replaced `static` fields by `enum`. Let me know if it is fine - if yes, I squash the commits.

agabrys: The build has failed in the Windows branch with error:
```
Cannot contact EC2 (aws) - Windows 2019 (i-02c6b5fa1020239a4): hudson.remoting.ChannelClosedException: Channel "hudson.remoting.Channel@67050906:EC2 (aws) - Windows 2019 (i-02c6b5fa1020239a4)": Remote call on EC2 (aws) - Windows 2019 (i-02c6b5fa1020239a4) failed. The channel is closing down or has closed down
```

agabrys:
I rebased the source branch with `master`. I haven't squashed the commits yet, because I'm waiting for the decision if the new implementation with `enum` is better 🙂

agabrys:
I squashed the commits hoping that a stray wanderer would get here and press the `approve` button and next merge it 😊

fmarot:
I'd love to see it merged. Build storms are a problem when creating new job referencing old repositories.

agabrys:
Rebased with `master`.

jtnord:
Would it be possible to add a migration test to show that any existing jobs configured with the property are migrated correctly with a newer version?

agabrys:
The default behaviour is tested by the [smokes](https://github.com/agabrys/branch-api-plugin/blob/feature/JENKINS-63673/src/test/java/jenkins/branch/NoTriggerBranchPropertyTest.java#L70-L121) method. In the 87. line ([see the code](https://github.com/agabrys/branch-api-plugin/blob/feature/JENKINS-63673/src/test/java/jenkins/branch/NoTriggerBranchPropertyTest.java#L87)) an instance of `NoTriggerBranchProperty` is created an no additional configuration is set.

Could you elaborate more how the migration tests should look like?

jglick:
https://javadoc.jenkins.io/component/jenkins-test-harness/org/jvnet/hudson/test/recipes/LocalData.html

agabrys:
@jtnord I added the test as you requested (`suppressAllWhenJobCreatedBeforeAddingTheSuppressionStrategyParameter`). I didn't use the `LocalData` annotation because the test has to replace two placeholders in the job XML configuration.

I made two additional changes:
* guarantee that the `getStrategy` method never returns `null`
* remove the `suppressNothing` test

I think there is no sense to add a property to skip triggers and next configure it to never skip anything.

I haven't squash the commits, so the review process should be easier (analyze only the new diff). When you accept the changes, I'll squash the commits, so the git history is clean 🙂

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
```diff
@@ -83,7 +108,12 @@ public boolean shouldSchedule(Queue.Task p, List<Action> actions) {
                                     } else {
                                         for (BranchProperty prop : ((MultiBranchProject) j.getParent()).getProjectFactory().getBranch(j).getProperties()) {
                                             if (prop instanceof NoTriggerBranchProperty) {
-                                                return false;
+                                                NoTriggerBranchProperty triggerProperty = (NoTriggerBranchProperty) prop;
+                                                boolean suppressIndexing = SuppressionStrategy.EVENTS != triggerProperty.getStrategy();
+                                                boolean suppressEvents = SuppressionStrategy.INDEXING != triggerProperty.getStrategy();
```
jtnord, line 113: NIT: at first glance this logic is a little hard to parse and looks like it can be optimized but it can't

```suggestion
                                                boolean suppressIndexing = SuppressionStrategy.EVENTS != triggerProperty.getStrategy(); // INDEXING or ALL
                                                boolean suppressEvents = SuppressionStrategy.INDEXING != triggerProperty.getStrategy(); // EVENTS or ALL
```

or add an abstract  `BranchCause` between `Cause` and the `Branch*Cause`s (or a marker interface) and move the logic to the enum with a `shouldSupress(BranchCause)`

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
```diff
@@ -95,7 +125,40 @@ public boolean shouldSchedule(Queue.Task p, List<Action> actions) {
...
 
+    /**
+     * Strategy which determines which builds should be suppressed.
+     * @since 2.7.1
+     */
+    public enum SuppressionStrategy {
+        /**
+         * All builds triggered by SCM are suppressed.
+         *
+         * @since 2.7.1
```
jtnord, line 138: NIT: same comment as above - or just remove this as it is the same as the enum type.

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
```diff
@@ -95,7 +125,40 @@ public boolean shouldSchedule(Queue.Task p, List<Action> actions) {
...
+         */
+        ALL(Messages._NoTriggerBranchProperty_strategy_all()),
+        /**
+         * Only builds triggered by {@link BranchIndexingCause} are suppressed.
+         *
+         * @since 2.7.1
```
jtnord, line 144: NIT: same comment as above - or just remove this as it is the same as the enum type.

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
```diff
@@ -95,7 +125,40 @@ public boolean shouldSchedule(Queue.Task p, List<Action> actions) {
...
+         */
+        INDEXING(Messages._NoTriggerBranchProperty_strategy_indexing()),
+        /**
+         * Only builds triggered by {@link BranchEventCause} are suppressed.
+         *
+         * @since 2.7.1
```
jtnord, line 150: NIT: same comment as above - or just remove this as it is the same as the enum type.

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
```diff
@@ -95,7 +125,40 @@ public boolean shouldSchedule(Queue.Task p, List<Action> actions) {
             }
             return true;
         }
-
     }
 
+    /**
+     * Strategy which determines which builds should be suppressed.
+     * @since 2.7.1
```
jtnord, line 132: NIT: probably as this is not a bug fix.
```suggestion
     * @since 2.8.0
```

jtnord (approved):
Thanks @agabrys this looks good!

src/main/resources/jenkins/branch/Messages.properties
```diff
@@ -28,6 +28,9 @@ DescriptionColumn.displayName=Project description
 ItemColumn.DisplayName=Name
 NamedExceptionsBranchPropertyStrategy.DisplayName=Named branches get different properties
 NoTriggerBranchProperty.suppress_automatic_scm_triggering=Suppress automatic SCM triggering
+NoTriggerBranchProperty.strategy.all=Suppress all, runs will not be triggered by indexing or webhook events
+NoTriggerBranchProperty.strategy.indexing=Suppress runs triggered by indexing
+NoTriggerBranchProperty.strategy.events=Suppress runs triggered by webhook events
```
jglick, line 33: I find the negative sense here somewhat hard to follow. Usually we prefer to use positive sense in UIs, e.g.
```suggestion
NoTriggerBranchProperty.strategy.all=Suppress builds triggered by indexing or webhook events (only start explicitly)
NoTriggerBranchProperty.strategy.indexing=Suppress builds triggered by indexing (continue to honor webhooks)
NoTriggerBranchProperty.strategy.events=Suppress builds triggered by webhooks (continue to trigger from indexing)
```

Similarly for Groovy in the `properties` step; do you have an updated example of the post-`enum` syntax to review?

jglick:
I think you also need to edit `NoTriggerOrganizationFolderProperty`, or the new feature will not be available for organization folders, only single-repository multibranch folders.

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
jglick: Or just skip the `@since` entirely. Only marginally helpful in Jenkins core, where you might be making a conscious decision about which `jenkins.version` to use a baseline. When it comes to plugin APIs, generally you would just be depending on the newest plugin version available, or whatever is in `bom`. If that version contains an API, use it—no need to know when it was introduced. If it does not, you would not know about the API to begin with, right?

jglick (changes requested):
I think https://github.com/jenkinsci/branch-api-plugin/pull/244#issuecomment-1005097120 should be treated as blocking. Though it is technically an enhancement that could be done as a follow-up, we try to avoid situations where something can be configured in a multibranch project which cannot be configured in an organization folder.

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
jglick: No need to squash BTW. Up to the plugin maintainer where to true-merge or squash-merge a PR. Also best to avoid rebasing with `master` as it interferes with GH reviews; better to true-merge (e.g. `git pull upstream master:master`).

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
agabrys: Okay, I'll remove them.

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
agabrys: I would like to not add comments at the end of the line because code should be self-descriptive 🙂I added the `shouldSuppress(Cause)` method. I haven't added the marker interface yet. Instead I added a logic responsible for checking if a correct type was passed. Based on the logic it has no chance to be executed, however it is a public API, so it probably has sense. Let me know what do you think.

If you prefer, I can add the marker interface but then we cannot prevent situations when somebody add it to completely other type (the suppression logic could start behaving strange).

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
agabrys: I hope somebody squash it on merge because commits like
* add upgrade (migration) tests
* Update src/main/resources/jenkins/branch/Messages.properties

don't bring any value when the git history is analyzed.

agabrys:
From https://github.com/jenkinsci/branch-api-plugin/pull/244#issuecomment-1005097120:
> I think you also need to edit `NoTriggerOrganizationFolderProperty`, or the new feature will not be available for organization folders, only single-repository multibranch folders.

From https://github.com/jenkinsci/branch-api-plugin/pull/244#pullrequestreview-843931847:
> I think [#244 (comment)](https://github.com/jenkinsci/branch-api-plugin/pull/244#issuecomment-1005097120) should be treated as blocking.

🤔 I disagree that it is blocking this ticket. Why? These two classes are not in any relation. The `NoTriggerOrganizationFolderProperty` class provides an option to filter by branch name which is missing in the `NoTriggerBranchProperty` class. I even wasn't aware about the `NoTriggerOrganizationFolderProperty` class because I've never used it. If they should always support the same options, I think it should be forced by code (interfaces and class hierarchy). I would like not to copy-paste lines from `NoTriggerBranchProperty` to `NoTriggerOrganizationFolderProperty`. So the question is: should this refactor been done within this ticket?

jglick:
FYI somewhat related: https://github.com/jenkinsci/basic-branch-build-strategies-plugin/blob/master/docs/user.adoc

> These two classes are not in any relation.

They are. `NoTriggerBranchProperty` applies to a single `MultiBranchProject`. `NoTriggerOrganizationFolderProperty` provides the corresponding function when you use an `OrganizationFolder`, which would otherwise be impossible because its `MultiBranchProject` children cannot be configured (they are generated, just like actual `Job`s beneath a `MultiBranchProject` are generated).

Probably https://github.com/jenkinsci/branch-api-plugin/blob/b600837d9d8b07a296b805b71123b795a51aa518/src/main/java/jenkins/branch/NoTriggerOrganizationFolderProperty.java#L98-L108 and https://github.com/jenkinsci/branch-api-plugin/blob/b600837d9d8b07a296b805b71123b795a51aa518/src/main/java/jenkins/branch/NoTriggerBranchProperty.java#L84-L88 could be made two clauses of a single `Queue.QueueDecisionHandler` to avoid duplication, though I do not suppose it is critical here.

agabrys:
Okay, I'll try to implement it in a correct way. I don't want to copy-paste code.

agabrys:
Just to let you know, I started working on the refactor. I pushed it to other branch ([see changes](https://github.com/agabrys/branch-api-plugin/compare/feature/JENKINS-63673...feature/JENKINS-63673v2)). When it will be ready, I'll push changes to this PR.

agabrys:
I refactored the code. Now both classes provide the same properties and the logic is in one place. I also updated the documentation (JavaDocs and HTML files):

From commit message:
>The `NoTriggerBranchProperty` and `NoTriggerOrganizationFolderProperty` classes have to provide the same features. The logic was copy-pasted between the classes, so the code was duplicated and hard to maintain (unwritten contract). The logic has been externalized to a new class `NoTriggerMultiBranchQueueDecisionHandler`. It must be extended by the `Dispatcher` class. Methods required by the `Dispatcher` class are defined in the `NoTriggerProperty` class which has to be extended by the property classes.
>
>The `NoTriggerOrganizationFolderProperty` class provides the `branches` property. It determines for which branches the automatic builds should not be suppressed. It has been added to the `NoTriggerBranchProperty` class. To make the logic backward compatible, it by default matches no branches (`^$`).
>
>To implement the parametrized tests easily, the `JUnitParams` library has been added. It has API very similar to JUnit 5, so it will be very easy to migrate to it in the future.

Btw. I think this code is unneeded: https://github.com/jenkinsci/branch-api-plugin/blob/ac9234ee647afb58ad1cad65c145dc83f0cd6b7b/src/main/java/jenkins/branch/OverrideIndexTriggersJobProperty.java#L78-L107
The logic responsible for processing the `OverrideIndexTriggersJobProperty` property was in the `NoTriggerBranchProperty` and `NoTriggerOrganizationFolderProperty` classes. Not it is in the shared class. Let me know if you any reason why this code should not be deleted. Test results without that code:
```
[INFO] Running jenkins.branch.OverrideIndexTriggersJobPropertyTest
[INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 140.641 s - in jenkins.branch.OverrideIndexTriggersJobPropertyTest
```

carpnick:
FYI also helpful in the situation when migrating from one Git to another Git server.  Allows for branch indexing to run without a bunch of build calls.

hkuersten:
I, too, would love to see this merged. Great work @agabrys!

IrisPeres:
This is what we are looking at our organization. Thanks for your time and research spent on this issue. 
We'll highly appreciate it if this can go forward and be merged 🙏

AnirudhDagar:
Thanks for the PR, this is really really needed! I hope it is merged and released soon. 💯

hkuersten:
Awfully quiet here, could somebody please provide an update on this? I'd really like to start using this feature...

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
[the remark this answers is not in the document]
car-roll: marking as resolved since this was migrated to enum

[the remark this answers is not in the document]
car-roll: marking as resolved since this was migrated to enum

src/main/java/jenkins/branch/NoTriggerBranchProperty.java
car-roll: Marking as resolved. No worries, I prefer squash-merge

car-roll: Marking as resolved. No worries, I prefer squash-merge

car-roll:
@jglick It looks like your changes were addressed, specifically that `NoTriggerOrganizationFolderProperty` was updated to support the new behaviors. Do you aggree?
https://github.com/jenkinsci/branch-api-plugin/pull/244/files#diff-2cb91cd8e5ce7272a41b0eca142802cdd0dbb6d4c763566009a823cd1433128f

@agabrys Feel free to update your branch to main (again) so we can run the checks one last time.

car-roll (approved):

car-roll:
I can confirm that when you create an organization folder the suppression strategy options are present.

GMouron:
🎉

jonesbusy:
🎉🎉

hkuersten:
woohooo 🤙🏻

PR #246, opened 2021-04-09
Tweak semantics for BRANCH_IS_PRIMARY env var

usmonster:
Instead of setting "false" in cases where the branch isn't a primary
branch, don't set the variable at all.

The goal is to make using the value easier and less error-prone. Since
the string "false" is truthy in most contexts, e.g. in
Groovy/Jenkinsfile pipelines and related scripts, its use can lead to
unexpected behavior if evaluated in a condition without first performing
an explicit type conversion.

Additionally, it's already a common practice among most consumers of
such "Boolean" environment variables to treat a nonempty value as an
indication of truth. (cf. the `CI` environment variable)

Related: #236.

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!
- [X] Please describe what you did
- [X] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/jenkins/branch/BranchNameContributor.java
+++ b/src/main/java/jenkins/branch/BranchNameContributor.java
@@ -64,8 +64,6 @@ public class BranchNameContributor extends EnvironmentContributor {
                 envs.put("BRANCH_NAME", head.getName());
                 if (branch.getAction(PrimaryInstanceMetadataAction.class) != null) {
                     envs.put("BRANCH_IS_PRIMARY", "true");
-                } else {
-                    envs.put("BRANCH_IS_PRIMARY", "false");
                 }
                 if (head instanceof ChangeRequestSCMHead) {
                     envs.putIfNotNull("CHANGE_ID", ((ChangeRequestSCMHead) head).getId());
--- a/src/main/resources/jenkins/branch/BranchNameContributor/buildEnv.properties
+++ b/src/main/resources/jenkins/branch/BranchNameContributor/buildEnv.properties
@@ -21,7 +21,7 @@
 # THE SOFTWARE.
 
 blurb.BRANCH_NAME=For a multibranch project, this will be set to the name of the branch being built, for example in case you wish to deploy to production from <code>master</code> but not from feature branches; if corresponding to some kind of change request, the name is generally arbitrary (refer to <code>CHANGE_ID</code> and <code>CHANGE_TARGET</code>).
-blurb.BRANCH_IS_PRIMARY=For a multibranch project, if the SCM source reports that the branch being built is a primary branch, this will be set to <code>true</code>, otherwise this will be set to <code>false</code>. Some SCM sources may report more than one branch as a primary branch while others may not supply this information. 
+blurb.BRANCH_IS_PRIMARY=For a multibranch project, if the SCM source reports that the branch being built is a primary branch, this will be set to <code>"true"</code>; else unset. Some SCM sources may report more than one branch as a primary branch while others may not supply this information. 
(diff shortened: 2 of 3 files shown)


usmonster:
[As instructed](https://github.com/jenkinsci/branch-api-plugin/pull/236#discussion_r610196882), I'm tagging @cwholmes. 🙂

src/main/java/jenkins/branch/BranchNameContributor.java
```diff
@@ -64,8 +64,6 @@ public void buildEnvironmentFor(Job j, EnvVars envs, TaskListener listener) thro
                 envs.put("BRANCH_NAME", head.getName());
                 if (branch.getAction(PrimaryInstanceMetadataAction.class) != null) {
                     envs.put("BRANCH_IS_PRIMARY", "true");
-                } else {
-                    envs.put("BRANCH_IS_PRIMARY", "false");
```
usmonster, line 68: Alternatively, which was my initial suggestion:
```suggestion
                } else {
                    envs.put("BRANCH_IS_PRIMARY", "");
```

This would at least allow scripts to distinguish between the `false` case and one where an older version of the plugin is in use.

cwholmes: I don't inherently have an issue with this proposal. The truthy part of a "false" value can be confusing (as can be treating any value as true).

usmonster: If you prefer my suggestion in the above comment to the implementation currently proposed in the PR, I can update it.

Thanks for the comment!

(Sorry for the delay, it wasn't clear to me if others would want to chime in first.)

bitwiseman: What happens if one attempts to us this variable inside the pipeline in a string interpolation for example:
```
sh "BRANCH_IS_PRIMARY = ${env.BRANCH_IS_PRIMARY}"
sh "BRANCH_IS_PRIMARY = $BRANCH_IS_PRIMARY"
```
Won't the result in an error for the false case if we take this change?

usmonster: You would get an error unless you change it to what I think you mean:

```groovy
sh "BRANCH_IS_PRIMARY=${env.BRANCH_IS_PRIMARY}"
sh "BRANCH_IS_PRIMARY=$BRANCH_IS_PRIMARY"
```

With this, neither the empty string nor the `"false"` case will result in an error, unless I'm missing something?

usmonster: Hello! I was just checking whether the PR is OK to merge as-is, or if you'd rather set the empty string as the alternative suggested in this comment, and I just realized I may have misunderstood @bitwiseman's question. Indeed, if someone tries to use a variable that isn't set, there will be an error. I wonder if it's an issue since there is the same possibility with [other `...; else unset` variables](https://github.com/jenkinsci/branch-api-plugin/blob/master/src/main/resources/jenkins/branch/BranchNameContributor/buildEnv.properties).

This would mean that the variable should only be checked via `env.BRANCH_IS_PRIMARY` and not `BRANCH_IS_PRIMARY` (unless not using string interpolation). An alternative to not setting it in the false case would be to set it to the empty string, as proposed here.

bitwiseman: > Indeed, if someone tries to use a variable that isn't set, there will be an error.

bitwiseman (approved):

bitwiseman:
@cwholmes 
So, do you approve of this change?

bitwiseman:
@cwholmes I missed that you indicated approval earlier.  Taking this change.

PR #249, opened 2021-04-22
[JENKINS-63344] Add a feature flag to be able to disable running SCM branch source scanning when job XMLs are updated

michelzanini:
This pull request adds a feature flag to allow users to disable a very annoying and serious issue when Jenkins contains many jobs and uses Bitbucket or Github with Job DSL plugin.

Be disabling this setting it will avoid causing hundreds of unnecessary calls and then get to the API rate limit of Bitbucket or Github.

This PR will replace the need for #225.
It also closes this issue on Jenkins https://issues.jenkins.io/browse/JENKINS-63344?attachmentViewMode=list.

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -34,6 +34,7 @@ import com.cloudbees.hudson.plugins.folder.views.AbstractFolderViewHolder;
 import com.thoughtworks.xstream.XStreamException;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.BulkChange;
 import hudson.Extension;
 import hudson.Util;
@@ -113,6 +114,7 @@ import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.StaplerResponse;
 import org.kohsuke.stapler.export.Exported;
+import jenkins.util.SystemProperties;
 
 import static hudson.Functions.printStackTrace;
 
@@ -125,6 +127,10 @@ import static hudson.Functions.printStackTrace;
 public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
         R extends Run<P, R>>
         extends ComputedFolder<P> implements SCMSourceOwner, IconSpec {
+
+    @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "Accessible via System Groovy Scripts")
+    private static /* not final */ boolean FIRE_SCM_SOURCE_BUILDS_AFTER_SAVE =
+        SystemProperties.getBoolean(MultiBranchProject.class.getName() + ".fireSCMSourceBuildsAfterSave", true);
 
     /**
      * Our logger.
@@ -233,7 +239,7 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
                 }
             }
         }
-        if (Items.currentlyUpdatingByXml()) {
+        if (Items.currentlyUpdatingByXml() && FIRE_SCM_SOURCE_BUILDS_AFTER_SAVE) {
             fireSCMSourceAfterSave(getSCMSources());
             if (isBuildable()) {
                 scheduleBuild();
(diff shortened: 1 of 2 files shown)


michelzanini:
I tested this on a live Jenkins and it's working nicely 👍

bitwiseman:
@michelzanini 
Excellent! Thanks for opening this new PR.  
I'll repeat what I said in #225:

> I just scanned through the history of comments on the related Jira tickets. I am sorry for the problems this has caused you and others and I think it is important to find a solution to this. There is a bug here and it should be fixed.

Also from #225:
> If you want to add a feature flag to disable this code, that seems reasonable. Even better though would be to provide a fix that preserves the improvement in #158 while fixing the way that it breaks your scenario.

So, this feature flag is a fine workaround.  You will need to add a test that verifies this behavior.

Once we have a test, I am open to taking this PR, but in the long term feature flags like this are brittle and prone to unexpected side-effects. 

In the Jira threads, I think you mentioned that this issue is happening due to a double update of the job XML by job-dsl when using views inside a job? Is that correct?  Could we fix that double update to be a single, or maybe add a setting to the job that stops the scan.  I think such a setting already exists - disabling the first scan on job creation.  It might be possible to extend that.

bitwiseman:
@michelzanini 
Also there appears to be an infrastructure issue right now, so we'll have to wait for CI checks.

michelzanini:
Hi @bitwiseman ,

> Once we have a test, I am open to taking this PR, but in the long term feature flags like this are brittle and prone to unexpected side-effects.

I was not able to add a test for it because the field is `private static` and I could not find a way of changing the System property from a test. Do you know any other thing we could do to get this unit tested? I could do it easily if the field wasn't private, but making it public would probably fail findbugs...

> In the Jira threads, I think you mentioned that this issue is happening due to a double update of the job XML by job-dsl when using views inside a job? Is that correct? Could we fix that double update to be a single, or maybe add a setting to the job that stops the scan. I think such a setting already exists - disabling the first scan on job creation. It might be possible to extend that.

The issue is bigger than I am able to understand fully in order to fix it. If you update a folder or a view, all jobs inside the view or folder will get triggered. Also, Job DSL seems to update job XMLs in some cases even if they have not changed. To properly fix this, it might be required to touch several plugins (Job DSL, branch api, folders, etc) and I don't have the source code knowledge to do this. Only a Jenkins maintainer would probably be able to do something like that.

I am open to get the feature flag in and tested, but regarding trying to stop triggers that should not happen would be outside of my capacity.

Let me know if you have an idea how to test when the flag is `false` (modify the system property from a test).
Thanks.

michelzanini:
I just found a way of testing this. Will be pushing the test in a moment.

bitwiseman:
> I was not able to add a test for it because the field is `private static` and I could not find a way of changing the System property from a test. Do you know any other thing we could do to get this unit tested? I could do it easily if the field wasn't private, but making it public would probably fail findbugs...

You can make it internal and then it can be set inside a test.  Or better you can also use a `FlagRule` which is part of the Jenkins test harness.

bitwiseman:
> To properly fix this, it might be required to touch several plugins (Job DSL, branch api, folders, etc) and I don't have the source code knowledge to do this. Only a Jenkins maintainer would probably be able to do something like that.

The way people gain that knowledge is by working on and fixing issues like this.  I can tell you honestly, I don't have that knowledge either right now.

michelzanini:
> You can make it internal and then it can be set inside a test. Or better you can also use a FlagRule which is part of the Jenkins test harness.
Tested added using the suggested FlagRule class, thanks for this.

This is as much as I can help, with this PR. I won't get enough time to dive deep / learn more about it.
If you can accept the feature flag, then that will be great as I would be able to unlock the Jenkins instances we have to use latest versions of the branch plugin and all other dependent plugins we have locked because of this.

If not, we will possibly need to fork and deploy a custom version of this plugin for some time at least, so it unblocks some plugin updates we require.

thecosmicfrog (approved):

thecosmicfrog:
This is a very relevant change to avoid hammering our Git provider's API (causing rate limit errors, etc.). It would be great to see it merged.

CyberJimbo:
I've just come across this PR from googling a similar issue - it would be helpful if this flag gets merged even for an interim period - as I am also currently at the mercy of the Bitbucket rate limit due to excessive scanning. Cheers.

car-roll (approved):
While this does not directly solve JENKINS-63344, I believe it is a valid workaround to relieve the current pain. It does look like `Items.currentlyUpdatingByXml` is the culprit, as mentioned in https://github.com/jenkinsci/branch-api-plugin/pull/225.

bitwiseman:
@car-roll @thecosmicfrog @michelzanini 
I'll get a release out today.

thecosmicfrog:
Thanks @bitwiseman - very much appreciated!

PR #258, opened 2021-07-06
Tone down warning about `rawWorkspaceDir` from `WorkspaceLocatorImpl`

jglick:
In the case of JFR https://github.com/jenkinsci/jenkinsfile-runner/blob/69c70e5c6ae070151cf2890f96ae60031aae9d23/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/commands/PipelineRunOptions.java#L16-L24 + https://github.com/jenkinsci/jenkinsfile-runner/blob/69c70e5c6ae070151cf2890f96ae60031aae9d23/bootstrap/src/main/java/io/jenkins/jenkinsfile/runner/bootstrap/commands/RunJenkinsfileCommand.java#L18-L41 this warning

```
WARNING	j.branch.WorkspaceLocatorImpl#getWorkspaceRoot: JENKINS-2111 path sanitization ineffective when using legacy Workspace Root Directory ‘/jenkins_build_data/build’; switch to ‘${JENKINS_HOME}/workspace/${ITEM_FULL_NAME}’ as in JENKINS-8446 / JENKINS-21942
```

is just noise.

### DIFF
--- a/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
+++ b/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
@@ -330,11 +330,16 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
     private static final Pattern GOOD_RAW_WORKSPACE_DIR = Pattern.compile("(.+)[/\\\\][$][{]ITEM_FULL_?NAME[}][/\\\\]?");
     static @CheckForNull FilePath getWorkspaceRoot(Node node) {
         if (node instanceof Jenkins) {
-            Matcher m = GOOD_RAW_WORKSPACE_DIR.matcher(((Jenkins) node).getRawWorkspaceDir());
+            String rawWorkspaceDir = ((Jenkins) node).getRawWorkspaceDir();
+            Matcher m = GOOD_RAW_WORKSPACE_DIR.matcher(rawWorkspaceDir);
             if (m.matches()) {
                 return new FilePath(new File(m.group(1).replace("${JENKINS_HOME}", ((Jenkins) node).getRootDir().getAbsolutePath())));
             } else {
-                LOGGER.log(Level.WARNING, "JENKINS-2111 path sanitization ineffective when using legacy Workspace Root Directory ‘{0}’; switch to ‘$'{'JENKINS_HOME'}'/workspace/$'{'ITEM_FULL_NAME'}'’ as in JENKINS-8446 / JENKINS-21942", ((Jenkins) node).getRawWorkspaceDir());
+                if (rawWorkspaceDir.startsWith("${ITEM_ROOTDIR}/")) {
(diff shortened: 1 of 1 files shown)


src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -330,11 +330,16 @@ private static Object lockFor(Node node) {
...
+            Matcher m = GOOD_RAW_WORKSPACE_DIR.matcher(rawWorkspaceDir);
             if (m.matches()) {
                 return new FilePath(new File(m.group(1).replace("${JENKINS_HOME}", ((Jenkins) node).getRootDir().getAbsolutePath())));
             } else {
-                LOGGER.log(Level.WARNING, "JENKINS-2111 path sanitization ineffective when using legacy Workspace Root Directory ‘{0}’; switch to ‘$'{'JENKINS_HOME'}'/workspace/$'{'ITEM_FULL_NAME'}'’ as in JENKINS-8446 / JENKINS-21942", ((Jenkins) node).getRawWorkspaceDir());
+                if (rawWorkspaceDir.startsWith("${ITEM_ROOTDIR}/")) {
+                    LOGGER.log(Level.WARNING, "JENKINS-2111 path sanitization ineffective when using legacy Workspace Root Directory ‘{0}’; switch to ‘$'{'JENKINS_HOME'}'/workspace/$'{'ITEM_FULL_NAME'}'’ as in JENKINS-8446 / JENKINS-21942", rawWorkspaceDir);
+                } else {
+                    LOGGER.log(Level.FINE, "JENKINS-2111 path sanitization ineffective when using unrecognized Workspace Root Directory ‘{0}’; switch to ‘$'{'JENKINS_HOME'}'/workspace/$'{'ITEM_FULL_NAME'}'’", rawWorkspaceDir);
```
dwnusbaum, line 341: IDK how useful this message was in the first place or whether this branch is a case we would ever expect non-JFR users to run into. Would it make sense to have JFR set a generic (`IsJenkinsfileRunner=true`) system property or something like that, and then we could check for that in these kinds of cases instead to avoid affecting non-JFR users? Or do we only expect this branch to be hit by JFR users?

dwnusbaum (approved):

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
jglick: Not sure. I just know that in case neither the old nor new default workspace dir pattern match, you are probably in some special setup and did not really care about the refined behavior of `WorkspaceLocatorImpl`. Especially on the controller node—none of this applies to agents.

jglick:
Could I get someone to merge & release this please?

car-roll (approved):

PR #271, opened 2021-09-01
Removing `CustomOrganizationFolderDescriptor`

jglick:
At some point someone thought it would be a good idea to have the **New Item** page show one entry per kind of `SCMNavigator` (GitHub, Bitbucket, etc.) even though each entry would actually create the same kind of thing, an `OrganizationFolder`, just with an `SCMNavigator` of that type initially added to the list. (Note that this was never an attempt at a real wizard, which would presumably prompt you for authentication and walk you through various steps and finally give you a fully ready-to-use folder.) The hack would also suppress all of the item types if the `workflow-multibranch` plugin were not installed for whatever reason, as there would then be no project type recognizer.

Over the years this code has proven to be full of tech debt and rather trouble-prone, as the various PCT failures and bugs involving dynamic plugin installation can attest to. It also complicates `FolderProperty.allowsParentToCreate` / `.allowsParentToHave`. Nothing else in Jenkins tries to do things this way. This PR just rips out the complicated stuff and leaves you with a routine configuration screen like any other item type would use, where you pick the kind of SCM source after creating the org folder.

Note that this then renders some methods in `SCMNavigatorDescriptor` unused, since they were delegated to in order to implement `TopLevelItemDescriptor` methods. They could be deprecated separately in `scm-api`.


src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -801,118 +775,20 @@ public String getCategoryId() {
...
-            }
-            return r;
-        }
-
-        /**
-         * Gets the {@link SCMSourceCategory#isUncategorized()} of a list of {@link SCMNavigatorDescriptor} instances.
-         * @param descriptors the {@link SCMNavigatorDescriptor} instances.
-         * @return the {@link SCMSourceCategory}.
-         */
-        private SCMSourceCategory genericScmSourceCategory(List<? extends SCMNavigatorDescriptor> descriptors) {
```
jglick, line 903: Cleaning up an unrelated unused method.

src/main/resources/jenkins/branch/OrganizationFolder/configure-entries.properties
```diff
@@ -1,26 +0,0 @@
...
-# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-# THE SOFTWARE.
-#
-deprecated=This organization folder has a legacy configuration and does not have a single repository source. The \
-  ability to have an organization folders with multiple repository sources may be removed in a future release \
```
jglick, line 25: This warning has been there for years but I do not know of any actual plan to switch `List<SCMNavigator>` to a single `SCMNavigator`.

src/test/java/jenkins/branch/CustomOrganizationFolderDescriptorTest.java
```diff
@@ -1,215 +0,0 @@
...
-                MultiBranchProjectFactoryDescriptor.class).get(
-                BasicMultiBranchProjectFactory.DescriptorImpl.class)));
-        assertTrue(ExtensionList.lookup(SCMNavigatorDescriptor.class)
-                .remove(ExtensionList.lookup(SCMNavigatorDescriptor.class).get(
-                        MockSCMNavigator.DescriptorImpl.class)));
-    }
-
-    @Test
-    public void noNavigatorNoFactoryInstalled() throws Exception {
-        assumeThat(ExtensionList.lookup(SCMNavigatorDescriptor.class), hasSize(1));
```
jglick, line 79: Hacks for PCT.

jglick:
Before-and-after screencasts could be produced upon demand, though the before case is a bit complicated (part of the point here!) because it varies depending on which SCM plugins you have installed.

Vlatombe (approved):
LGTM

fcojfernandez (approved):

src/main/java/jenkins/branch/CustomOrganizationFolderDescriptor.java
```diff
@@ -1,221 +0,0 @@
...
-import org.jenkins.ui.icon.IconSpec;
-import org.kohsuke.accmod.Restricted;
-import org.kohsuke.accmod.restrictions.NoExternalUse;
-
-/**
- * Ensures that when we have a non-null {@link SCMNavigatorDescriptor#newInstance(String)}, a special <b>New Item</b> entry is displayed instead of the generic {@link OrganizationFolder.DescriptorImpl}.
- */
-@SuppressWarnings("rawtypes") // not our fault
-@Restricted(NoExternalUse.class)
-public class CustomOrganizationFolderDescriptor extends TopLevelItemDescriptor implements IconSpec {
```
MRamonLeon, line 54: This is kind of used here: https://github.com/jenkinsci/scm-api-plugin/blob/d01f419bc9dcc408e834cbc333b285b1c8058e69/src/main/java/jenkins/scm/api/SCMNavigatorDescriptor.java#L177 Are we safe if we remove this?

src/main/java/jenkins/branch/OrganizationFolder.java
```diff
@@ -329,16 +323,6 @@ public void onLoad(ItemGroup<? extends Item> parent, String name) throws IOExcep
         return super.getItem(NameEncoder.encode(projectName));
     }
 
-    /**
-     * Returns {@code true} if this is a single origin {@link OrganizationFolder}.
-     *
-     * @return {@code true} if this is a single origin {@link OrganizationFolder}.
-     */
-    public boolean isSingleOrigin() {
```
MRamonLeon, line 337: This seems to be used here https://github.com/jenkinsci/tuleap-git-branch-source-plugin which has certain activity

MRamonLeon (commented):
I've found several usages of removed stuff by this PR, can we assess the impact to the dependent plugins?

src/main/java/jenkins/branch/CustomOrganizationFolderDescriptor.java
jglick: Just means that `newInstance(String)` could be deprecated as unused along with other methods mentioned in https://github.com/jenkinsci/branch-api-plugin/pull/271#issue-724943402.

src/main/java/jenkins/branch/OrganizationFolder.java
jglick: Mm, OK, I can restore it but deprecate it.

MRamonLeon (approved):

Vlatombe (approved):

fcojfernandez (approved):

rsandell (approved):

PR #281, opened 2021-10-16
Use `StandardCharsets` where possible

basil:
This reduces our dependency on a third-party library by relying on native Java Platform functionality.


jglick:
Already done at some point it seems.

PR #291, opened 2022-02-03
JENKINS-67725 - Extending BranchBuildStrategies to allow setting last revision built

carpnick:
See [JENKINS-67725](https://issues.jenkins.io/browse/JENKINS-67725) for more context.  

The basic use case:

- How do we enable the folks to "save" the first commit ID found, and only build everything after that.
- I realize this PR will probably not be merged.  But it is meant to be a conversation.

Discussion
- As written [this](https://github.com/jenkinsci/branch-api-plugin/blob/master/src/main/java/jenkins/branch/MultiBranchProject.java#L2017) will always evaluate to true even if a BuildStrategy keeps responding, do not build.  The next time a branch indexing comes in, it will evaluate to true since it did not build.  This feels like a design flaw.  As a plugin author I am telling you to not build it.  As a Jenkins Admin I am telling you not to build it.  But you are repeatedly trying to build it if I turn on branch indexing.
- As I see it we have 2 options
   - We extend BranchBuildStrategy and add another optional method that allows a plugin author to own if you can update the last built hash 
- We create a whole new extension point to give a plugin author the ability to define what `changesDetected` means.

# What is in the PR
- Added a new extension method(optional) for plugin authors can choose to update last built revision or not.  Default is false which is current behavior
- Added integration of new method into MultiBranchProject to evaluate if it should update lastbuiltrevision on disk.

### DIFF
--- a/src/main/java/jenkins/branch/BranchBuildStrategy.java
+++ b/src/main/java/jenkins/branch/BranchBuildStrategy.java
@@ -274,6 +274,33 @@ public abstract class BranchBuildStrategy extends AbstractDescribableImpl<Branch
         return isAutomaticBuild(source, head, currRevision, lastBuiltRevision, lastSeenRevision, listener);
     }
 
+
+
+
+
+    /**
+     * API:Should we update last built revision if we did not do a build?
+     *
+     * @return {@code true} if and only if we should consider whatever commit we received as built.
+     */
+    @SuppressWarnings("deprecation")
+    public final boolean updatingLastBuiltRevisionWithNoBuild() {
+        return isUpdatingLastBuiltRevisionWithNoBuild(new LogTaskListener(Logger.getLogger(getClass().getName()), Level.INFO));
+    }
+
+    /**
+     * SPI: Should we update last built revision if we did not do a build?
+     *
+     * @param listener     the TaskListener to be used
+     * @return {@code true} if and only if we should consider whatever commit we received as built.
+     */
+    @Restricted(ProtectedExternally.class)
+    public boolean isUpdatingLastBuiltRevisionWithNoBuild(@NonNull TaskListener listener) {
+        return false;
+    }
+
+
+
     /**
      * {@inheritDoc}
      */
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -2018,10 +2018,26 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
                     listener.getLogger()
                             .format("Changes detected: %s (%s → %s)%n", rawName, scmLastBuiltRevision, revision);
 
+
                     needSave = true;
                     // get the previous seen revision
                     SCMRevision scmLastSeenRevision = lastSeenRevisionOrDefault(project, scmLastBuiltRevision);
                     doAutomaticBuilds(head, revision, rawName, project, revisionActions, scmLastBuiltRevision, scmLastSeenRevision);
+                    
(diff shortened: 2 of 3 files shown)


carpnick:
Could I get a review @jtnord and/or @rsandell ?

carpnick:
Any kind of review would be appreciated...
@timja  or @car-roll

src/main/java/jenkins/branch/BranchBuildStrategy.java
```diff
@@ -160,13 +161,46 @@ public boolean isAutomaticBuild(@NonNull SCMSource source,
...
+    @Deprecated
+    @SuppressWarnings("deprecation")
+    @Restricted(ProtectedExternally.class)
+    public boolean isAutomaticBuild(@NonNull SCMSource source,
+                                             @NonNull SCMHead head,
+                                             @NonNull SCMRevision currRevision,
+                                             @CheckForNull SCMRevision lastBuiltRevision,
+                                             @CheckForNull SCMRevision lastSeenRevision,
+                                             @NonNull TaskListener listener){
+        throw new UnsupportedOperationException("Modern implementation accessed using legacy API method");
```
jtnord, line 173: this seems like a breaking change?  if an Extension overrides this calss and updated to the modern way removing the over ride on this deprecated function - then any consumers that have not updated will not blow up at runtime?

If the extension can do something about it - then it would seem this method is just un-needed - or could be used to call the modern method using the 
`Util.isOverridden` trick.

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -2207,6 +2207,21 @@ private void doAutomaticBuilds(@NonNull SCMHead head, @NonNull SCMRevision revis
                 );
             } else {
                 listener.getLogger().format("No automatic build triggered for %s%n", rawName);
+
+                //Should we update the last Build revision with no build performed?
+                if(isUpdatingLastBuiltRevisionWithNoBuild()){
```
jtnord, line 2212:
```suggestion
                if (isUpdatingLastBuiltRevisionWithNoBuild()) {
```

jtnord (commented):
a very quick first pass check seems like this has the potential to break things.

I would also expect at least one downstream PR to demonstrate this new API.

ref: https://www.jenkins.io/doc/developer/extensions/branch-api/#branchbuildstrategy

I need to find some time to go through this in some more detail, but I do not have that time right now,

src/main/java/jenkins/branch/BranchBuildStrategy.java
carpnick: I actually followed the example of what the rest of the plugin extension did.  IE - Line [111](https://github.com/jenkinsci/branch-api-plugin/blob/master/src/main/java/jenkins/branch/BranchBuildStrategy.java#L111) was originally shipped with 2.0.17 while line [138](https://github.com/jenkinsci/branch-api-plugin/blob/master/src/main/java/jenkins/branch/BranchBuildStrategy.java#L138) was shipped with 2.1.3.  Basically everytime there is a new version, the previous version is set to `UnsupportedOperationException`.

So the way I understand it the SPI method finds out dynamically which method is overridden (original, V2, V3, etc) and then calls that specific override.  So the SPI method uses `Util.isOverridden` to determine which API method to call.

carpnick:
> I would also expect at least one downstream PR to demonstrate this new API.
> 
> ref: https://www.jenkins.io/doc/developer/extensions/branch-api/#branchbuildstrategy

Fair comment - I will either do a PR in one of the plugins that exists and/or create a plugin and link it here.

PR #304, opened 2022-03-25
JENKINS-68116 Increase log size limit

timja:
Previously each log file was limited to 33KB and 5 files.
On our busy Jenkins instance we only got 15 minutes worth of logs before
they were rotated out.

It would be good for this to be increased to make analysis of issues
easier

see https://issues.jenkins.io/browse/JENKINS-68116

### DIFF
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -1138,7 +1138,7 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
                     250, TimeUnit.MILLISECONDS,
                     1024,
                     true,
-                    32 * 1024,
+                    FileUtils.ONE_MB * 10,
                     5
             );
         }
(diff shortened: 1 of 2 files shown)


timja:
This was very useful for debugging https://issues.jenkins.io/browse/JENKINS-68116

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -1138,7 +1138,7 @@ public File get() {
                     250, TimeUnit.MILLISECONDS,
                     1024,
                     true,
-                    32 * 1024,
+                    FileUtils.ONE_MB * 10,
```
jtnord, line 1141: isn't this just a tad excessive (50MB of event processing logs)?

if previously the size of logs with your load would keep 15 minutes then this would hold 305 times that (or 76 hours worth or a tad over 3 days).  I would expect that 8 hours would be enough which would be (assuming all my calculation are correct) :

```suggestion
                    FileUtils.ONE_MB,
```

jtnord (approved):
Approved with reservations, if others disagree with me feel free to take this as an approval

jtnord (approved):

timja:
Possible to release this please @jtnord ?

jtnord:
> Possible to release this please @jtnord ?

another repo without CD :-(

PR #347, opened 2022-11-22
Remove no longer necessary workaround

basil:
commit 2dae2ca429b155c43f20d3 added a workaround that is no longer necessary in recent versions of SpotBugs.


jglick (approved):

Saibamen:
This is a breaking change (available for Java 11 agents only)?

basil:
I do not see why it would be breaking.

PR #369, opened 2023-05-10
BuildStrategies implementation need to know the current job

olamy:
Signed-off-by: Olivier Lamy <olamy@apache.org>

### Submitter checklist

### DIFF
--- a/src/main/java/jenkins/branch/BranchBuildStrategy.java
+++ b/src/main/java/jenkins/branch/BranchBuildStrategy.java
@@ -28,6 +28,7 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.ExtensionPoint;
 import hudson.Util;
 import hudson.model.AbstractDescribableImpl;
+import hudson.model.Job;
 import hudson.model.TaskListener;
 import hudson.util.LogTaskListener;
 import java.util.logging.Level;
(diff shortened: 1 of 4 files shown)


src/main/java/jenkins/branch/BranchBuildStrategy.java
```diff
@@ -161,12 +162,22 @@ public boolean isAutomaticBuild(@NonNull SCMSource source,
...
+    public boolean isAutomaticBuild(@NonNull SCMSource source,
+                                             @NonNull SCMHead head,
+                                             @NonNull SCMRevision currRevision,
+                                             @CheckForNull SCMRevision lastBuiltRevision,
+                                             @CheckForNull SCMRevision lastSeenRevision,
+                                             @NonNull TaskListener listener) {
+        return isAutomaticBuild(source, head, currRevision, lastBuiltRevision, lastSeenRevision, listener, null);
+    }
+
     public abstract boolean isAutomaticBuild(@NonNull SCMSource source,
```
olamy, line 174: this pattern of keep adding parameters is getting worst and worst :(
Maybe we could have something such (pseudo code) to avoid breaking backward compat with every new data: 
```
class AutomaticBuild {
  SCMHead head;
  SCMRevision currRevision;
  // all the getters/setters
}
```
or wait to be 17 here and use `record` :)

jglick: We can use `record` now.

Is this PR still relevant? Is there some proposed caller?

PR #254, opened 2021-06-01
[JENKINS-55348] Use supplied object display names for the job

cobexer:
If the SCMHead has an ObjectMetadataAction that defines a display name use it for the job display name

This causes jobs for GitLab Merge Requests, GitHub Pull Requests, ... to display like:
"MR-1234 - Enhanced Job Display Names"
inspired by: https://github.com/jenkinsci/gitlab-branch-source-plugin/issues/141

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #374, opened 2023-05-17
[JENKINS-55348] Display Pull Request Name instead of ID in the UI

quilicicf:
https://issues.jenkins.io/browse/JENKINS-55348

## This PR builds upon another

I've merely pulled @cobexer's [previous pull request](https://github.com/jenkinsci/branch-api-plugin/pull/254) and added a test to tick all boxes below.

## Original description

If the `SCMHead` has an `ObjectMetadataAction` that defines a display name use it for the job display name

This causes jobs for GitLab Merge Requests, GitHub Pull Requests, ... to display like:
`MR-1234 - Enhanced Job Display Names`
inspired by: https://github.com/jenkinsci/gitlab-branch-source-plugin/issues/141

- [x] Make sure you are opening from a topic/feature/bugfix branch (right side) and not your master branch!

## Details on the test part

To test this new feature, I had to create a multi-branch job with a pull request inside.

I first tried to do that without adding other dependencies, by implementing a fake `SCMSource` but that proved to be too hard for my current knowledge of the Jenkins ecosystem.

I ended up using the `github-branch-source` plugin (only in tests, of course) and mocking all the REST API calls it does to the GitHub API.

This is coupled with GitHub which sucks a bit, but since
* it is one of the most used with 200k+ installs at the moment of writing, (page two of 1895 on the [plugins repository](https://plugins.jenkins.io)) => hopefully very stable
* it relies on GitHub which is a behemoth with a very stable API too

I think this is an acceptable solution.


lemeurherve:
Genuine question, why not adding these tests as a PR targeting https://github.com/cobexer/branch-api-plugin/tree/merge-request-project-display-names instead of creating a new one here?

quilicicf:
It's actually the same thing since I pulled @cobexer's commit so credit is given where it's due? 
I don't mind either, just didn't really think about it.

jtnord:
Having looked at this briefly I have some concerns.

The name that is possible to set in the SCM may not be useable in Jenkins as the disaplyName goes via a markup formatter (set globally).

so a name like "fix <bold> text"  when used with the "Safe HTML" formatter would not disaply correctly and could break other parts of the page.

given the need to correctly escape between systems that have differing escapes I am not convicinced that the entirity of the functionality should reside here (and not as an enum).  rather this project should merley provide a hook, and other plugins should be able to implement that taking into account whatever markup formater they desire.

jtnord:
Hi @quilicicf

> As for the escaping, I must admit I didn't realize this, I just added a test to the existing code without challenging the technical solution. I thought the design [had been accepted](https://github.com/jenkinsci/branch-api-plugin/pull/254#issuecomment-858236603) but I may have read too much into that.

Sorry about that - I was confusing `displayName` with `description`.  Whilst the latter has markup, the former does not.

> For the name length, I don't see how it's worse than in the `Branches` view. Branch names are unbound in git (or at least [no limitation is documented](https://git-scm.com/docs/git-check-ref-format)) so the length can be greater than 80 chars too?

This is correct - however in my experience people do not generally use long branch names that they build (as they would be filed as MRs, or are WIP so are on a fork).   Dependabot is bad - but these are always PRs/MRs so there should not be the need to disaply them,  and then you would be left with something like `/features/ticketref/subtask`  which would generally be shorter than "[ticket] make the foobar wibble visible to firefox users" (as by its nature the PR/MR title would generally be for humans whereas the branch name is less so.

> I hope the HTML tables are able to handle that somehow (an ellipse?) but I think the issue should be tackled in the UI?

As far as I am aware, it displays whatever you tell it to, unless you do something special in the Jelly that renders the "thing".

Having said that - I did a quick mock up by just changing the source code and things wrap - and this is opt-in so it's not that bad, so I will take another look through the code.

image

src/test/java/jenkins/branch/naming/ProjectNamingStrategyTest.java
```diff
@@ -0,0 +1,162 @@
...
+    public static final JenkinsRule r = new JenkinsRule();
+
+    public static String stubBaseUrl ;
+    public static HttpServer githubApiStub;
+
+    @BeforeClass
+    public static void setupGithubApiStub() throws Exception {
+        final int stubPort = findAvailablePort();
+        stubBaseUrl = String.format("http://localhost:%s", stubPort);
+        githubApiStub = HttpServer.create(new InetSocketAddress(stubPort), 0);
```
jtnord, line 60: See comment below on `findAvailablePort()`  
Additionally this binds to every interface which seems overkill and one jdk exploit / firewall misconfiguration away from an accident.


Untested but something like the following would be preferable.

```java
        githubApiStub = HttpServer.create(new InetSocketAddress(netAddress.getLoopbackAddress(), 0), 0);

       // all the other code as the port will not be available until we `start` the server

        // obtain the actual bound address.
        final InetSocketAddress boundSocket = githubApiStub.getAddress();
        final int port = boundSocket.getPort();

        final InetAddress addr = boundSocket.getAddress();
        // literal IPv6 addresses need special encoding
        final String address;
        if (addr instanceOf Inet6Address) {
            address = "[" + addr.getHostAddress()) + "]";
        } else {
            address = addr.getHostAddress());
       }
        stubBaseUrl = String.format("http://%s:%d", address, port);
```

src/test/java/jenkins/branch/naming/ProjectNamingStrategyTest.java
```diff
@@ -0,0 +1,162 @@
...
+        assertEquals(expectedName, prProject.getDisplayName(), "The job name doesn't match the naming strategy");
+    }
+
+    private static int findAvailablePort() {
+        try (final ServerSocket serverSocket = new ServerSocket(0)) {
+            return serverSocket.getLocalPort();
+        } catch (final IOException e) {
+            throw new RuntimeException("Could not find an available port for GitHub API stub");
+        }
+    }
```
jtnord, line 156: this is a pure race condition and will lead to flaky tests.

the port is available at T1, but may not be at T2 when it is used by wiremock.

Rather get HttpServer to bind to any available random port (`0`) and then ask it which port it bound to.

src/test/java/jenkins/branch/naming/ProjectNamingStrategyTest.java
```diff
@@ -0,0 +1,162 @@
...
+ * most used (and fingers-crossed, stable) in the Jenkins ecosystem, and mocking a REST API is trivial, the tradeoff is
+ * acceptable.
+ */
+public class ProjectNamingStrategyTest {
+    private static final Logger LOGGER = Logger.getLogger(ProjectNamingStrategyTest.class.getName());
+
+    @ClassRule
+    public static final JenkinsRule r = new JenkinsRule();
+
+    public static String stubBaseUrl ;
```
jtnord, line 53:
```suggestion
    public static String stubBaseUrl;
```

src/test/java/jenkins/branch/naming/githubapimock/ResourceHandler.java
```diff
@@ -0,0 +1,40 @@
...
+        }
+    }
+
+    @Override
+    public void handle(final HttpExchange exchange) throws IOException {
+        LOGGER.fine(String.format("Response for %s: %s", url, response));
+        exchange.sendResponseHeaders(statusCode, response.length());
+        final OutputStream outputStream = exchange.getResponseBody();
+        outputStream.write(response.getBytes());
+        outputStream.close();
```
jtnord, line 38:
```suggestion
        try (OutputStream outputStream = exchange.getResponseBody()) {
            outputStream.write(response.getBytes());
        }
```

src/test/java/jenkins/branch/naming/ProjectNamingStrategyTest.java
```diff
@@ -0,0 +1,162 @@
...
+    public static HttpServer githubApiStub;
+
+    @BeforeClass
+    public static void setupGithubApiStub() throws Exception {
```
jtnord, line 57: I wonder how maintainable this would be, there have been changes in the GH API that meant the branch source changes API calls, rather than implementing a few `@TestExtensions`  and/or [Mock SCM implementation](https://github.com/jenkinsci/scm-api-plugin/tree/master/src/test/java/jenkins/scm/impl/mock)

src/main/java/jenkins/branch/MultiBranchProjectDisplayNamingStrategy.java
```diff
@@ -0,0 +1,78 @@
...
+public enum MultiBranchProjectDisplayNamingStrategy {
+    /**
+     * Use the name as specified by the scm connector. (traditional behaviour)
+     */
+    RAW(false) {
+        @Override
+        public String generateName(@NonNull final String rawName, final String displayName) {
+            return rawName;
+        }
+    },
```
jtnord, line 46: why would you have this.  this is the behavior when the trait does not exist - so to get this behaviour just delete the trait?

pom.xml
```diff
@@ -147,10 +132,55 @@
       <classifier>tests</classifier>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>github-branch-source</artifactId>
+      <version>2.2.4</version>
+      <scope>test</scope>
+      <exclusions>
```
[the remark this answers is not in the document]
quilicicf: Strange, I only followed the advice of my IDE :thinking: 

Guess it's confused, can remove exclusions :+1:

src/main/java/jenkins/branch/MultiBranchProjectDisplayNamingTrait.java
```diff
@@ -0,0 +1,104 @@
...
+    public MultiBranchProjectDisplayNamingStrategy getDisplayNamingStrategy() {
+        return displayNamingStrategy;
+    }
+
+    /**
+     * Our descriptor.
+     */
+    @Symbol("multiBranchProjectDisplayNaming")
+    @Extension
+    @Discovery
```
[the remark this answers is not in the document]
quilicicf: I guess not really, that's @cobexer's commit and I didn't change much of it but if this is a copy-paste oopsie I can definitely remove this line :+1:

src/main/java/jenkins/branch/MultiBranchProjectDisplayNamingTrait.java
```diff
@@ -0,0 +1,104 @@
...
+        @Restricted(NoExternalUse.class)
+        @SuppressWarnings("unused") // stapler
+        public ListBoxModel doFillDisplayNamingStrategyItems() {
+            ListBoxModel result = new ListBoxModel();
+            result.add(Messages.MultiBranchProjectDisplayNamingTrait_Raw(), MultiBranchProjectDisplayNamingStrategy.RAW.name());
+            result.add(Messages.MultiBranchProjectDisplayNamingTrait_DisplayName(), MultiBranchProjectDisplayNamingStrategy.OBJECT_DISPLAY_NAME.name());
+            result.add(Messages.MultiBranchProjectDisplayNamingTrait_RawAndDisplayName(), MultiBranchProjectDisplayNamingStrategy.RAW_AND_OBJECT_DISPLAY_NAME.name());
+            return result;
+        }
+    }
```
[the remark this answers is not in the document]
quilicicf: I'll try to improve this :+1:

src/test/java/jenkins/branch/naming/ProjectNamingStrategyTest.java
quilicicf: I'll try this :+1:

src/test/java/jenkins/branch/naming/ProjectNamingStrategyTest.java
quilicicf: Yeah, I tried with mock SCM implementations but ended up in a pretty deep rabbit hole, probably because I'm not aware of how it should be implemented and that's a lot of investment to do to write a single test :sweat_smile: 

Is there any way to re-use an existing mock SCM implementation? Would be happy to re-write this test to use a mock SCM instance but a little more help would go a long way :sweat_smile:

src/main/java/jenkins/branch/MultiBranchProjectDisplayNamingStrategy.java
quilicicf: That's an interesting take.
Didn't consider it and never even played with it in the UI so I didn't really realize what it feels like using it. 
Will try to do that :+1:

src/main/java/jenkins/branch/MultiBranchProjectDisplayNamingTrait.java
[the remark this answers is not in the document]
quilicicf: Removed it.

src/main/java/jenkins/branch/MultiBranchProjectDisplayNamingTrait.java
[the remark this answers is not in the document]
quilicicf: I switched to an enum, and I followed exactly the pattern in the other enum I found in the repository: [SuppressionStrategy](https://github.com/quilicicf/branch-api-plugin/blob/4ec3ed07be4f8f30e5444b071d5d7bc525606bd0/src/main/java/jenkins/branch/NoTriggerMultiBranchQueueDecisionHandler.java#L148-L170). It is used in 
* [this jelly template](https://github.com/quilicicf/branch-api-plugin/blob/4ec3ed07be4f8f30e5444b071d5d7bc525606bd0/src/main/resources/jenkins/branch/NoTriggerBranchProperty/config.jelly#L30-L34)
* and [this one](https://github.com/quilicicf/branch-api-plugin/blob/4ec3ed07be4f8f30e5444b071d5d7bc525606bd0/src/main/resources/jenkins/branch/NoTriggerOrganizationFolderProperty/config.jelly#L31-L35)

Since it contains the `NONE` entry, and the enum field requires a default value, I kept the `RAW` strategy for now. If you prefer removing it, please tell me what default value I should use instead :pray:

src/test/java/jenkins/branch/naming/ProjectNamingStrategyTest.java
quilicicf: Tested OK, and added.

src/test/java/jenkins/branch/naming/ProjectNamingStrategyTest.java
quilicicf: Indeed, fixed.

src/main/java/jenkins/branch/MultiBranchProjectDisplayNamingStrategy.java
quilicicf: See [my other comment](https://github.com/jenkinsci/branch-api-plugin/pull/374#discussion_r1198644264) let's continue the discussion there.

quilicicf:
@jtnord Taken all the comments I could into account. 
I'll need another pass from you for the two remaining ones please :pray:

src/test/java/jenkins/branch/naming/ProjectNamingStrategyTest.java
jtnord: [`EnvironmentTest.given_multibranch_when_buildingAChangeRequest_then_environmentContainsBranchName()`](https://github.com/jenkinsci/branch-api-plugin/blob/da3c2aaf0c1160a5aea0146a3e7ec03906d42269/src/test/java/integration/EnvironmentTest.java#L86-L102)  may give you some pointers.

src/main/java/jenkins/branch/MultiBranchProjectDisplayNamingTrait.java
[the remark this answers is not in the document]
jtnord: I've asked our internal UX expert for opinions (and will update when I hear back)

in the interim my preference would be remove the `NONE` and make the `RawAndDisplayName` the default.

src/main/resources/jenkins/branch/MultiBranchProjectDisplayNamingTrait/help-displayNamingStrategy.html
jtnord, line 1: for some reason this does not show up, but also there is no overall help for the MultiBranchProjectDisplayNamingTrait

image

pom.xml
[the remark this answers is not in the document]
jtnord: alas the `hpi` packaging type is special and confuses many things due to its non standard hierachy of dependencies (for what actually gets packaged as a dependency)    (transitive dependencies that come from another plugin will not be packaged - but IDEs are unaware of this - similarly many tools they are unaware that depending on `someplugin v1`  where `v1` is vulnerable to a CVE is not an issue as all that the dependency is saying in the project is that v1 is a minimum version it can work with).

src/test/java/jenkins/branch/naming/ProjectNamingStrategyTest.java
quilicicf: I'll have a look, thanks for the pointer.

src/main/java/jenkins/branch/MultiBranchProjectDisplayNamingTrait.java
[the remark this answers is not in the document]
quilicicf: I'll hold off until we hear back from the expert, I'll have plenty of work in the meantime to try and re-implement the test with a mock SCM.

src/main/resources/jenkins/branch/MultiBranchProjectDisplayNamingTrait/help-displayNamingStrategy.html
quilicicf: I'll try to find out why :+1:

cobexer:
@quilicicf Thanks for pushing this along 😊

quilicicf:
I have implemented most suggestions on this PR with the notable exception of the re-implementation of the test with a mock which is going to take a longer working session than I could afford lately.

I discovered a nice little bug that impacts the whole form though. The trait help is only displayable when the trait is the first that appears in the form.

Whenever I click on the `?` icon next to a trait name, it opens the help of the first trait in the form, not the one I clicked on! 

I have absolutely no clue about how to fix it and that's out of scope of this PR, so I'll just post the screenshot of the full documentation like it appears when only the naming trait is selected and let you guys handle the other bug elsewhere (maybe it's already been reported before?).

Selection_007

I'll try to come up with a re-implementation of the test with a mock SCM ASAP so we can merge this.

src/main/java/jenkins/branch/MultiBranchProjectDisplayNamingTrait.java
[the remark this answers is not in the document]
quilicicf: Done!

quilicicf:
Re-writing the test was simpler than expected, it's done.
Let's try to make all tests pass now :sweat_smile:

quilicicf:
And... done. 
The ball's in your court @jtnord!

jtnord:
Many thanks @quilicicf @cobexer!

quilicicf:
You're welcome, thanks for your guidance :100: 
Please keep us posted when a new version gets released :pray:

car-roll:
@quilicicf this repo has automatic CD enabled, so the changes have already been released: https://github.com/jenkinsci/branch-api-plugin/releases/tag/2.1105.v472604208c55

car-roll:
@quilicicf done! https://github.com/jenkinsci/branch-api-plugin/releases/tag/2.1105.v472604208c55

danielnilsson9:
Is there no way to get the old behavior back, look like there is no RAW option? Problem with this is that it makes the list of builds in Bitbucket very hard to read and I would rather trade nice PR names in Jenkins for a readable list of build statuses in Bitbucket.

quilicicf:
Hi @danielnilsson9 

If I recall correctly, not setting the new parameter for the display name strategy keeps the old behavior so unless you've added it, you should see no change related to this PR. Check whether you have the configuration [multiBranchProjectDisplayNaming](https://github.com/quilicicf/presentations/blob/master/src/jenkins-for-the-noobs/gitops-repo/src/main/groovy/jobs/myproduct/myproduct.groovy#L31) in your Jobs and remove it to opt out.

I also think that the name of the build in Bitbucket/GitHub/... is unrelated to this feature, but it may have an impact I don't see (it's been a long time since I worked on this).

You should be able to give the builds in Bitbucket the names you want. I use the plugin [GitHub Custom Notification Context SCM Behaviour](https://plugins.jenkins.io/github-scm-trait-notification-context/) in [my jobs](https://github.com/quilicicf/presentations/blob/master/src/jenkins-for-the-noobs/gitops-repo/src/main/groovy/jobs/myproduct/myproduct.groovy#L36). There may be a similar plugin for Bitbucket ?

danielnilsson9:
Thanks for answers

@quilicicf  Yes I read somewhere that the behavior would change if the trait wasn't configured but the behavior change when we upgraded our plugins without doing any configuration changes in our job dsl generation scripts. We don't have the multiBranchProjectDisplayNaming trait configured at all. So I'm afraid it does actually change by default.

@jtnord Oh, yes, missed that this change PR was done in 2023. Maybe something in the PR you linked changed the default behavior. I will investigate and file an issue if there is one.

PR #58, opened 2016-11-17
[JENKINS-39355 follow up] Add eventing support to API

stephenc:
I give up tracking what Jenkins issues this sweeps up.

Big improvements to the branch API

- Lots of simplifications to how customizations and branding get applied
- Now we support events from SCM API and optimize reindexing using the event information
- Lots of tests of the event support added
- Some other bugs found and fixed

@reviewbybees

PR #383, opened 2023-06-21
`RateLimitBranchProperty`: Fix checking queue for other items tied to the same task

Vlatombe:
`getId` is not comparable, but `getInQueueSince` is.

This also adds a `LoggerRule` to the test class to make the test output more readable.

cc @jglick 

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/jenkins/branch/RateLimitBranchProperty.java
+++ b/src/main/java/jenkins/branch/RateLimitBranchProperty.java
@@ -533,23 +533,22 @@ public class RateLimitBranchProperty extends BranchProperty {
                     }
                     // ensure items leave the queue in the order they were scheduled
                     List<Queue.Item> items = Jenkins.get().getQueue().getItems(item.task);
-                    if (items.size() == 1 && item == items.get(0)) {
-                        return null;
-                    }
-                    for (Queue.Item i : items) {
-                        if (i.getId() < item.getId()) {
-                            LOGGER.log(Level.FINE, "{0} with queue id {1} blocked by queue id {2} was first",
-                                    new Object[]{
-                                            job.getFullName(),
-                                            item.getId(),
-                                            i.getId()
-                                    }
-                            );
+                    if (!items.isEmpty()) {
+                        Queue.Item i = items.get(0);
+                        if (item == i) { // This item is the first in queue for this task.
+                            LOGGER.log(Level.FINE, () -> "Allowing " + item.getId() + " " + item.getClass().getSimpleName() + " " + job.getFullName());
+                            return null;
+                        } else {
+                            // There is another item before the one currently processed, so block.
+                            LOGGER.log(Level.FINE, () -> item.getId() + " " + job.getFullName() + " blocked by queue id " + i.getId() + " was first");
                             long betweenBuilds = property.getMillisecondsBetweenBuilds();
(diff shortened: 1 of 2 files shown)


src/main/java/jenkins/branch/RateLimitBranchProperty.java
```diff
@@ -533,23 +533,22 @@ public CauseOfBlockage canRun(Queue.Item item) {
...
-                                    }
-                            );
+                    if (!items.isEmpty()) {
+                        Queue.Item i = items.get(0);
+                        if (item == i) { // This item is the first in queue for this task.
+                            LOGGER.log(Level.FINE, () -> "Allowing " + item.getId() + " " + item.getClass().getSimpleName() + " " + job.getFullName());
+                            return null;
+                        } else {
+                            // There is another item before the one currently processed, so block.
+                            LOGGER.log(Level.FINE, () -> item.getId() + " " + job.getFullName() + " blocked by queue id " + i.getId() + " was first");
```
jglick, line 543: This branch has no test coverage. It is not exactly clear to me the circumstances under which it would be activated—a parameterized build with multiple (distinct) items in queue? The former logic does not look reliable (there is no documentation saying `Queue.Item.getId` is comparable) but neither does this (there is no documentation indicating an order of `Queue.getItems(Queue.Task)`).

The logic being edited was added in 4ed032736012892bcee8b93c7128af84d1c66dd8 (#58) and added `RateLimitBranchPropertyTest.rateLimitsConcurrentBuilds`. With logging from this PR but logic from `master`, that test does print

```
FINE	j.b.RateLimitBranchProperty$QueueTaskDispatcherImpl#canRun: foo/master with queue id 4 blocked by queue id 3 was first
```

At that point `items` contains 4 then 3 in that order and we are checking 4, so the desired behavior is to block it.

I think what this should really be doing is closer to the original logic, but using `getInQueueSince()` rather than `getId()`.

jglick (changes requested):

src/main/java/jenkins/branch/RateLimitBranchProperty.java
```diff
@@ -537,7 +537,7 @@ public CauseOfBlockage canRun(Queue.Item item) {
                         return null;
                     }
                     for (Queue.Item i : items) {
-                        if (i.getId() < item.getId()) {
+                        if (i.getInQueueSince() < item.getInQueueSince()) {
```
jglick, line 540: Now printing

```
foo/master with queue id 4 blocked by queue id 3 was first
```

as expected.

jglick (approved):

jglick:
I suspect you need a short pause between https://github.com/jenkinsci/branch-api-plugin/blob/f9e35c47c34d25c4903bddd0d845d4731cb1df02/src/test/java/jenkins/branch/RateLimitBranchPropertyTest.java#L250-L253 to ensure that the items have distinct start times, since this is only millisecond resolution; or https://github.com/jenkinsci/branch-api-plugin/blob/f9e35c47c34d25c4903bddd0d845d4731cb1df02/src/main/java/jenkins/branch/RateLimitBranchProperty.java#L540 should use `<=` while checking (`&&`) that the `Item`s or their `id`s are `!=`.

Vlatombe:
The latter makes more sense to me since in theory this condition can still happen (if jobs are triggered via a script).

#385

PR #388, opened 2023-07-07
[BEE-37070] add inline documentation for build strategies

ampuscas:
Add inline documentation for Build Strategies

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/resources/jenkins/branch/BranchSource/config.jelly
+++ b/src/main/resources/jenkins/branch/BranchSource/config.jelly
@@ -43,7 +43,7 @@
   </j:if>
   <j:set var="descriptors" value="${descriptor.buildStrategiesDescriptors(it,sourceDescriptor)}"/>
   <j:if test="${!descriptors.isEmpty()}">
-    <f:entry title="${%Build strategies}">
+    <f:entry title="${%Build strategies}" field="buildStrategies">
     <f:hetero-list name="buildStrategies" items="${instance.buildStrategies}"
                    descriptors="${descriptors}"
                    hasHeader="true" honorOrder="true"/>


src/main/resources/jenkins/branch/BranchSource/config.jelly
```diff
@@ -43,7 +43,8 @@
   </j:if>
   <j:set var="descriptors" value="${descriptor.buildStrategiesDescriptors(it,sourceDescriptor)}"/>
   <j:if test="${!descriptors.isEmpty()}">
-    <f:entry title="${%Build strategies}">
+    <f:entry title="${%Build strategies}" field="buildStrategies">
+    <f:entry title="${%Strategy}" field="strategyId"></f:entry>
     <f:hetero-list name="buildStrategies" items="${instance.buildStrategies}"
```
jglick, line 48: The fake `f:entry` does not look right. You should not need to resort to tricks to have help for a standard form control. I think the problem here is that the original page used unnecessarily low-level controls. Better to use the higher-level `f:repeatableHeteroProperty` which groks databinding and should allow you to simply create a file with the conventional name, `help-buildStrategies.html`.

src/main/resources/jenkins/branch/BranchSource/config.jelly
```diff
@@ -43,7 +43,7 @@
   </j:if>
   <j:set var="descriptors" value="${descriptor.buildStrategiesDescriptors(it,sourceDescriptor)}"/>
   <j:if test="${!descriptors.isEmpty()}">
-    <f:entry title="${%Build strategies}">
+    <f:entry title="${%Build strategies}" field="buildStrategies">
```
jglick, line 46: So, showing https://github.com/jenkinsci/branch-api-plugin/blob/09d4261ec54b6773765cec76ac07e50bf879f191/src/main/resources/jenkins/branch/BranchSource/help-buildStrategies.html#L2-L4

jglick (commented):
Sort of works. Gets messed up when you click the help next to an actual strategy like **Skip initial build on first branch indexing**, which I suspect is due to this facet still using a nonstandard control.

buildStrategies

PR #425, opened 2024-01-04
Adapt the standard Jenkins icons

NotMyFault:
The change proposed adapts the standard Jenkins icons (ionicons) used in core itself. In this case, the icons look most the same, but ionicons support our theme-ability API and are recommended to use.


jglick (approved):
Looks OK in `hpi:run`.

NotMyFault:
Thanks!

PR #452, opened 2024-05-01
Use symbol for organisation icon

janfaracik:
Follow up to @mawinter69 changes in core (https://github.com/jenkinsci/jenkins/pull/9127). This PR uses the organisation symbol now that the 'New item' page supports displaying them.

**Before**
image

**After**
image

Happy to change the icon if others have better suggestions - I based it off of the 'New organisation' button in GitHub:

Screenshot 2024-05-01 at 12 56 50

### Proposed changelog entries

* Use symbol for organisation icon

### Submitter checklist

- [x] JIRA issue is well described
- [x] Changelog entry appropriate for the audience affected by the change (users or developer, depending on the change).
      * Use the `Internal: ` prefix if the change has no user-visible impact (API, test frameworks, etc.)
- [x] Appropriate autotests or explanation to why this change has no tests


pom.xml
```diff
@@ -68,7 +68,7 @@
     <revision>2</revision>
     <changelist>999999-SNAPSHOT</changelist>
     <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
-    <jenkins.version>2.426.3</jenkins.version>
+    <jenkins.version>2.454</jenkins.version>
```
jtnord, line 71: this is newer than the next LTS (2.452.1) which means any updates here will not work for another 3 months for the majority of users.
I find this not a good solution, can you come up with an alternative?  or should we just leave this PR for 3 months?

janfaracik: Restored the existing icon - it'll use the symbol on newer versions and the old icon on older versions of Jenkins.

jtnord (approved):
tested locally with `mvn hpi:run` and `mvn hpi:run -Djenkins.version=2.454` LGTM thankyou

pom.xml
jglick: We have been updating various plugins to 2.454 for the same reason. I do not see a problem. If there are features or fixes that are _requested_ by LTS users it is very easy to set up a backport branch and cherry-pick them.

jglick: There should at least be a follow-up PR here that reverts 9c7cd5634588ced62b3f2efb3013c77c5ba0bb8c so we know to clean up the tech debt; otherwise it will be forgotten forever.

jtnord: >it is very easy to set up a backport branch and cherry-pick them

As far as I am aware CD can not release stable branches - and as shown here there was no need to actually do this bump to get this feature.

> We have been updating various plugins to 2.454 for the same reason

that was your personal choice, and if I had seen it elswhere I would have also left the same comment for the same reasons.

>  If there are features or fixes that are requested by LTS users

the overwhelming majority of users are not using a weekly with this plugin. (18794 or on or below 2.440.2  vs 2434 on a version between 2.441 - 2.451  so approx. 10%.)  
So by pure guesswork and extrapolation I would probably say most of the bugs are from LTS users and given that if someone files a bug they would really want a fix or they would not bother, then that would be backporting every fix.

jtnord: Click the button in the workflow is CD 😂😂

PR #459, opened 2024-06-04
Deprecating `BranchProperty`s which do not work with Pipeline

jglick:
See https://github.com/jenkinsci/pipeline-plugin/pull/209. These would be used only by https://plugins.jenkins.io/multi-branch-project-plugin/ (deprecated), https://github.com/jenkinsci/freestyle-multibranch-plugin (not on UC), https://github.com/jenkinsci/literate-plugin (ditto), or https://github.com/jenkinsci/yaml-project-plugin (ditto).

### DIFF
--- a/src/main/java/jenkins/branch/BuildRetentionBranchProperty.java
+++ b/src/main/java/jenkins/branch/BuildRetentionBranchProperty.java
@@ -15,8 +15,9 @@ import java.security.AccessController;
 import java.security.PrivilegedAction;
 
 /**
- * @author Stephen Connolly
+ * @deprecated Suppressed by {@code JobPropertyStep.HideSuperfluousBranchProperties} for Pipeline.
  */
+@Deprecated
 public class BuildRetentionBranchProperty extends BranchProperty {
 
     private final BuildDiscarder buildDiscarder;
--- a/src/main/java/jenkins/branch/RateLimitBranchProperty.java
+++ b/src/main/java/jenkins/branch/RateLimitBranchProperty.java
@@ -53,9 +53,9 @@ import org.kohsuke.stapler.StaplerRequest;
 
 /**
  * A branch property that limits how often a specific branch can be built.
- *
- * @author Stephen Connolly
+ * @deprecated Suppressed by {@code JobPropertyStep.HideSuperfluousBranchProperties} for Pipeline.
  */
+@Deprecated
 @SuppressWarnings("unused") // instantiated by stapler
 public class RateLimitBranchProperty extends BranchProperty {
 
--- a/src/main/java/jenkins/branch/UntrustedBranchProperty.java
+++ b/src/main/java/jenkins/branch/UntrustedBranchProperty.java
@@ -62,9 +62,9 @@ import org.jvnet.tiger_types.Types;
  * locations</li>
  * </ul>
  *
- * @author Stephen Connolly
- * @author Kohsuke Kawaguchi
+ * @deprecated Not used by Pipeline.
  */
+@Deprecated
 public class UntrustedBranchProperty extends BranchProperty {
 
     private final Set<String> publisherWhitelist;


jonesbusy:
I'm a bit confused with this deprecation.

I found some of my user jobs (created by JobDSL using such properties)

```groovy
multibranchPipelineJob("job-name") {
    ....
    branchSources {
        branchSource {
            ...
            strategy {
                ...
                namedBranchesDifferent {
                    defaultProperties {
                        ...
                        buildRetention {
                            buildDiscarder {
                                logRotator {
                                    numToKeepStr("5")
                                    artifactNumToKeepStr("5")
                                    daysToKeepStr("-1")
                                    artifactDaysToKeepStr("-1")
                                }
                            }
                        }
                    }
                }
                ...
            }
        }
    }
    ...
}
```

Looking at the job configuration it seems to assign properties and works well (looking at the discard of build to 5. Changing the value on the UI and running a branch correctly cleanup build)

property_strategy

I'm probably missing something, but why such deprecation ?

jglick:
Not sure about `job-dsl` plugin but this option should not be offered in the GUI.

sparrowt:
We also use this via `jobdsl` and this deprecation is causing a warning now every time our seed job runs:
```
Warning: (JobCreator.groovy, line 401) buildRetentionBranchProperty is deprecated
```

I'm aware of `buildDiscarder(logRotator(...))` in declarative pipelines, but we use scripted pipelines and also we use jobdsl to generate the main multi-branch pipeline jobs (which define the retention settings) within which then the jenkinsfiles run (scripted pipeline code defined in the repo).

So should this deprecation be reverted to avoid this spurious warning? Would [this code](https://github.com/jenkinsci/workflow-multibranch-plugin/blob/fd072169e96f0bc81d32edf4fb53bbe4535bd775/src/main/java/org/jenkinsci/plugins/workflow/multibranch/JobPropertyStep.java#L229-L233) still do an adequate job of hiding it from the UI if that is the desire?

jglick:
AFAICT https://jenkinsci.github.io/job-dsl-plugin/#path/multibranchPipelineJob-properties is about folder properties, not job properties.

The `properties` step is just part of the `Jenkinsfile` when using Scripted syntax, just as the syntactic sugar `options` block would be in Declarative. I am just responding to

> I'm aware of buildDiscarder(logRotator(...)) in declarative pipelines, but we use scripted pipelines

Regardless of Scripted or Declarative, there is no supported mechanism for declaring build retention strategies as a branch property. Some job properties would benefit from the option of being defined via branch property because they could be needed prior to even starting the first build of a given branch project (when the `properties` step is run, typically in an early stage, and sets up the job). For example, for unusual use cases you might want to define build parameters for a branch project which will only ever be triggered explicitly and not in response to branch indexing / webhooks, and the explicit triggers must specify particular parameter values (there is no sensible default for the first build). This type of consideration does not apply to build retention, since by definition it only matters after you have completed at least one build.

sparrowt:
Yes you're right I'm getting this when trying to use `buildDiscarder` under `multibranchPipelineJob` > `properties`
```
No signature of method: javaposse.jobdsl.dsl.helpers.properties.FolderPropertiesContext.buildDiscarder() is applicable for argument types: (blahblah...$_closure9)
```

Thanks for the background - yes I can see it makes sense that retention _could_ be set in Jenkinsfile (as it doesn't have to be set before first run) though it's kind of sad that this is going to be come the _only_ way with the existing method of setting this in the job definition outside of the Jenkinsfile being deprecated.

That longstanding way of doing it is helpful for several reasons:
- **Grouping:** the retention periods for all jobs are configured in 1 repo (the DSL repo defining all jobs on our jenkins instance)
- **Disk space:** the grouping helps it be an infrastructure level decision because the combined total affects disk space - if we have to move all those into individual Jenkinsfiles scattered around 20+ repos, an individual team can then unilaterally decide "we want to keep our builds for 3 years" and then fill up the disk, or more likely just forget to specify any retention period
- **Defaults:** our 'job creator' groovy (where we're currently using `buildRetentionBranchProperty`) means we can have the DSL repo specify a default which we rely on for the majority of our jobs where we don't pass something more specific when calling the jobcreator

Specifically on the 3rd point (though the other 2 are important), with this deprecation, how would we specify a default so that the disk doesn't fill up indefinitely if someone adds a Jenkinsfile but forgets to specify an explicit retention period?

jglick:
> the grouping helps it be an infrastructure level decision

Note that there is a global build discarder which can be set via JCasC, in addition to the job-level setting (IIRC both will apply if both are defined). That is probably what you want, unless you really meant to have different retention settings per folder or something like that.

sparrowt:
> there is a global build discarder which can be set via JCasC, in addition to the job-level setting (IIRC both will apply if both are defined)

If "both will apply" then presumably a global discarder (e.g. "delete after 30 days") would still delete things after 30 days even if the Jenkinsfile said "delete after 90 days"? If so then that is not a "default" and isn't a viable alternative for obvious reasons.

PR #476, opened 2024-08-16
[JENKINS-60082] Show pull request title in name column

janfaracik:
See [JENKINS-60082](https://issues.jenkins.io/browse/JENKINS-60082)

Would be good to get some thoughts on this/if this is the right approach.

Currently doing a survey on what users find useful on the Jenkins Dashboard (https://forms.gle/6XgzRshaF34kn4wLA) and quite a few of the responses list being able to show the pull request title on the dashboard as being useful (and I agree). So thought I'd take a look into that.

The plugin currently displays the pull request title as the `title` attribute of the link, meaning users have to hover over the PR to reveal what it is. This PR changes that so that it displays the title inline.

Branch job dashboard:
image

Pull requests dashboard:
image

Project status page:

image

### Testing done

* PR titles show inline with the PR number
* Tested with GitHub and GitLab

Multi branch pipelines from before this change will have the old display name but if they are scanned they will get the new display name

### Submitter checklist

### DIFF
--- a/src/main/java/jenkins/branch/ItemColumn.java
+++ b/src/main/java/jenkins/branch/ItemColumn.java
@@ -101,31 +101,6 @@ public class ItemColumn extends ListViewColumn {
     }
 
     /**
-     * Gets the title of a job.
-     *
-     * @param job the job.
-     * @return the unescaped title for displaying in a table column.
-     */
-    @SuppressWarnings("unused") // used via Jelly EL binding
-    public String getTitle(Object job) {
-        // Jelly will take care of quote and ampersand escaping for us
-        if (job instanceof Actionable) {
-            Actionable actionable = (Actionable) job;
-            ObjectMetadataAction action = actionable.getAction(ObjectMetadataAction.class);
-            if (action != null) {
-                String displayName = action.getObjectDisplayName();
-                if (StringUtils.isBlank(displayName) || displayName.equals(actionable.getDisplayName())) {
-                    // if the display name is the same, then the description is more useful
-                    String description = action.getObjectDescription();
-                    return description != null ? description : displayName;
-                }
-                return displayName;
-            }
-        }
-        return null;
-    }
-
-    /**
      * Our extension.
      */
     @Extension
--- a/src/main/java/jenkins/branch/ItemColumn.java
+++ b/src/main/java/jenkins/branch/ItemColumn.java
@@ -101,10 +101,10 @@ public class ItemColumn extends ListViewColumn {
     }
 
     /**
-     * Gets the tool-tip title of a job.
+     * Gets the title of a job.
      *
      * @param job the job.
-     * @return the tool-tip title unescaped for use in an attribute.
+     * @return the unescaped title for displaying in a table column.
      */
     @SuppressWarnings("unused") // used via Jelly EL binding
     public String getTitle(Object job) {
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -2139,7 +2139,7 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
 
             // Default to displaying the project's display name if a trait hasn't been provided
             if (naming == null) {
-                naming = MultiBranchProjectDisplayNamingStrategy.OBJECT_DISPLAY_NAME;
+                naming = MultiBranchProjectDisplayNamingStrategy.RAW_AND_OBJECT_DISPLAY_NAME;
             }
 
             final ObjectMetadataAction action = naming.needsObjectDisplayName()
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -2137,8 +2137,9 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
                     .orElse(null);
             }
 
+            // Default to displaying the project's display name if a trait hasn't been provided
             if (naming == null) {
-                return rawName;
+                naming = MultiBranchProjectDisplayNamingStrategy.OBJECT_DISPLAY_NAME;
             }
 
             final ObjectMetadataAction action = naming.needsObjectDisplayName()
--- a/src/main/java/jenkins/branch/MultiBranchProject.java
+++ b/src/main/java/jenkins/branch/MultiBranchProject.java
@@ -2104,22 +2104,16 @@ public abstract class MultiBranchProject<P extends Job<P, R> & TopLevelItem,
                 throw new IllegalStateException(
                     "Name of created project " + project + " did not match expected " + encodedName);
             }
-            // HACK ALERT
-            // ==========
-            // We don't want to trigger a save, so we will do some trickery to ensure that observer.created(project)
-            // performs the save
-            BulkChange bc = new BulkChange(project);
-            try {
-                project.setDisplayName(getProjectDisplayName(project, rawName));
-            } catch (IOException e) {
-                // ignore even if it does happen we didn't want a save
-            } finally {
-                bc.abort();
-            }
             // decorate contract is to ensure it does not trigger a save
             _factory.decorate(project);
             // ok it is now up to the observer to ensure it does the actual save.
-            observer.created(project);
+            try (BulkChange bc = new BulkChange(project);) {
+                observer.created(project);
+                project.setDisplayName(getProjectDisplayName(project, rawName));
+                bc.commit();
+            } catch (IOException e) {
+                // Ignored
+            }
             doAutomaticBuilds(head, revision, rawName, project, revisionActions, null, null);
         }
 
--- a/src/main/java/jenkins/branch/MultiBranchProjectDisplayNamingStrategy.java
+++ b/src/main/java/jenkins/branch/MultiBranchProjectDisplayNamingStrategy.java
@@ -60,7 +60,14 @@ public enum MultiBranchProjectDisplayNamingStrategy {
                 return rawName;
             }
 
-            return format("%s - %s", rawName, displayName);
+            // The raw name provided here in the context of pull requests is the pull request ID
+            // We tidy up the ID so that they display consistently between SCMs
+            String cleanedUpBranchName = rawName;
+            if (cleanedUpBranchName.startsWith("MR-") || cleanedUpBranchName.startsWith("PR-")) {
+                cleanedUpBranchName = "#" + cleanedUpBranchName.substring(3);
+            }
+
+            return format("%s (%s)", displayName, cleanedUpBranchName);
         }
     },
     ;
--- a/src/main/resources/jenkins/branch/ItemColumn/column.jelly
+++ b/src/main/resources/jenkins/branch/ItemColumn/column.jelly
@@ -25,26 +25,36 @@ THE SOFTWARE.
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout"
          xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:i="jelly:fmt">
-  <td style="${indenter.getCss(job)}">
+  <d:taglib uri="local">
+    <d:tag name="link">
+      <a href="${jobBaseUrl}${job.shortUrl}" class='model-link inside jenkins-table__link'>
+        <j:choose>
+          <j:when test="${it.getTitle(job) == null or h.getRelativeDisplayNameFrom(job, itemGroup).equals(it.getTitle(job))}">
+            <l:breakable value="${h.getRelativeDisplayNameFrom(job, itemGroup)}"/>
+          </j:when>
+          <j:otherwise>
+            ${it.getTitle(job)}
+            <span style="opacity: 0.75; margin-left: 10px"><l:breakable value="${h.getRelativeDisplayNameFrom(job, itemGroup)}"/></span>
+          </j:otherwise>
+        </j:choose>
+      </a>
+    </d:tag>
+  </d:taglib>
+
+  <td style="${indenter.getCss(job)}" xmlns:local="local">
     <j:choose>
       <j:when test="${it.isOrphaned(job)}">
         <s>
-          <a href="${jobBaseUrl}${job.shortUrl}" class='model-link inside jenkins-table_link' title="${it.getTitle(job)}">
-            <l:breakable value="${h.getRelativeDisplayNameFrom(job, itemGroup)}"/>
-          </a>
+          <local:link />
         </s>
       </j:when>
       <j:when test="${it.isPrimary(job)}">
         <strong>
-          <a href="${jobBaseUrl}${job.shortUrl}" class='model-link inside jenkins-table_link' title="${it.getTitle(job)}">
-            <l:breakable value="${h.getRelativeDisplayNameFrom(job, itemGroup)}"/>
-          </a>
+          <local:link />
         </strong>
       </j:when>
       <j:otherwise>
-        <a href="${jobBaseUrl}${job.shortUrl}" class='model-link inside jenkins-table_link' title="${it.getTitle(job)}">
-          <l:breakable value="${h.getRelativeDisplayNameFrom(job, itemGroup)}"/>
-        </a>
+        <local:link />
       </j:otherwise>
     </j:choose>
   </td>
--- a/src/main/resources/jenkins/branch/ItemColumn/column.jelly
+++ b/src/main/resources/jenkins/branch/ItemColumn/column.jelly
@@ -34,7 +34,7 @@ THE SOFTWARE.
           </j:when>
           <j:otherwise>
             ${it.getTitle(job)}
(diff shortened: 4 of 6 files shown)


jtnord:
NB: have not looked at the code, but I am sure there have been some discussions about this in the past but I can not localte them, only the now referenced ticket.

jtnord (commented):
Seems reasonable, javadoc on [`ItemColumn.getTitle()`](https://github.com/jenkinsci/branch-api-plugin/blob/b4227cdbd5a2903adc7d5e3c7f2d67a690853392/src/main/java/jenkins/branch/ItemColumn.java#L103-L110) will be incorrect with this change.

I wonder how this scales with longer titles though.

jtnord (approved):
Going to leave open for a period to let others comment that may recall any history better than me.

src/main/resources/jenkins/branch/ItemColumn/column.jelly
```diff
@@ -25,26 +25,36 @@ THE SOFTWARE.
...
+  <d:taglib uri="local">
+    <d:tag name="link">
+      <a href="${jobBaseUrl}${job.shortUrl}" class='model-link jenkins-table__link'>
+        <j:choose>
+          <j:when test="${it.getTitle(job) == null or h.getRelativeDisplayNameFrom(job, itemGroup).equals(it.getTitle(job))}">
+            <l:breakable value="${h.getRelativeDisplayNameFrom(job, itemGroup)}"/>
+          </j:when>
+          <j:otherwise>
+            ${it.getTitle(job)}
+            <span style="opacity: 0.75; margin-left: 0.625rem; white-space: nowrap;"><l:breakable value="${h.getRelativeDisplayNameFrom(job, itemGroup)}"/></span>
```
timja, line 37: What about adding brackets around the PR branch to more clearly separate? 
```suggestion
            <span style="opacity: 0.75; margin-left: 0.625rem; white-space: nowrap;"><l:breakable value="(${h.getRelativeDisplayNameFrom(job, itemGroup)})"/></span>
```

timja, line 37: would be good if we could drop the `PR-` and convert it to a `#` Would be simpler and consistent with how GitHub displays it

timja (commented):
Looks really good and is an improvement over the current state, one question and one request for a PR to be filed so that users get consistency

----

For consistency can you please send a PR for whatever sets the title for the project index page too please?

image

timja:
I added a screenshot to description from my testing (also shows dark mode)

janfaracik:
> For consistency can you please send a PR for whatever sets the title for the project index page too please?

Browsed around a bit, tried setting `displayName` of the job to the PR title however it's overridden by https://github.com/jenkinsci/branch-api-plugin/blob/b4227cdbd5a2903adc7d5e3c7f2d67a690853392/src/main/java/jenkins/branch/MultiBranchProject.java#L2113

Haven't played too much but it appears there's a trait that can be supplied to set the display name to the [PR title](https://github.com/jenkinsci/branch-api-plugin/blob/b4227cdbd5a2903adc7d5e3c7f2d67a690853392/src/main/java/jenkins/branch/MultiBranchProject.java#L2148)

timja:
> Haven't played too much but it appears there's a trait that can be supplied to set the display name to the [PR title](https://github.com/jenkinsci/branch-api-plugin/blob/b4227cdbd5a2903adc7d5e3c7f2d67a690853392/src/main/java/jenkins/branch/MultiBranchProject.java#L2148) - maybe it'd make sense to provide a default trait or something like that.

Right just change this `orElse(null)` I think:
https://github.com/jenkinsci/branch-api-plugin/blob/b4227cdbd5a2903adc7d5e3c7f2d67a690853392/src/main/java/jenkins/branch/MultiBranchProject.java#L2137

FWIW the traits don't look the best right now and could do with an improvement:

> Job display name with fallback to name

Some have opacity lowered but not all for some reason

image

> Job display name with fallback to name

image

jtnord:
perhaps this is all wrong (I forgot about the traits).  
That is should there be a column for the ID of the PR **and** a column for the title of the PR (which can come from the configured traits)?
As a bonus the view should be configurable to show/hide the columns or add any other column (currently it is all hard coded?).

timja:
> perhaps this is all wrong (I forgot about the traits). 

I think a default trait should be added and most of this can likely be reverted in the jelly at least

> That is should there be a column for the ID of the PR **and** a column for the title of the PR (which can come from the configured traits)? 

No I don't think we want columns, sensible display names by default is what is requested here.

> As a bonus the view should be configurable to show/hide the columns or add any other column (currently it is all hard coded?).

Separate to this but sure.

janfaracik:
image

Updated it to default to `RAW_AND_OBJECT_DISPLAY_NAME`, and modified that to change the formatting to `PR title (#PR number)`

I've done a quick hack in this to format PRs consistently (i.e. drop the PR- prefix and replace it with a #), this should probably be done in the branch source plugins instead.

timja:
Whats with all the test failures?

src/main/java/jenkins/branch/MultiBranchProjectDisplayNamingStrategy.java
```diff
@@ -60,7 +60,14 @@ public String generateName(@NonNull final String rawName, final String displayNa
                 return rawName;
             }
 
-            return format("%s - %s", rawName, displayName);
+            // The raw name provided here in the context of pull requests is the pull request ID
+            // We tidy up the ID so that they display consistently between SCMs
+            String cleanedUpBranchName = rawName;
+            if (cleanedUpBranchName.startsWith("MR-") || cleanedUpBranchName.startsWith("PR-")) {
```
timja, line 66: may need a regex that it's PR-\d+ so its less likely to hit an actual branch name

janfaracik:
Failing on `prj.scheduleBuild2(0).getFuture().get();` - works when `@ClassRule` is changed to `@Rule` - very much out of my knowledge of expertise but maybe something isn't being torn down fast enough?

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -2137,8 +2137,9 @@ private String getProjectDisplayName(@NonNull P project, @NonNull String rawName
                     .orElse(null);
             }
 
+            // Default to displaying the project's display name if a trait hasn't been provided
             if (naming == null) {
-                return rawName;
+                naming = MultiBranchProjectDisplayNamingStrategy.RAW_AND_OBJECT_DISPLAY_NAME;
```
timja, line 2142: This is the line causing the failure, previously it returned but now on L2146 it retrieves the `ObjectMetadataAction` which needs to load more information which is failing.

The `ClassRule` will be needed for performance as it only loads one instance of Jenkins for the whole test. 1s per test rather than like 10s.

I haven't figured out what's causing this though.

@jtnord any ideas?

jtnord: a very quick rough guess would be that there have been no builds of this project yet and the project may not actually be persisted at this point?  (or the metadata is attached after the displayname has used
This now then falls into a lookup,  as the namings strategy requires the "displayName" which means the metadata needs to be looked up, which during creation is not present.

```
java.lang.AssertionError
	at jenkins.model.lazy.AbstractLazyLoadRunMap.load(AbstractLazyLoadRunMap.java:644)
	at jenkins.model.lazy.AbstractLazyLoadRunMap.getByNumber(AbstractLazyLoadRunMap.java:543)
	at jenkins.model.lazy.LazyBuildMixIn.getBuildByNumber(LazyBuildMixIn.java:240)
	at hudson.model.AbstractProject.getBuildByNumber(AbstractProject.java:931)
	at hudson.model.AbstractProject.getBuildByNumber(AbstractProject.java:137)
	at jenkins.model.PeepholePermalink.resolve(PeepholePermalink.java:107)
	at hudson.model.Job.getLastSuccessfulBuild(Job.java:945)
	at jenkins.tasks.SimpleBuildStep$LastBuildActionFactory.createFor(SimpleBuildStep.java:198)
	at jenkins.tasks.SimpleBuildStep$LastBuildActionFactory.createFor(SimpleBuildStep.java:184)
	at hudson.model.Actionable.createFor(Actionable.java:115)
	at hudson.model.Actionable.getAction(Actionable.java:332)
	at jenkins.branch.MultiBranchProject$SCMHeadObserverImpl.getProjectDisplayName(MultiBranchProject.java:2146)
	at jenkins.branch.MultiBranchProject$SCMHeadObserverImpl.observeNew(MultiBranchProject.java:2113)
	at jenkins.branch.MultiBranchProject$SCMHeadObserverImpl.observe(MultiBranchProject.java:2023)
	at jenkins.scm.api.trait.SCMSourceRequest.process(SCMSourceRequest.java:357)
	at jenkins.scm.api.trait.SCMSourceRequest.process(SCMSourceRequest.java:286)
	at jenkins.scm.impl.mock.MockSCMSource.retrieve(MockSCMSource.java:137)
	at jenkins.scm.api.SCMSource._retrieve(SCMSource.java:372)
	at jenkins.scm.api.SCMSource.fetch(SCMSource.java:282)
	at jenkins.branch.MultiBranchProject.computeChildren(MultiBranchProject.java:654)
	at com.cloudbees.hudson.plugins.folder.computed.ComputedFolder.updateChildren(ComputedFolder.java:265)
	at com.cloudbees.hudson.plugins.folder.computed.FolderComputation.run(FolderComputation.java:167)
	at jenkins.branch.MultiBranchProject$BranchIndexing.run(MultiBranchProject.java:1057)
```

IIRC `updateChildren" also creates the new Project. 

NB: there are also some traits where a job gets created for a PR/Branch but will prevent the initial build from triggering (so this could also impact if a build is needed to retreive the info)

jglick:
> works when `@ClassRule` is changed to `@Rule`

Typically means that tests are not cleaning up after themselves, and are interfering with one another for example by trying to use the same job name.

`@ClassRule` is a bit faster but you have to pay more attention to maintenance.

src/main/java/jenkins/branch/MultiBranchProject.java
timja: That's what I was wondering but I've tried what Jan said and switching it from a `ClassRule` to a `Rule` does fix the tests. (as does running tests individually).

I've also tried adding sleeps in some places to see if something is just slow but it didn't help.

Any other thoughts?

> IIRC `updateChildren" also creates the new Project.

is that a suggestion for an alternative method? if so its package-private.

jtnord: > is that a suggestion for an alternative method? if so its package-private.

no - its at the base of the stack - so was just that it may be during creating projects and not just updating existing ones.

timja: I think when you build a multi branch project thats just running indexing. So its after creation and its the the `scheduleBuild2` call that is triggering this.

src/main/resources/jenkins/branch/ItemColumn/column.jelly
```diff
@@ -25,26 +25,28 @@ THE SOFTWARE.
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout"
          xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:i="jelly:fmt">
-  <td style="${indenter.getCss(job)}">
+  <d:taglib uri="local">
+    <d:tag name="link">
+      <a href="${jobBaseUrl}${job.shortUrl}" class='model-link jenkins-table__link'>
```
timja, line 30: extracts the same code into one place and makes minor changes to the classes

src/main/java/jenkins/branch/MultiBranchProject.java
```diff
@@ -2104,22 +2104,16 @@ private void observeNew(@NonNull SCMHead head, @NonNull SCMRevision revision, @N
...
             _factory.decorate(project);
             // ok it is now up to the observer to ensure it does the actual save.
-            observer.created(project);
+            try (BulkChange bc = new BulkChange(project);) {
+                observer.created(project);
+                project.setDisplayName(getProjectDisplayName(project, rawName));
+                bc.commit();
+            } catch (IOException e) {
+                // Ignored
+            }
```
timja, line 2116: this change fixed the tests. the observer.created is what creates the `RunMap` that was null before. (noticed by walking back through a stacktrace of a successful one to see where it was created).

I'm not 100% sure on if there's any impact here and why it was so keen to avoid saving / why a bulk change couldn't be used around everything...

src/main/java/jenkins/branch/MultiBranchProjectDisplayNamingStrategy.java
timja: a branch name won't be hit here because the `rawName` and `displayName` would be the same so the above isn't a concern.

timja (approved):

rsandell:
Personal preference for me woul'd be to have the cr/pr id first then the title, e.g. `#3 did some changes to x`

src/main/java/jenkins/branch/ItemColumn.java
```diff
@@ -100,31 +100,6 @@ public boolean isOrphaned(Object item) {
...
     }
 
-    /**
-     * Gets the tool-tip title of a job.
-     *
-     * @param job the job.
-     * @return the tool-tip title unescaped for use in an attribute.
-     */
-    @SuppressWarnings("unused") // used via Jelly EL binding
-    public String getTitle(Object job) {
```
rsandell, line 110: Removing an old public method signature.

joseblas (approved):

timja:
> Personal preference for me woul'd be to have the cr/pr id first then the title, e.g. `#3 did some changes to x`

Mine is the other way, I will identify far clearer by the name than the number, (which is also what GitHub does):
e.g. 
image

jtnord:
> > Personal preference for me woul'd be to have the cr/pr id first then the title, e.g. `#3 did some changes to x`
> 
> Mine is the other way, I will identify far clearer by the name than the number, (which is also what GitHub does)

So for display purposes I would agree with Tim, but for sorting purposes I would want them sorted by PR number (which is what GH does!).  (this is why I origially said would this not be better as 2 columns because the use cases can not be solved here).
(fortunately 🧌 Jenkins does not at this point in time support pagination - so I can always `<ctrl>+f`)

src/main/java/jenkins/branch/MultiBranchProjectDisplayNamingStrategy.java
jtnord, line 66: this would appear to be putting SCM specifics into an SCM agnostic piece of code?
I believe other systems exist and Gerrit uses `C-xxxx`

timja: Is this a blocker, and if so would this mean an API likely needs to be created where the SCMs set their prefix?

jtnord: So initially I was not thinking it was a blocker, but something that should be addressed.

But now I am thinking is this just confusing?  we use `#nnn` for build numbers in Jenkins and do I really want or need consistency across SCMs?   Indeed having PR-123 in one job and MR-245 in another is a reminder that one project is in my GitHub.com account  and the other is in GitLab on premise.

timja: I think in context with more than one pull request it will be really clear and not going to get confusing at all, e.g.:

image

I don't think most users will care where the SCM is when looking at a job.

jtnord (approved):
the PR-xx -> #xx can always be reverted/changed/made robust with a new API in the future if there is any feedback in this area.

timja:
🚢  ?

src/main/resources/jenkins/branch/ItemColumn/column.jelly
```diff
@@ -25,26 +25,28 @@ THE SOFTWARE.
...
+        <l:breakable value="${h.getRelativeDisplayNameFrom(job, itemGroup)}"/>
+      </a>
+    </d:tag>
+  </d:taglib>
+
+  <td style="${indenter.getCss(job)}" xmlns:local="local">
     <j:choose>
       <j:when test="${it.isOrphaned(job)}">
         <s>
-          <a href="${jobBaseUrl}${job.shortUrl}" class='model-link inside jenkins-table_link' title="${it.getTitle(job)}">
```
daniel-beck, line 32: @janfaracik Why did you remove the `inside` class?

janfaracik: It doesn't do anything from what I've seen from core.

timja:
See https://github.com/jenkinsci/branch-api-plugin/pull/492 for the Firefox fix

src/main/resources/jenkins/branch/ItemColumn/column.jelly
daniel-beck: It controlled whether the context menu dropdown chevron was considered part of the area of the `model-link`. If it doesn't do anything anymore, that's an oversight in https://github.com/jenkinsci/jenkins/pull/6084.

2.332.x `inside`: 
Screenshot 2024-10-29 at 13 23 29

2.332.x otherwise (ugly but this hover effect was fairly short-lived):
Screenshot 2024-10-29 at 13 23 22

basil:
FYI [JENKINS-74811](https://issues.jenkins.io/browse/JENKINS-74811)

PR #481, opened 2024-09-06
Fix incorrect display of the pop-up arrow

gkottr:
Before marked with X
After marked with V

changes


gkottr:
310215976-0abbe2d0-a973-4cd4-8ebd-1bf07cbfdd20

PR #490, opened 2024-10-21
[JENKINS-73947] Improve CSP compatibility

yaroslavafenkin:
* remove redundant inline event handlers in MonitorImpl/message.jelly
* extract inline event handler from MonitorImpl/message.jelly to a separate JavaScript file

https://issues.jenkins.io/browse/JENKINS-73947

### Testing done
Used the following script from JIRA to make the administrative monitor appear:
```groovy
import hudson.model.AdministrativeMonitor
import jenkins.branch.PropertyMigration.MonitorImpl
import jenkins.branch.NoTriggerOrganizationFolderProperty

def monitor = ExtensionList.lookup(AdministrativeMonitor.class).get(MonitorImpl.class)
monitor.add(new NoTriggerOrganizationFolderProperty.PropertyMigrationImpl())
```

https://www.loom.com/share/6ffffdf055be48a7bee6a0fe3c74aeda

https://www.loom.com/share/41c42ef5a87743fba81bb86c9e59bf73

Regarding the removal of `addCrumb` calls. I might be missing the point. I can't find any traces of such method ([trivial org search](https://github.com/search?q=org%3Ajenkinsci+addCrumb&type=code) doesn't fetch any results).
[This part of core](https://github.com/jenkinsci/jenkins/blame/d0c0d07bc4644bc7045ac321472d0a9822dbb8b8/war/src/main/webapp/scripts/hudson-behavior.js#L1450-L1474) has been there for a while, so I'm not sure what that inline extra call is supposed to do. Without it plugin can be installed just fine as you can see on the screen recording.

### Submitter checklist

### DIFF
--- a/src/main/resources/jenkins/branch/PropertyMigration/MonitorImpl/message.jelly
+++ b/src/main/resources/jenkins/branch/PropertyMigration/MonitorImpl/message.jelly
@@ -1,5 +1,5 @@
 <?jelly escape-by-default='true'?>
-<j:jelly xmlns:j="jelly:core">
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler">
   <div class="alert alert-warning" role="alert">
     ${%blurb}
     <table width="100%" class="pane bigtable">
@@ -28,8 +28,9 @@
                 </div>
               </j:when>
               <j:otherwise>
+                <st:adjunct includes="jenkins.branch.PropertyMigration.MonitorImpl.branch-api-property-migration-disabled-form"/>
                 <form method="GET" action="${rootURL}/safeRestart" disabled="disabled"
-                      onsubmit="return false;"
+                      class="branch-api-property-migration-monitor-disabled-form"
                       style="float: none; display: inline; position: inherit;">
                   <span class="yui-button yui-link-button">
                     <span class="first-child">
@@ -65,7 +66,6 @@
                         enctype="multipart/form-data"
                         disabled="disabled"
                         action="${rootURL}/pluginManager/install"
-                        onsubmit="return addCrumb('install-${p.pluginName}');"
                         style="float: none; display: inline; position: inherit;">
                     <input type="hidden" name="plugin.${p.pluginName}.default" checked="true"/>
                     <span class="yui-button yui-link-button" name="dynamicLoad">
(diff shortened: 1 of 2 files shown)


src/main/resources/jenkins/branch/PropertyMigration/MonitorImpl/message.jelly
```diff
@@ -28,8 +28,9 @@
                 </div>
               </j:when>
               <j:otherwise>
+                <st:adjunct includes="jenkins.branch.PropertyMigration.MonitorImpl.branch-api-property-migration-disabled-form"/>
                 <form method="GET" action="${rootURL}/safeRestart" disabled="disabled"
-                      onsubmit="return false;"
+                      class="branch-api-property-migration-monitor-disabled-form"
                       style="float: none; display: inline; position: inherit;">
                   <span class="yui-button yui-link-button">
                     <span class="first-child">
```
yaroslavafenkin, line 36: Alternatively we could take down this whole part that's inside the `otherwise` block. It's a no-op markup that renders an inactive form with an inactive button. And if someone re-enables the form and the button and submits the form the `return false;` statement will prevent the request to action URL.

I could speculate that this code is so rarely (if ever) reachable by now that nobody will ever notice the change here.

src/main/resources/jenkins/branch/PropertyMigration/MonitorImpl/message.jelly
```diff
@@ -65,7 +66,6 @@
                         enctype="multipart/form-data"
                         disabled="disabled"
                         action="${rootURL}/pluginManager/install"
-                        onsubmit="return addCrumb('install-${p.pluginName}');"
```
rsandell, line 68: No alternative, or not needed?

yaroslavafenkin: I'm not confident here.
From PR description:
> Regarding the removal of addCrumb calls. I might be missing the point. I can't find any traces of such method ([trivial org search](https://github.com/search?q=org%3Ajenkinsci+addCrumb&type=code) doesn't fetch any results).
[This part of core](https://github.com/jenkinsci/jenkins/blame/d0c0d07bc4644bc7045ac321472d0a9822dbb8b8/war/src/main/webapp/scripts/hudson-behavior.js#L1450-L1474) has been there for a while, so I'm not sure what that inline extra call is supposed to do. Without it plugin can be installed just fine as you can see on the screen recording.

basil (approved):

basil:
@yaroslavafenkin Can you test the "Restart Jenkins" feature before and after this PR? It might require a little more work to figure out how to induce the button to be active (perhaps making some temporary changes to the Java code).

yaroslavafenkin:
> @yaroslavafenkin Can you test the "Restart Jenkins" feature before and after this PR? It might require a little more work to figure out how to induce the button to be active (perhaps making some temporary changes to the Java code).

I have tested that scenario. Active button doesn't have any CSP violations in it. The form submission to `safeRestart` simply redirects me to that page. I get an error message as it's not available on windows in default mode.

basil (approved):

PR #495, opened 2024-12-10
[JENKINS-74975] Fix Semaphore over-allocation of permits / limit all CleanupTasks

Dohbedoh:
(amends https://github.com/jenkinsci/branch-api-plugin/pull/237)

The `CleanupTask` of the built-in node were scheduled instantly without acquiring permit from the Semaphore but they were releasing permits, hence creating additional permits over time. Reducing the effectiveness of the limit overtime.

Add the Jenkins `built-in` node to list of nodes to be limited.

NOTES: I guess another solution is to continue not limit the number of `CleanupTask` executed for the `built-in` node but that  looks rather confusing in the first place to have a `CLEANUP_THREAD_LIMIT` that only partially limit the Cleanup tasks..

### Testing done

* Create a Jenkins instance with `-Djenkins.branch.WorkspaceLocatorImpl\$Deleter.CLEANUP_THREAD_LIMIT=1`
* In **Manage Jenkins > Script Console**, execute `jenkins.branch.WorkspaceLocatorImpl.Deleter.cleanupPool.availablePermits()` and validate that it shows `1`
* Create a Multibranch pipeline with at least one branch
* Add a filter to have all branches discarded and removed by the orphaned item strategy and save
* In **Manage Jenkins > Script Console**, execute `jenkins.branch.WorkspaceLocatorImpl.Deleter.cleanupPool.availablePermits()` and validate that it still shows `1`

### Submitter checklist

### DIFF
--- a/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
+++ b/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
@@ -66,6 +66,8 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import jenkins.MasterToSlaveFileCallable;
 import jenkins.model.Jenkins;
@@ -395,9 +397,9 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
             }
             TopLevelItem tli = (TopLevelItem) item;
             Jenkins jenkins = Jenkins.get();
-            Computer.threadPoolForRemoting.submit(new CleanupTask(tli, jenkins));
             // Starts provisioner Thread which is tasked with starting cleanup Threads
-            new CleanupTaskProvisioner(tli, jenkins.getNodes()).run();
+            new CleanupTaskProvisioner(tli,
+                Stream.concat(Stream.of(jenkins), jenkins.getNodes().stream()).collect(Collectors.toList())).run();
         }
 
         @Override
--- a/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
+++ b/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
@@ -65,6 +65,7 @@ import java.util.Map;
 import java.util.Queue;
 import java.util.TreeMap;
 import java.util.WeakHashMap;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
@@ -399,17 +400,22 @@ public class WorkspaceLocatorImpl extends WorkspaceLocator {
         private static int runningTasks;
 
         private static ExecutorService executorService() {
-            return CLEANUP_THREAD_LIMIT > 0
-                ? new ContextResettingExecutorService(
+            if (CLEANUP_THREAD_LIMIT > 0) {
+                ThreadPoolExecutor tpe = new ThreadPoolExecutor(CLEANUP_THREAD_LIMIT, CLEANUP_THREAD_LIMIT, 10L, TimeUnit.SECONDS, new LinkedBlockingQueue<>(),
+                    new ExceptionCatchingThreadFactory(
+                        new NamingThreadFactory(
+                            new ClassLoaderSanityThreadFactory(new DaemonThreadFactory()),
+                            "Deleter.cleanupTask")));
+                // Use allowCoreThreadTimeOut to keep a lightweight ThreadPoolExecutor. The Thread pool grows to the
+                // limit and then queue tasks. Thread will then be removed when idle
+                tpe.allowCoreThreadTimeOut(true);
+                return new ContextResettingExecutorService(
                     new ImpersonatingExecutorService(
-                        new ErrorLoggingExecutorService(
-                            new ThreadPoolExecutor(0, CLEANUP_THREAD_LIMIT, 10L, TimeUnit.SECONDS, new SynchronousQueue<>(),
-                                new ExceptionCatchingThreadFactory(
-                                    new NamingThreadFactory(
-                                        new ClassLoaderSanityThreadFactory(new DaemonThreadFactory()),
-                                        "Deleter.cleanupTask")))),
-                        ACL.SYSTEM2))
-                : Computer.threadPoolForRemoting;
+                        new ErrorLoggingExecutorService(tpe),
+                            ACL.SYSTEM2));
+            } else {
+                return Computer.threadPoolForRemoting;
+            }
         }
 
         @Override
--- a/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
+++ b/src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
@@ -43,7 +43,9 @@ import hudson.security.ACL;
 import hudson.security.ACLContext;
 import hudson.slaves.ComputerListener;
 import hudson.slaves.WorkspaceList;
+import hudson.util.ClassLoaderSanityThreadFactory;
 import hudson.util.DaemonThreadFactory;
+import hudson.util.ExceptionCatchingThreadFactory;
 import hudson.util.NamingThreadFactory;
 import hudson.util.TextFile;
 import java.io.BufferedReader;
@@ -56,8 +58,6 @@ import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Semaphore;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
@@ -65,7 +65,8 @@ import java.util.Map;
 import java.util.Queue;
 import java.util.TreeMap;
 import java.util.WeakHashMap;
-import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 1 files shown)


src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -395,9 +397,9 @@ public void onDeleted(Item item) {
             }
             TopLevelItem tli = (TopLevelItem) item;
             Jenkins jenkins = Jenkins.get();
-            Computer.threadPoolForRemoting.submit(new CleanupTask(tli, jenkins));
             // Starts provisioner Thread which is tasked with starting cleanup Threads
-            new CleanupTaskProvisioner(tli, jenkins.getNodes()).run();
+            new CleanupTaskProvisioner(tli,
+                Stream.concat(Stream.of(jenkins), jenkins.getNodes().stream()).collect(Collectors.toList())).run();
```
dwnusbaum, line 402: The comment here seems misleading given the code. I don't think there are any other threads involved. It looks like we synchronously call `run`, which synchronously calls `deleter.acquireThread()`, which synchronously calls `cleanupPool.acquire()`, so IIUC this code can cause `ItemListener.onDeleted` itself to block, which seems undesirable.

Maybe this went unnoticed previously because of the over-allocation you are trying to fix? Or perhaps I misunderstand. If my analysis is correct, then I worry that this PR might cause unexpected blockages, and bigger changes would be needed to fix things. For example, maybe we should instead all of these `CleanupTaskProvisioner` objects to a queue, and then have a single thread processing that queue and submitting the individual tasks to `Computer.threadPoolForRemoting` but making sure not to exceed whatever the desired limits are.

Dohbedoh: Before the change, the listener would still possibly hang unless there is no Jenkins nodes. Now it might even if there is no Jenkins nodes. So you are right, in a way it is more likely to hang in that sense.

Another way to quickly fix the problem is to NOT do the `cleanupPool.release()` for the Controller built-in executor. Directly addresses the issue. But maybe not the more general inconsistent limit and the `ItemListener` contention.

> For example, maybe we should instead submit all of these CleanupTaskProvisioner objects to a queue, and then have a single thread processing that queue and submitting the individual tasks to Computer.threadPoolForRemoting but making sure not to exceed whatever the desired limits are.

Sounds good. Or was also wondering if `Computer.threadPoolRemoting` was required here ? Why not directly use a specific thread pool with unbounded queue (initial thread to 1, max number of threads to the limit set, 60s keep alive time) ?

Dohbedoh: I have update the PR with a single thread executor service to manage the scheduling.

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -382,22 +389,27 @@ public static class Deleter extends ItemListener {
 
         private static /* almost final */ int CLEANUP_THREAD_LIMIT = SystemProperties.getInteger(Deleter.class.getName() + ".CLEANUP_THREAD_LIMIT", Integer.valueOf(0)).intValue();
 
+        private final transient ExecutorService executorService = Executors.newSingleThreadExecutor(threadFactory());
```
rsandell, line 392: I'm unsure; Will this survive a reload configuration from disk?

dwnusbaum: I don't think this class is saved to disk at all; it's just an extension. If you restart Jenkins while a lot of deletion tasks are waiting to run, I do think they will all be lost if that's what you mean, but it looks like that was the case before this PR too.

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -382,22 +389,27 @@ public static class Deleter extends ItemListener {
...
         /** Semaphore for limiting number of scheduled {@link CleanupTask} */
-        private static Semaphore cleanupPool = new Semaphore(CLEANUP_THREAD_LIMIT, true);
+        private static final Semaphore cleanupPool = new Semaphore(CLEANUP_THREAD_LIMIT, true);
 
         /** Number of {@link CleanupTask} which have been scheduled but not yet completed. */
         private static int runningTasks;
 
+        private static ThreadFactory threadFactory() {
+            return new NamingThreadFactory(new DaemonThreadFactory(), "DeleterCleanupTask");
+        }
```
dwnusbaum, line 402: If you go down the route of creating a new executor service I think you need quite a few more modifiers to avoid various problems: https://github.com/jenkinsci/jenkins/blob/d22cc2fa3c426420514240e351faa56ea12e5446/core/src/main/java/hudson/model/Computer.java#L1348-L1353

With the current approach of a single thread executor to dispatch the tasks, it might be simpler to instead use `new SingleLaneExecutorService(Computer.threadPoolForRemoting)` since we're not using the new thread pool for the actual `CleanupTask`s anyway.

> Or was also wondering if Computer.threadPoolRemoting was required here ? Why not directly use a specific thread pool with unbounded queue (initial thread to 1, max number of threads to the limit set, 60s keep alive time) ?

This idea seems fine too and I think would allow us to get rid of all the semaphores to simplify things, right? The only thing I would consider changing would be to make it so that if `CLEANUP_THREAD_LIMIT` is 0, continue to use `Computer.threadPoolForRemoting` directly to avoid creating an extra unbounded thread pool. You'd no longer be able to edit `CLEANUP_THREAD_LIMIT` without restarting Jenkins, but that seems fine.

Dohbedoh: > This idea seems fine too and I think would allow us to get rid of all the semaphores to simplify things, right? The only thing I would consider changing would be to make it so that if CLEANUP_THREAD_LIMIT is 0, continue to use Computer.threadPoolForRemoting directly to avoid creating an extra unbounded thread pool. You'd no longer be able to edit CLEANUP_THREAD_LIMIT without restarting Jenkins, but that seems fine.

Allright, I'll propose something like that then.

Dohbedoh: Per my understanding of environment where this happen (and branch indexing in general) the expected behavior is to see spikes of CleanupTasks at irregular interval during the day, rather than seeing a constant queuing of them.

So I propose `ThreadPoolExecutor` with a `corePoolSize` at `0` and setting the max per `CLEANUP_THREAD_LIMIT` with `SynchronousQueue` so that the thread pool can react quickly to the load, spawning up to the maximum number of threads and eventually retiring the 10s idle threads until the next batch.

src/main/java/jenkins/branch/WorkspaceLocatorImpl.java
```diff
@@ -380,24 +391,44 @@ static String minimize(String name) {
...
 
         /** Number of {@link CleanupTask} which have been scheduled but not yet completed. */
         private static int runningTasks;
 
+        private static ExecutorService executorService() {
+            return CLEANUP_THREAD_LIMIT > 0
+                ? new ContextResettingExecutorService(
+                    new ImpersonatingExecutorService(
+                        new ErrorLoggingExecutorService(
+                            new ThreadPoolExecutor(0, CLEANUP_THREAD_LIMIT, 10L, TimeUnit.SECONDS, new SynchronousQueue<>(),
```
dwnusbaum, line 406: New approach looks mostly good, but why `SynchronousQueue`? If I understand correctly, that will make it so that whenever there are more than `CLEANUP_THREAD_LIMIT` tasks, each submission will block until another task completes. I would use `LinkedBlockingQueue` instead so that when there is a burst of activity, we allow tasks to be queued up and do not block in `ItemListener.onDeleted`.

dwnusbaum: Hmm, reading through https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/ThreadPoolExecutor.html though it seems that using an unbounded queue would probably result in only ever starting a single thread, and worse, it seems like with `SynchronousQueue` we will actually _reject_ submitted tasks if we ever have more than `CLEANUP_THREAD_LIMIT` in flight at once.

Looking at Javadoc it does not seem possible to get the behavior we want with `ThreadPoolExecutor` alone. I guess the simplest thing would be to have the core pool size be `CLEANUP_THREAD_LIMIT`, leaving those threads basically permanently idle. Another approach could be to keep `SynchronousQueue`, but then have a separate unbounded queue that we add tasks to here, and then have a `PeriodicWork` or some other similar single-thread utility that repeatedly tries to submit tasks in the unbounded queue to the thread pool. Awkward.

Dohbedoh: > Hmm, reading through https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/ThreadPoolExecutor.html though it seems that using an unbounded queue would probably result in only ever starting a single thread, and worse, it seems like with SynchronousQueue we will actually reject submitted tasks if we ever have more than CLEANUP_THREAD_LIMIT in flight at once.

Correct. Tasks are being rejected..

> the core pool size be CLEANUP_THREAD_LIMIT, leaving those threads basically permanently idle

Seems to me that people impacted would put a value in the 1X / 1XX. And keeping 1XX threads alive might be overkill. But with a keepAlive time and [allowCoreThreadTimeOut](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/ThreadPoolExecutor.html#allowCoreThreadTimeOut(boolean)) set to `true` this should have the behavior we want. Bringing threads up to the limit when needed, and then bring the pool down to `0` when no tasks are remaining and threads are becoming idle.

Dohbedoh: Done. Couple of things that I tested:

* Startup size of the pool is `0`. With `CLEANUP_THREAD_LIMIT=100` on startup, I don't see any Deleter thread.
* Tasks enqueued when reaching the limit
* Threads are reuse. With `CLEANUP_THREAD_LIMIT=2` for the deletion of 4 branches across 4 nodes, I can see that the same thread (IDs) are being reused.

dwnusbaum: Ah interesting, I wasn't familiar with `allowCoreThreadTimeOut`. That definitely looks useful here.

dwnusbaum (approved):

Dohbedoh:
@dwnusbaum Can we merge it ?

dwnusbaum:
@Dohbedoh sure yes sorry I missed your earlier message

PR #497, opened 2024-12-15
Remove owner icon spec from metadata folder icon

timja:
see https://github.com/jenkinsci/cloudbees-folder-plugin/pull/451, https://github.com/jenkinsci/cloudbees-folder-plugin/pull/450 and https://github.com/jenkinsci/jenkins/pull/10049

This was introduced in massive PR: https://github.com/jenkinsci/branch-api-plugin/pull/58
No idea if its actually used / still needed, will try to investigate.

### Testing done

Tests in this plugin are passing
Tested in Bom PCT are passing https://github.com/jenkinsci/bom/pull/4104

I'll try run this with a breakpoint on in a few setups and see if I can hit it

### Submitter checklist

### DIFF
--- a/src/main/java/jenkins/branch/MetadataActionFolderIcon.java
+++ b/src/main/java/jenkins/branch/MetadataActionFolderIcon.java
@@ -100,12 +100,6 @@ public class MetadataActionFolderIcon extends FolderIcon {
                 }
                 // otherwise the metadata doesn't want to control the icon, so fall back to the descriptor's default
             }
-            if (owner instanceof IconSpec) {
-                String result = ((IconSpec) owner).getIconClassName();
-                if (result != null) {
-                    return result;
-                }
-            }
             return owner.getDescriptor().getIconClassName();
         }
         return "icon-folder";


timja:
Used in Organisation folders...

jglick:
Not sure in detail what this code is for, but from the description, it might be used only to set an icon for an org folder. Which is very cosmetic and something we could get rid of if it has some ill effect on performance. Compare https://github.com/jenkinsci/github-branch-source-plugin/pull/700#issuecomment-1710750053.

timja:
> Not sure in detail what this code is for, but from the description, it might be used only to set an icon for an org folder. Which is very cosmetic and something we could get rid of if it has some ill effect on performance. Compare [jenkinsci/github-branch-source-plugin#700 (comment)](https://github.com/jenkinsci/github-branch-source-plugin/pull/700#issuecomment-1710750053).

Possibly, doesn't effect performance though just impacting the command palette, I've done a workaround in the folders plugin, not the cleanest but everything works: https://github.com/jenkinsci/cloudbees-folder-plugin/pull/452

